PR #3, opened 2018-01-17
Bpk 1215 split pod into subspecs

k0nserv:


Example/Tests/BPKColorTest.m
```diff
@@ -18,7 +18,7 @@
 
 #import <XCTest/XCTest.h>
 
-#import <Backpack/Backpack.h>
+#import <Backpack/Color.h>
```
matthewdavidson: To confirm, this is the breaking change?

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# Unreleased
+
+*Breaking:*
+
+- Moved colors to a subspec in `Backpack/Color`.
```
matthewdavidson: Side note, we should add dates to the changelog when we release (YYYY-MM-DD)

Example/Tests/BPKColorTest.m
k0nserv: Yeah

CHANGELOG.md
k0nserv: Sure thing :)

k0nserv:
I'm thinking we release this as `0.1.0`

PR #4, opened 2018-01-17
[BPK-1208] Add fonts

k0nserv:


k0nserv:
Here's what it looks like


image

tzahola:
Goot to merge by me

Backpack/Classes/Font/Generated/BPKFont.h
```diff
@@ -0,0 +1,37 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+@class UIFont;
+@interface BPKFont: NSObject
+
++ (UIFont *)textXs;
```
bogren: Tip: add `/// System font of size 11, regular weight` or similar. I at least tend to forget the size and weights.

k0nserv: The use of Backpack at Skyscanner means design docs will already be specifying one of these sizes so when using it you shouldn't have to care about the underlying values.

k0nserv:
Good point @shaundon I'll change those

k0nserv:
Since Travis is being painfully slow here's the result of `bundle exec rake ci` locally. 

image

@shaundon can you please give me a ✅ ?

PR #8, opened 2018-01-19
WIP: [BPK-1202] API docs WIP

k0nserv:


k0nserv:
Here's a preview of what these docs look like

[docs.zip](https://github.com/Skyscanner/backpack-ios/files/1646378/docs.zip)

matthewdavidson:
@k0nserv is this dead? can we close?

k0nserv:
I wouldn't say dead. The code might still be useful if we decide to pursue docs for iOS

PR #10, opened 2018-01-19
[BPK-1205] Ensure gulp has been run on CI

k0nserv:


.travis.yml
```diff
@@ -4,8 +4,17 @@
 
 osx_image: xcode9.2
 language: objective-c
-cache: cocoapods
+cache:
+  - cocoapods
+  - bundler
+  - npm
```
matthewdavidson: I dont think thsi actually works - i think you need to use:
```
  directories:
  - node_modules
```

k0nserv: What makes you think it doesn't?

.travis.yml
```diff
@@ -4,8 +4,15 @@
...
 install:
+  - nvm install 8.9.4
```
matteo-hertel: this is less than ideal but nvm does not play well with travis on this base image so for now this will do

package.json
```diff
@@ -8,7 +8,7 @@
   },
   "scripts": {
     "preinstall": "node -e \"$(curl -fsSL https://raw.githubusercontent.com/Skyscanner/ensure-node-env/master/dist/index.js)\"",
-    "gulp": "gulp"
+    "build": "gulp"
```
matteo-hertel: consistency between iOS and Android

matteo-hertel:
Example of build stopped because the artifact of `npm run build` are different from the commited artifacts https://travis-ci.org/Skyscanner/backpack-ios/builds/334709948

.travis.yml
matthewdavidson: See https://docs.travis-ci.com/user/caching/#Configuration

```yaml
language: ruby
cache:
  bundler: true
  directories:
  - node_modules # NPM packages
```

The example in there doesn;t have `npm` as a identifier for the `cache` option, it specifies it as a directory

.travis.yml
```diff
@@ -4,8 +4,15 @@
...
+  - npm install
+  - npm run build
+  - (! git status --porcelain | grep -q .) || (echo "Please ensure you have ran `npm run gulp`to generate the templated classes." && false)
```
matthewdavidson: Can we use the same script that @matteo-hertel created for the android codebase?

matteo-hertel:
After changing the script to check the pristine state of the artifact, this is an example of a failing build https://travis-ci.org/Skyscanner/backpack-ios/builds/335092192

PR #11, opened 2018-01-29
[BPK-1209] Add spacing to cocoapod

shaundon:
Usage:

```
#import <Backpack/Spacing.h>

NSNumber *someNumber = [NSNumber numberWithFloat: BPKSpacingSm];
```

Initially I wanted this to be more 'classy', e.g. we'd do `BPKSpacing.spacingSm` but from what I've seen online that isn't really a thing with constants:
1. https://stackoverflow.com/questions/538996/constants-in-objective-c
2. http://www.ios-blog.co.uk/tutorials/objective-c/constants-in-objective-c/

Also I initially tried this with `@interface` but @k0nserv popped up and said we don't need that.

I haven't written tests yet, I'm unsure of how to do this right now as `CGFloat` doesn't seem to work with `respondsToSelector` so I'm going to wait until Hugo is back to tackle that.

screen shot 2018-01-29 at 11 25 18


Example/Backpack/View Controllers/BPKSpacingsViewController.m
```diff
@@ -0,0 +1,64 @@
...
+- (void)viewDidLoad {
+    [super viewDidLoad];
+    
+    [self setSpacings];
+}
+
+#pragma mark - Private
+
+- (void)setSpacings {
+    NSArray<NSString *> *spacingKeys = @[
```
k0nserv: Use a `NSDictionary` instead

shaundon: I did use that initially but apparently they're unordered, so using the labels iterator to pull out spacings wouldn't be reliable.

PR #14, opened 2018-02-06
[BPK-1253] Add BPKRadii values

shaundon:
simulator screen shot - iphone 8 plus - 2018-02-06 at 10 49 45

### DIFF
--- a/Backpack.podspec
+++ b/Backpack.podspec
@@ -48,6 +48,11 @@ Pod::Spec.new do |s|
     ss.public_header_files = 'Backpack/Classes/Spacing/**/*.h'
   end
 
+  s.subspec 'Radii' do |ss|
+    ss.source_files = 'Backpack/Classes/Radii/**/*.{h,m}'
+    ss.public_header_files = 'Backpack/Classes/Radii/**/*.h'
+  end
+
   s.frameworks = "UIKit", "Foundation"
   s.requires_arc = true
 end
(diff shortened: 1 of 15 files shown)


Backpack.podspec
```diff
@@ -16,7 +16,7 @@
 
 Pod::Spec.new do |s|
   s.name             = "Backpack"
-  s.version          = "0.3.0"
+  s.version          = "0.4.0"
```
k0nserv: This increment should be done by the release task so don't do it here

Example/Tests/BPKRadiiTest.m
```diff
@@ -0,0 +1,43 @@
...
+
+NS_ASSUME_NONNULL_BEGIN
+@interface BPKRadiiTest : XCTestCase
+
+@end
+
+@implementation BPKRadiiTest
+
+- (void)testRadiusValueExists {
+    XCTAssertNotNil([NSNumber numberWithFloat:BPKBorderRadiusSm]);
```
k0nserv: This test is a bit redundant since the test below will fail to compile all together if these values don't exist

Backpack.podspec
shaundon: Ah - I was 50/50 on whether I did it or the release did. I chose...poorly.

Example/Tests/BPKRadiiTest.m
shaundon: Makes sense - we have the same thing for spacings, so I'll remove that too.

Backpack.podspec
k0nserv: No problem. The code is here https://github.com/Skyscanner/backpack-ios/blob/ebeb518d36250225c6bb0af9e8139a4d29ddae71/Rakefile#L75-L77

PR #15, opened 2018-02-08
Bpk 1247 add shadow tokens

k0nserv:
Here's what it looks like

simulator screen shot - iphone 8 - 2018-02-08 at 16 10 23

### DIFF
--- a/Backpack/Classes/Shadow/Generated/BPKShadow.m
+++ b/Backpack/Classes/Shadow/Generated/BPKShadow.m
@@ -1,0 +1,103 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2018 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#import "BPKShadow.h"
+#import <UIKit/UIKit.h>
+
+NS_ASSUME_NONNULL_BEGIN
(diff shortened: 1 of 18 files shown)


Backpack/Classes/Shadow/Generated/BPKShadow.m
```diff
@@ -0,0 +1,103 @@
...
+    layer.shadowRadius = self.radius;
+    layer.shadowOpacity = self.opacity;
+}
+
+
++ (instancetype)shadowSm {
+    static dispatch_once_t shadowSmOnceToken;
+    static BPKShadow *shadowSm;
+
+    dispatch_once(&shadowSmOnceToken, ^{
```
k0nserv: This pattern is a thread safe way to ensure only a single instance of a value gets created.

Example/Tests/BPKShadowTest.m
```diff
@@ -0,0 +1,39 @@
...
+        XCTAssert([BPKShadow respondsToSelector:selector]);
+        XCTAssertNotNil([BPKShadow performSelector:selector]);
+    }
+}
+
+- (void)testStylesAreCached {
+    BPKShadow *shadow1 = [BPKShadow shadowSm];
+    BPKShadow *shadow2 = [BPKShadow shadowSm];
+
+    XCTAssertTrue(shadow1 == shadow2, @"BPKShadow should cache shadow styles");
```
k0nserv: This verifies we never create more than a single value per shadow type by doing pointer comparison. Object comparison in objective-c is done with `isEqual:`

README.md
```diff
@@ -65,6 +65,20 @@ CGFloat myAppBorderRadiusSmall = BPKBorderRadiusSm;
...
+### Backpack/Shadow
+
+`Backpack/Shadow` contains the Backpack shadow definitions in the class `BPKShadow`. Add `Backpack/Shadow` to your `Podfile` or `podspec` and then use it as follows.
+
+```objective-c
+#import <Backpack/Shadow.h>
+
+BPKShadow *shadow = [BPKShadow shadowSm];
+// BPKShadow *shadowLg = [BPKShadow shadowLg];
+// BPKShadow *shadowXl = [BPKShadow shadowXl];
```
shaundon: Just want to confirm that these are supposed to be commented out?

k0nserv: Yeah

PR #18, opened 2018-02-23
added alpha to ios colors

danielnagyx:


k0nserv:
Per my comment in https://github.com/Skyscanner/backpack/pull/534#issuecomment-367986648 we should strip the color prefix here instead. With regards to clear I think the best option is just add a non-templated method to `BPKColor` that just calls `[UIColor clearColor]`

gulpfile.js
```diff
@@ -85,7 +85,14 @@ const generatePrefixedConst = ({ type, name, value }) => {
 const parseTokens = tokensData => {
   const colors = _.chain(tokensData.properties)
     .filter(({ type }) => type === 'color')
-    .map(({ value, ...rest }) => ({ value: parseColor(value), ...rest }))
+    .map(({ value, name, ...rest }) => {
+      const newName = name.replace('color', '');
+      return {
+        value: parseColor(value),
+        name: newName[0].toLowerCase() + newName.slice(1),
```
k0nserv: 👍

k0nserv:
Looking good. The Objective-C tests need to be updated and you should ensure that that the javascript has been formatted by prettier(`npm run prettier`).

PR #25, opened 2018-07-13
[BPK-1671] Integrate prettier

k0nserv:
Use the new `eslint-config-skyscanner-with-prettier`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,18 +29,10 @@
   },
   "homepage": "https://github.com/Skyscanner/backpack-ios#readme",
   "devDependencies": {
-    "babel-eslint": "^8.2.1",
     "bpk-tokens": "^27.0.4",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -49,7 +41,7 @@
     "lint-staged": "^7.0.0",
     "lodash": "^4.17.4",
     "merge2": "^1.2.1",
-    "prettier": "^1.10.2",
+    "prettier": "^1.13.7",
```
matthewdavidson: You don't need this right? its shipped as part of `eslint-config-skyscanner-with-prettier`?

k0nserv: Ehm yeah that's a good point

matthewdavidson: Also just noticed that prettier should probably be removed from the `backpack-android` deps too...

PR #27, opened 2018-08-15
[BPK-1807] Port panel from monorepo

k0nserv:


Backpack.podspec
```diff
@@ -14,6 +14,18 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+define_subspec = Proc.new do |name, root_spec|
```
tevelee: very nice refactor!

Backpack/Classes/Panel/BPKPanel.m
```diff
@@ -0,0 +1,96 @@
...
+        self.layoutMargins = self.originalLayoutMargins;
+    } else {
+        self.layoutMargins = UIEdgeInsetsZero;
+    }
+    
+    _padded = padded;
+}
+
+- (void)addSubview:(UIView *)view {
+    NSAssert(self.subviews.count == 0, @"BPKPanel can only have a single subview");
```
tevelee: why do we have this restriction?

Backpack/Classes/Panel/BPKPanel.m
```diff
@@ -0,0 +1,96 @@
...
+@property (assign) UIEdgeInsets originalLayoutMargins;
+
+- (void)setup;
+@end
+
+
+@implementation BPKPanel
+
+- (instancetype)initWithFrame:(CGRect)frame {
+    return [self initWithPadded:YES];
```
tevelee: I think you should call. `[super initWithFrame:frame]`, so the frame property will be correctly set. This also removes confusion whether `iniWithPadded` calls any super initialisers

Backpack/Classes/Panel/BPKPanel.m
```diff
@@ -0,0 +1,96 @@
...
+
+    return self;
+}
+
+- (instancetype)initWithPadded:(BOOL)padded {
+    self = [super initWithFrame:CGRectZero];
+
+    if (self) {
+        [self setup];
+        self.padded = padded;
```
tevelee: Not sure it's an issue, but setup and self.padded will call the setPadded method twice during the runtime of this method. (Also, the order of calling setup and the padded setter matters.) 

Maybe it would be nicer to provide isPadded as a parameter of the setup method.
All other initialisers then can call setup with `[self setup:BPKPanelDefaultPaddedValue];`

Example/Backpack.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>IDEDidComputeMac32BitWarning</key>
```
tevelee: I don't think this warning flag needs to be committed

Backpack.podspec
k0nserv: Too bad it doesn't actualyl work, had to remove it :(

Example/Backpack.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist
k0nserv: I did some reading and I believe you should commit `xcsharedddata`

Backpack/Classes/Panel/BPKPanel.m
k0nserv: Because otherwise managing the complexities of the layout become unwieldy. A consumer can add a single view and then add as many more as they want. Initially I was working with a `contentView` approach instead, but AFAIK you can't make that work correctly when the view is used from IB

Backpack/Classes/Panel/BPKPanel.m
k0nserv: Yeah good shout

Backpack/Classes/Panel/BPKPanel.m
k0nserv: I like it

PR #28, opened 2018-08-15
[BPK-1807] Port Badge from monorepo

k0nserv:


k0nserv:
> All looking good. Tests don't look particularly extensive but I guess that's intentional as we're just trying to establish the way forward

Yeah we will tackle snapshot tests later

PR #35, opened 2018-08-28
[BPK-1826] Restructure pod to prepare for docs

k0nserv:
With the documentation approach we'll keep `README.md`s with each
component. The previous structure wasn't helpful for this kind of
approach so the structure had to change.


shaundon:
CI is failing because of the pristine check 😭

Example/Backpack.xcodeproj/xcshareddata/xcschemes/Backpack Native.xcscheme
```diff
@@ -26,6 +26,7 @@
       buildConfiguration = "Debug"
       selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
       selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
+      language = ""
```
shaundon: (Not an actual comment I just wanted to use that gif)

k0nserv:
> Does this break the paths in Backpack.h or will they keep working?
It keeps working. `#import` in Objective-C and `#include` in C/C++ aren't based on location in the file tree, but on what are called header search paths. As long as the files are in the header search paths they are found.

k0nserv:
I have no idea why CI is breaking, `Backpack/Classes` has been removed. We used to have `Backpack/Classes/.gitkeep`, but I've removed that too. My only guess is that Travis has some funky cache of a previous checkout that causes the issue

PR #38, opened 2018-08-29
[BPK-1826] Create per subspec `README.md`s

k0nserv:


Backpack/Shadow/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Usage
+
+`Backpack/Shadow` contains the Backpack shadow definitions in the class `BPKShadow`.
+
+```objective-c
+#import <Backpack/Shadow.h>
+
+BPKShadow *shadowSm = [BPKShadow shadowSm];
+// BPKShadow *shadowLg = [BPKShadow shadowLg];
```
matthewdavidson: why are these commented out?

k0nserv: Just to illustrate that they are there, makes it easier to copy the code

Backpack/Gradient/README.md
```diff
@@ -0,0 +1,26 @@
...
+pod 'Backpack/Gradient'
+```
+
+and then run `pod install`.
+
+## Usage
+
+`Backpack/Gradient` contains the Backpack gradient definitions in the class `BPKGradient`. `BPKGradient` has the two properties:
+
+* `colors` of type `NSArray<UIColor *> *`
```
tiagoengel: Can we move props to a different section like we do with react/react native components?

k0nserv: I think we should just remove these all together in favour of API docs in code

PR #41, opened 2018-08-31
[NoJira] improve consistency with the monorepo

matteo-hertel:
Just some consistency improvement with the monorepo

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,6 @@
-# Unreleased
+# Backpack iOS changelog
+
+[Unreleased changes](./UNRELEASED.md).
 
 ## 0.9.0
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -27,7 +27,7 @@ Tests can be ran as usual from Xcode(Product -> Test or cmd+U). Snapshot tests s
 
 > Backpack team only
 
-To issue a new release make sure you've set the project up as above, that you have push acess to the Backpack CocoaPod and that you're logged in to CocoaPods trunk. `bundle exec pod trunk me` should print your info and including the Backpack pod in the output. Make sure you've added a new entry in the `CHANGELOG.md` with the version you're publishing then run `bundle exec rake release`.
+To issue a new release make sure you've set the project up as above, that you have push acess to the Backpack CocoaPod and that you're logged in to CocoaPods trunk. `bundle exec pod trunk me` should print your info and including the Backpack pod in the output. Move all the entries from the `UNRELEASED.md` file to the `CHANGELOG.md` and add them under the header with the new version you're about to publish, then run `bundle exec rake release`.
 
 
 [0]: https://github.com/rbenv/rbenv
--- a/README.md
+++ b/README.md
@@ -1,4 +1,6 @@
-# Backpack
+# Backpack iOS
+
+> Backpack is a collection of design resources, reusable components and guidelines for creating Skyscanner's products.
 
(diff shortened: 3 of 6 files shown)


README.md
```diff
@@ -1,5 +1,7 @@
 # Backpack
```
matteo-hertel: I'd like this to be `Backpack Design System` like the monorepo, thoughts?

CHANGELOG.md
```diff
@@ -1,4 +1,6 @@
-# Unreleased
+# Backpack changelog
```
matthewdavidson: Should be `Backpack iOS changelog`

README.md
matthewdavidson: How about `Backpack iOS`?

decisions/README.md
```diff
@@ -0,0 +1,7 @@
+# Backpack decisions
```
matthewdavidson: Hmm I would have thought we should add this when we actually have decisions ...

UNRELEASED.md
```diff
@@ -0,0 +1,3 @@
+# Unreleased
```
k0nserv: If we are going to add this file we need to update the instructions in `CONTRIBUTING.md` for releases. Also do we really need this in this repo? The reason we have it in the main repo was the volume of change we accumulate?

matteo-hertel:
The key for this PR is to have as much consistency as possible across our repos, so switching between them is less painless, unless major concerns I'd really like this changes to go through 

As per the name I believe we had that discussion back in the days and we chose not to add iOS as it's in the name of the repo (or something like that)

README.md
shaundon: I'm happy with that

k0nserv:
I don't feel as strongly about consistency, but I am happy with the change 🤷‍♂️

matthewdavidson:
I agree with consistency, but the decisions in particular seem to be YAGNI. Also the names surely need to be different ... just look at the URL ^^

matteo-hertel:
I can rename to Backpack iOS if we're all happy with it cc @shaundon @georgegillams 

As per the decisions, I don't see why YAGNI, it'll drive consistency in the repos and it'll save somebody else the copy paste from the Backpack repo when we come down to make a decision, does it look weird having it? if so I'll happily remove it

georgegillams:
I like keeping things as consistent as possible - I imagine as time goes on we will have some members of the squad who predominantly work on one platform over another. We're gonna get stupid errors if the process varies between every repo

CONTRIBUTING.md
```diff
@@ -27,7 +27,7 @@ Tests can be ran as usual from Xcode(Product -> Test or cmd+U). Snapshot tests s
 
 > Backpack team only
 
-To issue a new release make sure you've set the project up as above, that you have push acess to the Backpack CocoaPod and that you're logged in to CocoaPods trunk. `bundle exec pod trunk me` should print your info and including the Backpack pod in the output. Make sure you've added a new entry in the `CHANGELOG.md` with the version you're publishing then run `bundle exec rake release`.
+To issue a new release make sure you've set the project up as above, that you have push acess to the Backpack CocoaPod and that you're logged in to CocoaPods trunk. `bundle exec pod trunk me` should print your info and including the Backpack pod in the output. Make sure you've added a new entry in the `UNRELEASED.md` and update the `CHANGELOG.md` file with the version you're publishing then run `bundle exec rake release`.
```
k0nserv: This is not correct, when releasing you should move the entries from `UNRELEASED.md` to `CHANGELOG.md` and add them under the header with the new version and then run `bundle exec rake release`

PR #43, opened 2018-09-04
[BPK-1878] Setup matrix build

k0nserv:
Build and run tests on Xcode 9, Xcode 8, and Xcode 7 with the latest
versions of iOS 11, iOS 10 and iOS 9 respectively.


Example/Backpack_Tests copy-Info.plist
```diff
@@ -0,0 +1,22 @@
...
+	<key>CFBundlePackageType</key>
+	<string>BNDL</string>
+	<key>CFBundleShortVersionString</key>
+	<string>1.0</string>
+	<key>CFBundleSignature</key>
+	<string>????</string>
+	<key>CFBundleVersion</key>
+	<string>1</string>
+</dict>
+</plist>
```
shaundon: Is this a dupe?

.travis.yml
```diff
@@ -9,20 +9,26 @@ cache:
...
+matrix:
+  include:
+    - osx_image: xcode9.4
+      env: BUILD_SDK=iphonesimulator11.4 DESTINATION="platform=iOS Simulator,name=iPhone 8" MAIN_STAGE="true"
+    - osx_image: xcode8.3
+      env: BUILD_SDK=iphonesimulator10.3 DESTINATION="platform=iOS Simulator,name=iPhone 7" SNAPSHOT_TESTS="false" MAIN_STAGE="false"
 before_install:
   - nvm install 8.9.4
-  - npm install -g greenkeeper-lockfile@1
+  - '[ "$MAIN_STAGE" = true ] && npm install -g greenkeeper-lockfile@1 || true'
```
shaundon: I think the naming here is good, but just wondering if we should have a comment at the top of the file or something to explain _why_ we only run tests on >= iOS 11

PR #45, opened 2018-09-06
[BPK-1873] Run analyser on CI

shaundon:
Added a new `analyze` task to `rake ci` task, which does the following:

1. Runs `xcodebuild analyze`
2. Pipes stderr to stdout
3. greps for any lines that start with '⚠️', as this is what `xcpretty` prints when a file doesn't pass analysation
4. Returns 5 lines below any lines that match, for context
5. Inverts the error code so that it exits with `1` if the error message is in the output

**Notes:**

The analyze task only fails correctly if this is the first time you run it. So locally you need to do a clean each time it's run. On CI this should be ok as it'll always be clean.

To make it fail locally, pop some code that'll set it off in one of your view controllers:

``` 
int index = 0;
index = 1;
```

Then run `(cd Example && bundle exec rake analyze)` (make you sure clean first!)


shaundon:
This is what it looks like when it fails:

screen shot 2018-09-11 at 08 51 47

Pertinent part:

```
The following commands produced analyzer issues:
	Analyze Backpack/View\ Controllers/BPKFontsViewController.m normal x86_64
(1 command with analyzer issues)
```

shaundon:
Caveats: I don't think this is perfect - for one thing it doesn't actually show the errors, just the files they appear in. Nonetheless in true 'done is better than perfect' fashion, I think this is fine for our current needs.

Man, these tools are shocking..

shaundon:
> I can imagine that being annoying down the line.

I've updated it to return the entire output if it fails, so errors will now be surfaced.

shaundon:
Update: I've improved the error display a lot, it now shows you the actual errors when it fails. Here's an example:

screen shot 2018-09-12 at 09 41 34

Example/Backpack.xcodeproj/project.pbxproj
```diff
@@ -776,11 +777,6 @@
 				BUNDLE_LOADER = "$(TEST_HOST)";
 				CLANG_WARN_DEPRECATED_OBJC_IMPLEMENTATIONS = YES;
 				CLANG_WARN_OBJC_IMPLICIT_RETAIN_SELF = YES;
-				FRAMEWORK_SEARCH_PATHS = (
-					"$(SDKROOT)/Developer/Library/Frameworks",
-					"$(inherited)",
-					"$(DEVELOPER_FRAMEWORKS_DIR)",
-				);
```
shaundon: This is to fix this error:

screen shot 2018-09-12 at 09 53 36

PR #46, opened 2018-09-07
[BPK-1875] Add colors view

georgegillams:


georgegillams:
This is what I've got so far. Getting a bit stuck when it comes to auto-layout, but apart from that I think this is ok. It even chooses an appropriate label color based on

 
screen shot 2018-09-10 at 14 08 00

Example/Backpack/Views/CollectionView Cells/ColorPreviewCollectionViewCell.swift
```diff
@@ -0,0 +1,86 @@
...
+        
+        self.setup()
+    }
+    
+    required init?(coder aDecoder: NSCoder) {
+        fatalError("NSCoding not supported")
+    }
+
+    public static func referenceSize() -> CGSize {
+        // let mockLabel = Backpack.Label(fontStyle: BPKFontStyle.textLgEmphasized)
```
georgegillams: This should be calculated rather than hard coded :)

Example/Backpack/View Controllers/ColorsViewController.swift
```diff
@@ -0,0 +1,101 @@
...
+        collectionView?.delegate = self
+        collectionView?.dataSource = self
+        
+        guard let layout = collectionView?.collectionViewLayout as? UICollectionViewFlowLayout else {
+            fatalError("ColorsViewController collectionView must be using UICollectionViewFlowLayout")
+        }
+        
+        collectionView?.contentInset = UIEdgeInsets(top: BPKSpacingBase, left: BPKSpacingBase, bottom: BPKSpacingBase, right: BPKSpacingBase)
+        
+        layout.estimatedItemSize = ColorPreviewCollectionViewCell.referenceSize()
```
k0nserv: We should call the static method `ColorPreviewCollectionViewCell` estimatedItemSize for consistency

Example/Backpack/View Controllers/ColorsViewController.swift
```diff
@@ -0,0 +1,101 @@
...
+    }
+}
+
+// MARK: datasource UICollectionViewDataSource
```
k0nserv: Can we remove the `datasource` bit here? It's redudnant given that the full name of the protocol is spelled out

Example/Backpack/View Controllers/ColorsViewController.swift
```diff
@@ -0,0 +1,101 @@
...
+            // Customize headerView here
+            switch(indexPath.section){
+            case 0:
+                headerView.name = "Primary colors"
+            case 1:
+                headerView.name = "Secondary colors"
+            case 2:
+                headerView.name = "Grays"
+            default:
+                headerView.name = "No such section"
```
k0nserv: We should `fatalError` here so it's clear to developer that they need add another case

Example/Backpack/View Controllers/ColorsViewController.swift
```diff
@@ -0,0 +1,101 @@
...
+            
+            return headerView
+        }
+        fatalError("Headers are the only supplimental elements available")
+    }
+}
+
+extension ColorsViewController: UICollectionViewDelegateFlowLayout {
+    func collectionView(_ collectionView: UICollectionView, layout collectionViewLayout: UICollectionViewLayout, referenceSizeForHeaderInSection section: Int) -> CGSize {
+      return ColorPreviewCollectionViewHeader.referenceSize(collectionView: collectionView, text: "Some colors")
```
k0nserv: Could we use the actual text for the given section instead of `"Some colors"`? Maybe we can adjust the data structure a bit fit the section header names in there too? Also we should really be using NSLocalizedString for the headers in case we ever whish to localize the app

Example/Backpack/Views/CollectionView Cells/ColorPreviewCollectionViewCell.swift
```diff
@@ -0,0 +1,85 @@
...
+        contentView.addSubview(colorView)
+        contentView.addSubview(label)
+        
+        contentView.translatesAutoresizingMaskIntoConstraints = false
+        label.translatesAutoresizingMaskIntoConstraints = false
+        colorView.translatesAutoresizingMaskIntoConstraints = false
+        
+        colorView.widthAnchor.constraint(equalToConstant: 100).isActive = true
+        
+        contentView.leadingAnchor.constraint(equalTo: colorView.leadingAnchor).isActive = true
```
k0nserv: Can we use `NSLayoutConstraint.activate` for these two blocks?

Example/Backpack/Views/CollectionView Headers/ColorPreviewCollectionViewHeader.swift
```diff
@@ -0,0 +1,62 @@
...
+    private let label: Backpack.Label
+    
+    override init(frame: CGRect) {
+        self.label = Backpack.Label(fontStyle: .textXlEmphasized)
+        
+        super.init(frame: frame)
+        
+        self.setup()
+    }
+
```
k0nserv: Can we keep one empty line between items in classes?

Example/Backpack/Views/CollectionView Headers/ColorPreviewCollectionViewHeader.swift
```diff
@@ -0,0 +1,62 @@
...
+    required init?(coder aDecoder: NSCoder) {
+        fatalError("NSCoding not supported")
+    }
+    
+    public static func referenceSize(collectionView: UICollectionView, text: String) -> CGSize {
+        let mockLabel = Backpack.Label(fontStyle: BPKFontStyle.textLgEmphasized)
+        mockLabel.text = text
+        mockLabel.sizeToFit()
+        
+        return CGSize(width: collectionView.frame.width, height: mockLabel.frame.height + 8+4)
```
k0nserv: Instead of using hard coded values we can reference the spacing values. Further we could move these to private statics values so that we don't have to change in two places if we decide to alter it in the future

Example/Backpack/Views/CollectionView Cells/ColorPreviewCollectionViewCell.swift
georgegillams: Why should it be different for these two compared with all the others? Should we not stick to one or the other?

Example/Backpack/View Controllers/ColorsViewController.swift
georgegillams: Not sure what you mean here exactly

PR #49, opened 2018-09-10
[No JIRA] Add nonnull decision

k0nserv:


matthewdavidson:
You need to add a link to this in [`decisions/README.md`](https://github.com/Skyscanner/backpack-ios/blob/master/decisions/README.md)

k0nserv:
Done

decisions/assume-non-null-by-default.md
```diff
@@ -0,0 +1,44 @@
+# Assume non-null by default in Objective-C
+
+## TL;DR
+
+Swift uses a default model where nullability has to be explicitly expressed, in Objective-C the default is the opposite. To avoid confusion and helps us ensure null-saftiy we invert the behaviour in Objecitve-C to assume all pointers are non-null unless otherwise specified.
```
shaundon: ```helps us ensure null-saftiy```

should be

```help us ensure null-safety```

Comment (shaundon) [reply decisions/assume-non-null-by-default.md]: Also ```Objecitve-C``` is misspelled.

File: decisions/assume-non-null-by-default.md [hunk 2]
```diff
@@ -0,0 +1,44 @@
...
+@end
+
+@implementation
+@end
+NS_ASSSUME_NONNULL_END
+```
+
+## Thinking
+
+Aligning with the Swift default will make our code nicer to use from Swift and reduce typing as a most pointers are `nonnull` rather than `nullable`. Furhter more the compiler will help us more by preventing us from breaking this contract than if pointers were `nullable` by default.
```
shaundon: ```Furhter more``` should be ```Furthermore,```

decisions/assume-non-null-by-default.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Anything else
```
shaundon: You don't need this block.

decisions/assume-non-null-by-default.md
k0nserv: Fucking english 😂

decisions/assume-non-null-by-default.md
k0nserv: Fair point

PR #52, opened 2018-09-17
[BPK-1838] Add deploy step to backpack-ios travis build

georgegillams:
Difficult to test this properly until we merge and perform a release.

Things left to do after this is merged:
 - [x] Prevent deploying twice (there are two build configs, and each one runs the travis script)

It skips the docs deployment on non-master branches (and also if there's no tag)
https://travis-ci.org/Skyscanner/backpack-ios/jobs/429660677#L1105

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,8 +35,9 @@ script:
   - 'echo "$TRAVIS_TAG"'
   - '[ "$MAIN_STAGE" = true ] && ./scripts/check-pristine-state package-lock.json || true'
   - '[ "$MAIN_STAGE" = true ] && [ "$TRAVIS_TAG" ] && mkdir -p ./docs/versions/$TRAVIS_TAG'
-  - '[ "$MAIN_STAGE" = true ] && [ "$TRAVIS_TAG" ] && echo "SIMPLER" > ./docs/index.html'
-  - '[ "$MAIN_STAGE" = true ] && [ "$TRAVIS_TAG" ] && mkdir -p ./docs/versions/latest && echo "SIMPLER" > ./docs/versions/latest/index.html'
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -31,13 +31,30 @@ install:
...
     - /^greenkeeper/.*$/
+
+deploy:
+  provider: pages
+  skip_cleanup: true
+  github_token: $IOS_DOCS_DEPLOY_TOKEN
+  # on:
+  #  branch: master
+  # TODO CHECK FOR TAGS
+  local_dir: docs/
```
k0nserv: Can we make this `versions/{latest_version}`? So that we keep the history and can point to old releases?

.travis.yml
```diff
@@ -31,13 +31,32 @@ install:
 before_script: '[ "$MAIN_STAGE" = true ] && greenkeeper-lockfile-update || true'
 script:
   - npm run build
+  - rake "docs['./docs/versions/$TRAVIS_TAG']"
+  - echo "<!DOCTYPE html> <html> <script type=\"text/javascript\"> window.location = \"/ios/versions/latest\"; </script> </html>" > ./docs/index.html
+  - mkdir ./docs/versions/latest && echo "<!DOCTYPE html> <html> <script type=\"text/javascript\"> window.location = \"/ios/versions/$TRAVIS_TAG\"; </script> </html>" > ./docs/versions/latest/index.html
```
k0nserv: We don't wanna run any of these when `MAIN_STAGE` is not `true`

georgegillams:
Cool. I think I've added checks so that this only happens on `MAIN_STAGE` when there's a `TRAVIS_TAG`, but if I'm entirely honest, I have no idea what I am doing

https://media3.giphy.com/media/SRx5tBBrTQOBi/200w.gif?cid=c94812d05ba0da4648763066771cf3aa

PR #53, opened 2018-09-17
[BPK-1875] Implement icon component `BPKIcon`

k0nserv:


Backpack/Icon/Classes/BPKIcon.m
```diff
@@ -0,0 +1,236 @@
...
+
+
+        NSURL *url = [bundle URLForResource:@"iconMapping" withExtension:@"json"];
+        NSAssert(url, @"`iconMapping.json` must be present in the bundle.");
+        if (url == nil) {
+            _iconMapping = nil;
+            return;
+        }
+
+        NSData *data = [NSData dataWithContentsOfURL:url];
```
tzahola: Could have used `dataWithContentsOfURL:options:error:` for more accurate error reporting

Backpack/Icon/Classes/BPKIcon.m
```diff
@@ -0,0 +1,236 @@
...
+    NSData *fontData = [NSData dataWithContentsOfURL:url];
+
+    CGDataProviderRef provider = CGDataProviderCreateWithCFData((CFDataRef)fontData);
+    CGFontRef font = CGFontCreateWithDataProvider(provider);
+
+    if (font) {
+        CFErrorRef error = NULL;
+        if (CTFontManagerRegisterGraphicsFont(font, &error) == NO) {
+            CFStringRef errorDescription = CFErrorCopyDescription(error);
+            @throw [NSException exceptionWithName:NSInternalInconsistencyException reason:(__bridge NSString *)errorDescription userInfo:@{ NSUnderlyingErrorKey: (__bridge NSError *)error }];
```
tzahola: `errorDescription` has +1 retain count due to the copy-rule, so `__bridge` should be `__bridge_transfer`. Of course, we're the process is about the crash here, so it doesn't really matter :D

Backpack/Icon/Classes/BPKIcon.m
```diff
@@ -0,0 +1,236 @@
...
+        }
+
+        CFRelease(font);
+    }
+
+    CFRelease(provider);
+}
+
++ (UIImage *)iconNamed:(NSString *)name color:(UIColor *)color size:(BPKIconSize)size {
+    NSAssert([NSThread isMainThread], @"iconNamed:color:size: should only be called on the main thread.");
```
tzahola: Is there a legitimate reason to prevent loading images on a background thread? 

- `NSCache` is thread-safe
- `UIGraphicsContext` is thread-safe (*"This function may be called from any thread of your app."* - https://developer.apple.com/documentation/uikit/1623912-uigraphicsbeginimagecontextwitho?language=objc)

Backpack/Icon/Classes/BPKIcon.m
```diff
@@ -0,0 +1,236 @@
...
+            return CGSizeMake(16, 16);
+        case BPKIconSizeLarge:
+            return CGSizeMake(24, 24);
+        default:
+            NSAssert(NO, @"Unsupported icon size");
+            return CGSizeMake(16, 16);
+    }
+}
+
++ (NSString *)stringForUnicodeCodepoint:(nullable NSString *)codepoint {
```
tzahola: Why don't you store the Unicode characters as-is in the JSON file? I.e. `"EA87" --> "\uEA87"`

k0nserv:
> Beyond offering a more convenient API, Asset Catalogs let apps take advantage of app thinning, resulting in smaller installations that are optimized for each user’s device.

Is that just if the metadata allows Apple to exclude certain assets for specific devices? That wouldn't apply in this specific case AFAIK

Backpack/Icon/Classes/BPKIcon.m
k0nserv: Sounds good yeah

Backpack/Icon/Classes/BPKIcon.m
k0nserv: Primarily because I don't wish to change the json file, it's shared with React Native. Ideally we would just store a raw number instead for Objective-C

Backpack/Icon/Classes/BPKIcon.m
k0nserv: I just stole this code from elsewhere but like you say it will crash here. I am not sure if crashing is the best option. `NSAssert` + falling back to SF and rendering http://www.fileformat.info/info/unicode/char/25a1/index.htm might be better

Backpack/Icon/Classes/BPKIcon.m
tzahola: Oh, ok, it makes sense then.

Backpack/Icon/Classes/BPKIcon.m
tzahola: Yeah, the assert + fallback is more production-friendly IMO.

k0nserv: Updated

gulpfile.js
```diff
@@ -292,5 +292,11 @@ gulp.task('template', () => {
   return merge2(streams).pipe(gulp.dest(PATHS.output));
 });
 
-gulp.task('default', ['template']);
+gulp.task('copy-icon-font', () => {
+  gulp
+    .src('node_modules/bpk-svgs/dist/font/{BpkIcon.ttf,iconMapping.json}')
+    .pipe(gulp.dest(path.join(PATHS.output, 'Icon', 'Assets')));
```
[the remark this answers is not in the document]
k0nserv: That's for code so I don't think the pattern is as importnat here, we can always change it later tho

PR #58, opened 2018-09-20
[BPK-1838] add docs build to travis

georgegillams:


.travis.yml
```diff
@@ -32,6 +32,10 @@ before_script: 'if [[ "$MAIN_STAGE" = true ]]; then greenkeeper-lockfile-update;
 script:
   - npm run build
   - 'if [[ "$MAIN_STAGE" = true ]]; then ./scripts/check-pristine-state package-lock.json; fi'
+  - 'if [[ "$MAIN_STAGE" = true && -n "$TRAVIS_TAG" ]]; then mkdir -p ./docs/versions/$TRAVIS_TAG; fi'
+  - 'if [[ "$MAIN_STAGE" = true && -n "$TRAVIS_TAG" ]]; then rake "docs[\"./docs/versions/$TRAVIS_TAG\"]"; fi'
+  - 'if [[ "$MAIN_STAGE" = true && -n "$TRAVIS_TAG" ]]; then echo "<!DOCTYPE html> <html> <script type=\"text/javascript\"> window.location = \"/ios/versions/latest\"; </script> </html>" > ./docs/index.html; fi'
```
k0nserv: Any reason we aren't just using https://css-tricks.com/snippets/html/meta-refresh/ which works for non-JS clients?

georgegillams: I did not know that was a thing. Will check it out now!

PR #65, opened 2018-09-25
[No JIRA] Add iOS 12 to test matrix

k0nserv:


georgegillams:
Snapshot tests are failing on the new config 😢

k0nserv:
They shouldn't be running just yet, I've disabled them for now

PR #68, opened 2018-09-26
Generate icon names

tevelee:


Backpack/Icon/Classes/BPKIconName.h
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#import <Foundation/Foundation.h>
+
+typedef NSString * NS_SWIFT_NAME(IconName) BPKIconName NS_EXTENSIBLE_STRING_ENUM;
```
k0nserv: I know there's nothing in this file that requires it be generated, but could we generate it anyway so it ends up next to the implementation file?

k0nserv: Or just move this typedef to `BPKIconNames.h` maybe?

gulpfile.js
```diff
@@ -58,7 +60,9 @@ const LEGIBLE_NAMES = [
   { identifier: 'None', legibleName: 'none' },
 ];
 
-const format = s => s[0].toUpperCase() + _.camelCase(s.substring(1));
+function capitalizeFirstLetter(string) {
```
k0nserv: This isn't needed because `capitalize` from `lodash` can be used

k0nserv: Further why are we changing this at all? It doesn't seem like it is needed

gulpfile.js
```diff
@@ -299,5 +303,33 @@ gulp.task('copy-icon-font', () => {
     .pipe(gulp.dest(path.join(PATHS.output, 'Icon', 'Assets')));
 });
 
-gulp.task('default', ['template', 'copy-icon-font']);
+gulp.task('generate-icon-names', () => {
```
k0nserv: This should really be part of the `template` task imo

package.json
```diff
@@ -35,6 +35,7 @@
     "del": "^3.0.0",
     "eslint-config-skyscanner-with-prettier": "^0.3.0",
     "eslint_d": "^7.1.1",
+    "fs": "0.0.1-security",
```
k0nserv: This looks like it shouldn't be included

Backpack/Icon/Classes/BPKIconName.h
tevelee: since both BPKIconNames.h and BPKIcon.h needed it (independently from each other) I thought it's a good idea to put it in a separate file

gulpfile.js
tevelee: I used it in a helper method (where I transform the icon names into constants), and the same logic was there in two different methods, so I just used `capitalizeFirstLetter` for both

gulpfile.js
tevelee: true dat, I'll move it there

package.json
tevelee: I read the json file using fs (to parse the token names)

k0nserv: `fs` is builtin to node. This is probably something completely different. Please remove it

gulpfile.js
tevelee: capitalizeFirstLetter seemed to be a better suited name for the use case, but I rename it back if you want to

PR #83, opened 2018-10-05
[BPK-1960] Fix card bugs

georgegillams:


Backpack/Card/Classes/BPKCard.m
```diff
@@ -28,6 +28,7 @@
 @interface BPKCard()
 
 - (void)setupWithPadded:(BOOL)padded;
+@property(nonatomic) UIView *innerView;
```
k0nserv: Can we specify `strong` and call this contentView?

Backpack/Card/Classes/BPKCard.m
```diff
@@ -65,9 +66,9 @@ - (instancetype)initWithPadded:(BOOL)padded {
 
 - (void)setPadded:(BOOL)padded {
     if (padded) {
-        self.layoutMargins = UIEdgeInsetsMake(BPKSpacingBase, BPKSpacingBase, BPKSpacingBase, BPKSpacingBase);
+        self.innerView.layoutMargins = UIEdgeInsetsMake(BPKSpacingBase, BPKSpacingBase, BPKSpacingBase, BPKSpacingBase);
```
k0nserv: I would execpt us to still set the `layoutMaring` on `self` here and have the innerView constrianted to the layout guides of `self`

Backpack/Card/Classes/BPKCard.h
```diff
@@ -32,6 +32,11 @@ NS_SWIFT_NAME(Card) IB_DESIGNABLE @interface BPKCard: UIControl
  */
 @property (nonatomic, assign, getter=isPadded) IBInspectable BOOL padded;
 
+/**
+ * The view to display within the card.
+ */
+@property(nonatomic, assign, nullable) UIView *subview;
```
tevelee: I was thinking, `contentView` might be a better name. Would align more with UIKit standard naming, for example on UICollectionViewCell.h:
```objc
@property (nonatomic, readonly) UIView *contentView; // add custom subviews to the cell's contentView
```

k0nserv: Possibly, it needs to make sense for the divided card too

PR #84, opened 2018-10-08
[BPK-1919] Update pod API for Swift consumption

georgegillams:
**Once this is merged, we should release `backpack-ios v1.0.0`**

I can't wait

### DIFF
--- a/Backpack/Badge/Classes/BPKBadge.h
+++ b/Backpack/Badge/Classes/BPKBadge.h
@@ -55,7 +55,7 @@ typedef NS_ENUM(NSUInteger, BPKBadgeType) {
 };
 
 NS_ASSUME_NONNULL_BEGIN
-IB_DESIGNABLE @interface BPKBadge: UIView
+NS_SWIFT_NAME(Badge) IB_DESIGNABLE @interface BPKBadge: UIView
 - (instancetype)initWithFrame:(CGRect)frame NS_DESIGNATED_INITIALIZER;
 
 /**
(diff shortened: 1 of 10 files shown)


k0nserv:
+ `BPKDividedCard`'s `-setPrimarySubview:secondarySubview:` needs to be adjusted.
+ `BPKColor`s  `+ (UIColor *)clear` should be converted to a class property
+ `BPKFont`'s `+fontWithStyle:` should be renamed `makeFont(style:)`

k0nserv:
Another outstanding question is what do we do about the enums we currently have. They are currently being imported as `BPK{Name}`, however in Swift they are likely to very often be referenced just as `.variant` so the type name might not matter that much. The same thing applies to spacing values and other constants. `BPKSpacingSm` will be a bit weird in Swift code, but simply calling it `spacingSm` isn't that nice either maybe we should consider calling such constants `BackpackSpacingSm`, but that's very wordy

georgegillams:
> https://github.com/Skyscanner/backpack-ios/pull/84#issuecomment-428120008

@k0nserv I've addressed these now 👍

shaundon:
Looks like Salesforce's design system in Swift doesn't use prefixes at all, FWIW:

```
let backgroundColor = UIColor.sldsBackgroundColor(.colorBackground)
let borderColor = UIColor.sldsBorderColor(.colorBorderBrand)
let fillColor = UIColor.sldsFill(.brand)
let textColor = UIColor.sldsTextColor(.colorTextDefault)
```

Taken from https://www.lightningdesignsystem.com/platforms/ios/#swift-usage

k0nserv:
Those are categories on `UIColor` though. I did internal research and people were not in favour of using categories based on that research

PR #85, opened 2018-10-08
[BPK-1968] Implement dialog component

k0nserv:
Will follow up with tests and docs.


Backpack/Dialog/Classes/BPKDialogController.h
```diff
@@ -0,0 +1,61 @@
...
+
+- (instancetype)init __attribute__((unavailable("init not available, use alertControllerWithTitle:message:style:shadow:headColor:iconImage: instead")));
++ (instancetype)dialogControllerWithTitle:(NSString *)title
+                                  message:(NSString *)message
+                                    style:(BPKDialogControllerStyle)style
+                              shadowStyle:(BPKDialogControllerShadowStyle)shadowStyle
+                                headColor:(UIColor *)headColor
+                                iconImage:(UIImage *)iconImage;
+
+- (void)addPresentingKeyFrameContentAnimationWithRelativeStartTime:(double)startTime relativeDuration:(double)duration;
```
k0nserv: Theses two methods should really be protected, but since Objective-c doesn't have anything like it they have to be public. I guess one option would be to not declare them and instead call them via `performSelector:`

Backpack/Dialog/Classes/BPKDialogController.h
```diff
@@ -0,0 +1,61 @@
...
+typedef NS_ENUM(NSInteger, BPKDialogControllerShadowStyle) {
+    BPKDialogControllerShadowStyleNoShadow = 0,
+    BPKDialogControllerShadowStyleShadow
+};
+
+NS_SWIFT_NAME(DialogController) @interface BPKDialogController: UIViewController<BPKDialogViewDelegate, UIViewControllerTransitioningDelegate>
+@property(nonatomic, assign) BPKDialogControllerShadowStyle shadowStyle;
+@property(nonatomic, assign) BPKDialogControllerStyle style;
+
+- (void)setScrimAlpha:(double)scrimAlpha;
```
k0nserv: This is also not really intended for public use, but it needs to be declared so the animator can use it

Backpack/Dialog/Classes/BPKDialogController.h
```diff
@@ -0,0 +1,61 @@
...
+- (void)addButtonAction:(BPKDialogButtonAction *)action NS_SWIFT_NAME(addButtonAction(_:));
+- (void)addScrimAction:(BPKDialogScrimAction *)action NS_SWIFT_NAME(addScrimAction(_:));
+
+- (instancetype)init __attribute__((unavailable("init not available, use alertControllerWithTitle:message:style:shadow:headColor:iconImage: instead")));
++ (instancetype)dialogControllerWithTitle:(NSString *)title
+                                  message:(NSString *)message
+                                    style:(BPKDialogControllerStyle)style
+                              shadowStyle:(BPKDialogControllerShadowStyle)shadowStyle
+                                headColor:(UIColor *)headColor
```
k0nserv: I don't like the name `headColor`, but I can't really think of anything better.

Backpack/Dialog/Classes/BPKDialogController.m
```diff
@@ -0,0 +1,224 @@
...
+    self.scrimView.userInteractionEnabled = YES;
+    self.scrimView.backgroundColor = BPKColor.gray900;
+    self.scrimView.alpha = 0.5;
+    self.scrimView.translatesAutoresizingMaskIntoConstraints = NO;
+    [self.scrimView addGestureRecognizer:tapGesture];
+
+    self.dialogView = [[BPKDialogView alloc] initWithFrame:CGRectZero];
+    self.dialogView.translatesAutoresizingMaskIntoConstraints = NO;
+    self.dialogView.delegate = self;
+    [self.dialogView setTitle:self.titleText];
```
k0nserv: These should be normal properties

Backpack/Dialog/Classes/BPKDialogButtonAction.m
```diff
@@ -0,0 +1,61 @@
...
+
+    return self;
+}
+
++ (instancetype _Nonnull)actionWithTitle:(NSString *)title style:(BPKButtonStyle)style handler:(BPKDialogButtonActionHandler)handler {
+    return [[self alloc] initWithButtonStyle:style title:title actionHandler:handler];
+}
+
+- (id)copyWithZone:(nullable NSZone *)zone {
+    BPKDialogButtonAction *copy = [[[self class] allocWithZone:zone] init];
```
tevelee: here you can use the already existing intialiser. (after you called allocWithZone)
`return [[self.class allocWithZone:zone] initWithTitle:self.title.copy style:self.style handler:self.handler.copy];`
This way you can discard the ivar arrow magic down there

Backpack/Dialog/Classes/BPKDialogController.m
```diff
@@ -0,0 +1,224 @@
...
+                              shadowStyle:(BPKDialogControllerShadowStyle)shadowStyle
+                                headColor:(UIColor *)headColor
+                                iconImage:(UIImage *)iconImage {
+    return [[self alloc] initWithTitle:title message:message style:style shadowStyle:shadowStyle headColor:headColor iconImage:iconImage];
+}
+
+- (void)setupViews {
+    UITapGestureRecognizer *tapGesture = [[UITapGestureRecognizer alloc] initWithTarget:self action:@selector(scrimTapped:)];
+    
+    self.scrimView = [UIView new];
```
tevelee: sometimes you use alloc + initWithFrame:CGRectZero, sometimes new

Backpack/Dialog/Classes/BPKDialogController.m
```diff
@@ -0,0 +1,224 @@
...
+    }
+}
+
+- (void)addViews {
+    [self.view addSubview:self.scrimView];
+    [self.view addSubview:self.dialogView];
+}
+
+- (void)scrimTapped:(UITapGestureRecognizer *)gestureRecognizer {
+    if( gestureRecognizer.state == UIGestureRecognizerStateRecognized ) {
```
tevelee: inconsistent coding style `if( xxx )` vs `if (xxx)`everywhere else

Backpack/Dialog/Classes/BPKDialogControllerAnimator.m
```diff
@@ -0,0 +1,92 @@
...
+}
+
+- (NSTimeInterval)transitionDuration:(nullable id<UIViewControllerContextTransitioning>)transitionContext {
+    return [[self class] duration];
+}
+
+#pragma mark - Private
+
++ (NSTimeInterval)duration {
+    return 0.35;
```
tevelee: don't we have a backpack constant for this?

Backpack/Dialog/Classes/BPKDialogIconView.m
```diff
@@ -0,0 +1,116 @@
...
+- (void)setupViews;
+
++ (CGSize)iconContainerSize;
+@end
+
+@implementation BPKDialogIconView
+
+- (instancetype)init {
+    CGSize size = [[self class] viewSize];
+    self = [super initWithFrame:CGRectMake(0, 0, size.width, size.height)];
```
tevelee: alternatively, you can say `(CGRect){CGPointZero, size}`

Backpack/Dialog/Classes/BPKDialogView.m
```diff
@@ -0,0 +1,221 @@
...
+                                              [self.buttonStackView.superview.layoutMarginsGuide.trailingAnchor constraintEqualToAnchor:self.buttonStackView.trailingAnchor constant:(BPKSpacingXl - BPKSpacingLg)],
+                                              [self.buttonStackView.superview.layoutMarginsGuide.bottomAnchor constraintEqualToAnchor:self.buttonStackView.bottomAnchor]
+                                              ]];
+}
+
+- (void)layoutSubviews {
+    [super layoutSubviews];
+
+    if (self.hasShadow) {
+        [[BPKShadow shadowLg] applyToLayer:self.layer];
```
tevelee: Have you considered setting shadowPath on the layer (to improve performance)?

https://developer.apple.com/documentation/quartzcore/calayer/1410771-shadowpath
Specifying an explicit path usually improves rendering performance.

Backpack/Dialog/Classes/BPKDialogView.m
```diff
@@ -0,0 +1,221 @@
...
+    return self.titleLabel.text;
+}
+
+
+- (void)setDescription:(NSString *_Nullable)description {
+    self.descriptionLabel.text = description;
+}
+
+- (NSString *_Nullable)description {
+    return self.descriptionLabel.text;
```
tevelee: overriding description will stop having the memory address when printing it

Backpack.podspec
```diff
@@ -89,6 +89,15 @@ Pod::Spec.new do |s|
       ss.dependency 'Backpack/RTLSupport'
   end
 
+  s.subspec 'Dialog' do |ss|
+    ss.source_files = 'Backpack/Dialog/Classes/**/*.{h,m}'
+    ss.public_header_files = 'Backpack/Dialog/Classes/**/*.h'
+    ss.dependency 'Backpack/Button'
+    ss.dependency 'Backpack/Shadow'
+    ss.dependency 'Backpack/Radii'
+    ss.dependency 'Backpack/Color'
```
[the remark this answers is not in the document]
k0nserv: Good catch, we should figure out a away to make this more apparent when developing

Backpack/Dialog/Classes/BPKDialogButtonAction.h
```diff
@@ -0,0 +1,36 @@
...
+
+NS_ASSUME_NONNULL_BEGIN
+@class BPKDialogButtonAction;
+typedef void (^BPKDialogButtonActionHandler)(BPKDialogButtonAction *);
+
+NS_SWIFT_NAME(DialogButtonAction) @interface BPKDialogButtonAction : NSObject <NSCopying>
+
+@property (nonatomic, assign, readonly) BPKButtonStyle style;
+@property (nonatomic, copy, readonly) NSString *title;
+@property (nonatomic, strong, readonly) BPKDialogButtonActionHandler handler;
```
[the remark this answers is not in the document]
k0nserv: Yeah I tend to prefer being explicit about those personally, but I think we should discuss it as a team. The defaults are `readwrite`, `atomic`, and `assign` see https://stackoverflow.com/questions/7827237/what-are-the-default-attributes-for-objective-c-properties#7827261. I am open to changing our conventions to only specify attributes that diverge from the default. cc @zsbee, @shaundon, @georgegillams what do you think?

Backpack/Dialog/Classes/BPKDialogButtonAction.m
```diff
@@ -0,0 +1,61 @@
...
+- (instancetype _Nonnull)initWithButtonStyle:(BPKButtonStyle)style title:(NSString *)title actionHandler:(BPKDialogButtonActionHandler)handler;
+@end
+
+@implementation BPKDialogButtonAction
+
+- (instancetype _Nonnull)initWithButtonStyle:(BPKButtonStyle)style title:(NSString *)title actionHandler:(BPKDialogButtonActionHandler)handler {
+    self = [super init];
+
+    if (self) {
+        self.style = style;
```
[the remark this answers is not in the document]
k0nserv: I tend to favour never using ivars outside of overridden setters and getters which is why I have re-declared them as `readwrite`. I've been meaning to write this up as a proposed descision, but it might be worth discussing further too

Backpack/Dialog/Classes/BPKDialogButtonAction.m
k0nserv: I like it 👍

Backpack/Dialog/Classes/BPKDialogController.m
k0nserv: I think I use `initWithFrame:CGRectZero` pretty consistently, but this code in particular was written by both me and @zsbee which is why it's a bit inconsistent.

Backpack/Dialog/Classes/BPKDialogController.m
k0nserv: Good catch, should be tight

Backpack/Dialog/Classes/BPKDialogControllerAnimator.m
k0nserv: Nope, this is based the default timings UIKit uses for their transitions.

Backpack/Dialog/Classes/BPKDialogView.m
k0nserv: Really cood catch 👍

Backpack/Dialog/Classes/BPKDialogView.m
k0nserv: Not really, @zsbee wrote this code specific code, maybe here tried it?

k0nserv: I've updated it to use a computed path for the contentView + iconView

Backpack/Dialog/Classes/BPKDialogButtonAction.h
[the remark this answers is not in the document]
georgegillams: I also prefer explicit, but I think it's more important to align as a team / company

Backpack/Dialog/Classes/BPKDialogController.h
k0nserv: Added leading underscores to these and mentioned in the doc comments that they are not meant for public use

Backpack/Dialog/Classes/BPKDialogController.h
k0nserv: Changed to `iconBackgroundColor`

Backpack/Dialog/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+BPKDialogScrimAction *scrimAction = [BPKDialogScrimAction actionWithHandler:^(BOOL didDismiss) {
+    //
+} shouldDismiss:YES];
+
+BPKDialogButtonAction *buttonAction = [BPKDialogButtonAction actionWithTitle:@"Continue" style:BPKButtonStylePrimary handler:^{
+    //
+}];
+
+BPKDialogController *dialogController = [BPKDialogController dialogControllerWithTitle:@"Such Wow!"
```
georgegillams: We should probably update the readme example to use the copy we applied to the stories

Backpack/Dialog/Classes/BPKDialogIconView.h
```diff
@@ -0,0 +1,34 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#import <UIKit/UIKit.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface BPKDialogIconView : UIView
```
georgegillams: Should this be `NS_SWIFT_NAME(DialogIconView)`?
Or did we omit it as it's not for consumer use?

k0nserv: Yeah consumers aren't  using this so we don't consider it part of the public interface

Backpack/Dialog/README.md
k0nserv: Sounds like a good idea yeah

k0nserv:
This is ready to merge now, just needs ✅

Backpack/Dialog/Classes/BPKDialogControllerAnimator.m
shaundon: I reckon we should document that - quick comment above it or something? Principle of least surprise etc.

PR #87, opened 2018-10-15
Switch to Xcode 10

k0nserv:


shaundon:
Runs fine on my machine - only thing is this warning - is this on my end or is it something we've missed?

screen shot 2018-10-16 at 11 28 48

k0nserv:
> One question - why have the snapshot images changed? They look exactly the same to me

Tiny rendering difference between iOS 11 and iOS 12.

k0nserv:
> Runs fine on my machine - only thing is this warning - is this on my end or is it something we've missed?

Let me look at

k0nserv:
I don't get that warning locally

shaundon:
Ok, thanks for checking. Let's just assume it's my setup.

PR #89, opened 2018-10-15
[No JIRA] Add swiftlint

k0nserv:


.travis.yml
```diff
@@ -36,6 +36,7 @@ before_script: 'if [[ "$MAIN_STAGE" = true ]]; then greenkeeper-lockfile-update;
 script:
   - npm run build
   - 'if [[ "$MAIN_STAGE" = true ]]; then ./scripts/check-pristine-state package-lock.json; fi'
+  - 'if [[ "$MAIN_STAGE" = true ]]; then Example/Pods/SwiftLint/swiftlint; fi'
```
georgegillams: Shouldn't this exit with `1` if there are warnings?

georgegillams: There doesn't seem to be a swiftlint argument for this, but we could run `swiftlint autocorrect` before we check pristine state?

k0nserv: Yes it should exit with `1` if there are warnings. I don't really wanna run `autocorrect` in CI

PR #90, opened 2018-10-22
[BPK-1985] Implement new type ramp

k0nserv:


Backpack/Font/Classes/Generated/BPKFont.h
```diff
@@ -21,74 +21,194 @@ typedef NS_ENUM(NSUInteger, BPKFontStyle) {
...
+
+/**
+ * The `textCaps` text style from the Backpack typograhpy stack.
+ *
+ * @warning This method is deprecated and using the font does not correctly render text with the Skyscanner style.
+*/
++ (UIFont *)textCaps DEPRECATED_MSG_ATTRIBUTE("Use `attributedStringWithFontStyle:content` instead.");
+
+/**
+ * The `textCapsEmphasized` text style from the Backpack typograhpy stack.
```
shaundon:
```suggestion
 * The `textCapsEmphasized` text style from the Backpack typography stack.
```

Backpack/Font/Classes/Generated/BPKFont.h
```diff
@@ -21,74 +21,194 @@ typedef NS_ENUM(NSUInteger, BPKFontStyle) {
...
 /**
  * The `textBaseEmphasized` text style from the Backpack typograhpy stack.
+ *
+ * @warning This method is deprecated and using the font does not correctly render text with the Skyscanner style.
 */
-+ (UIFont *)textBaseEmphasized;
++ (UIFont *)textBaseEmphasized DEPRECATED_MSG_ATTRIBUTE("Use `attributedStringWithFontStyle:content` instead.");
+
+/**
+ * The `textCaps` text style from the Backpack typograhpy stack.
```
shaundon: I had to find _something_ to criticise 😉 

```suggestion
 * The `textCaps` text style from the Backpack typography stack.
```

Backpack/Font/Classes/Generated/BPKFont.h
```diff
@@ -21,74 +21,194 @@ typedef NS_ENUM(NSUInteger, BPKFontStyle) {
...
+
+/**
+ * The `textCapsEmphasized` text style from the Backpack typograhpy stack.
+ *
+ * @warning This method is deprecated and using the font does not correctly render text with the Skyscanner style.
+*/
++ (UIFont *)textCapsEmphasized DEPRECATED_MSG_ATTRIBUTE("Use `attributedStringWithFontStyle:content` instead.");
 
 /**
  * The `textLg` text style from the Backpack typograhpy stack.
```
shaundon:
```suggestion
 * The `textLg ` text style from the Backpack typography stack.
```

shaundon: There's more `typograhpy ` below because of copypasta but I won't highlight them all

shaundon:
The view controller for labels has a strange font weight and kerning, I think.

Here's the root view controller:

screen shot 2018-10-23 at 12 05 34

And here's the labels one:

screen shot 2018-10-23 at 12 05 37

k0nserv:
The label one actually uses our style which is why it looks a bit off. We should start switching them all over to that

k0nserv:
> Ok good plan. It looks to my eyes like it's using the emphasized one too, not sure if that's desirable. We can fix it later though

Is is emphasized. I checked with @willhg he liked it like that

Backpack/Font/Classes/Generated/BPKFont.m
```diff
@@ -38,55 +54,301 @@ + (UIFont * _Nullable)fontWithStyle:(BPKFontStyle)style {
...
+            case BPKFontStyleTextXxxl:
+              
+                return  nil;
+              
+            case BPKFontStyleTextXxxlEmphasized:
+              
+                return  nil;
+              
+            case BPKFontStyleTextXxxlHeavy:
+
```
shaundon: I'm all for letting the code breathe, but I think it might be breathing too much in this `switch` statement. Is there a reason it's so spaced out?

k0nserv: Generated code and nunjucks, not sure how to make it tighter

shaundon: Fair enough - it's not a big deal

georgegillams:
Shouldn't `badge` and `dialog` be updated to use the new type scale too?

k0nserv:
> Shouldn't badge and dialog be updated to use the new type scale too?

They are?

georgegillams:
> > Shouldn't badge and dialog be updated to use the new type scale too?
> 
> They are?

Oh because they use BPKLabel - I was being an idiot 🤦‍♂️

georgegillams:
regular

PR #96, opened 2018-10-26
[No JIRA] Upgrade iOSSnapshotTestCase to 4.0.0

k0nserv:


georgegillams:
Can't seem to get tests to pass with this change. Were you expecting that we would have to regenerate them?

k0nserv:
They passed locally 🤔

PR #107, opened 2018-11-07
[BPK-1975] Add chip component

georgegillams:
Still todo in subsequent PRs:
 - [ ] Update story to match RN
 - [ ] Add tests

screenshot 2018-11-09 at 14 37 50

### DIFF
--- a/Backpack/Chip/Classes/BPKChip.h
+++ b/Backpack/Chip/Classes/BPKChip.h
@@ -1,0 +1,52 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2018 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#import <Foundation/Foundation.h>
+#import <UIKit/UIKit.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ * `BPKChip` is a subclass of `UIControl` configured with Skyscanner style properties.
+ */
+NS_SWIFT_NAME(Chip) IB_DESIGNABLE @interface BPKChip: UIControl
+
+/**
+ * The title to display inside the chip.
+ */
+@property(nonatomic, copy, nullable) NSString *title;
+
+/**
+ * Creates a `BPKChip` with a decoder (typically when creating from Storyboards)
+ *
+ * @param aDecoder Decoder object to extract parameters from
+ * @return `BPKChip` instance.
+ */
+- (nullable instancetype)initWithCoder:(NSCoder *)aDecoder NS_DESIGNATED_INITIALIZER;
+
+/**
+ * Create a `BPKChip` with a given frame.
+ *
+ * @param frame The initial frame of the chip.
+ * @return `BPKChip` instance.
+ */
+- (instancetype)initWithFrame:(CGRect)frame NS_DESIGNATED_INITIALIZER;
+
+@end
+
+NS_ASSUME_NONNULL_END
--- a/Backpack/Chip/Classes/BPKChip.m
+++ b/Backpack/Chip/Classes/BPKChip.m
@@ -1,0 +1,180 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2018 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#import "BPKChip.h"
+
+#import <Backpack/Font.h>
+#import <Backpack/Label.h>
+#import <Backpack/Color.h>
+#import <Backpack/Shadow.h>
+#import <Backpack/Spacing.h>
+
+NS_ASSUME_NONNULL_BEGIN
+@interface BPKChip()
+@property(nonatomic) BPKLabel* titleLabel;
+@property(nonatomic) CALayer* tintLayer;
+
+@property(nonatomic) UIColor *textColor;
+@end
+
+@implementation BPKChip
+
+- (instancetype)initWithFrame:(CGRect)frame {
+    self = [super initWithFrame:frame];
+
+    if (self) {
+        [self setup];
+    }
+
+    return self;
+}
+
+- (nullable instancetype)initWithCoder:(NSCoder *)aDecoder {
+    self = [super initWithCoder:aDecoder];
+
+    if (self) {
+        [self setup];
+    }
+
+    return self;
+}
+
+- (void)setup {
+    self.tintLayer = [CALayer layer];
+    self.tintLayer.backgroundColor = BPKColor.gray600.CGColor;
+    self.tintLayer.opacity = 0;
+    [self.layer addSublayer:self.tintLayer];
+
+    self.titleLabel = [[BPKLabel alloc] initWithFrame:CGRectZero];
+    self.titleLabel.fontStyle = BPKFontStyleTextSm;
+    [self addSubview:self.titleLabel];
+
+    self.titleLabel.textAlignment = NSTextAlignmentCenter;
+
+    [self addTarget:self action:@selector(handleSingleTap:) forControlEvents:UIControlEventTouchUpInside];
+
+    [self setupConstraints];
+
+    BPKShadow *shadow = [BPKShadow shadowSm];
+    [shadow applyToLayer:self.layer];
+
+    [self updateStyle];
+}
+
+#pragma mark - Layout overrides
+
+- (void)layoutSubviews {
+    self.tintLayer.frame = self.bounds;
+    self.tintLayer.cornerRadius = self.bounds.size.height / 2;
+
+    self.layer.cornerRadius = self.bounds.size.height / 2;
+}
+
+#pragma mark - State setters
+
+- (void)setTitle:(NSString *_Nullable)title {
+    _title = [title copy];
+
+    [self updateTitle];
+}
+
+- (void)setSelected:(BOOL)selected {
+    [super setSelected:selected];
+    [self updateStyle];
+}
+
+- (void)setHighlighted:(BOOL)highlighted {
+    [super setHighlighted:highlighted];
+
+    [CATransaction begin];
+    [CATransaction setAnimationDuration:highlighted ? .2 : 0];
+    self.tintLayer.opacity = highlighted ? 0.2 : 0;
+    [CATransaction commit];
+}
+
+- (void)setEnabled:(BOOL)enabled {
+    [super setEnabled:enabled];
+    [self updateStyle];
+}
+
+#pragma mark - Layout
+
+- (void)setupConstraints {
+    CGFloat chipHorizontalSpacing = [self chipHorizontalSpacing];
+    CGFloat chipVerticalSpacing = [self chipVerticalSpacing];
+
+    self.titleLabel.translatesAutoresizingMaskIntoConstraints = NO;
+
+    [NSLayoutConstraint activateConstraints:
+     @[
+       [self.titleLabel.leadingAnchor constraintEqualToAnchor:self.leadingAnchor constant:chipHorizontalSpacing],
+       [self.titleLabel.topAnchor constraintEqualToAnchor:self.topAnchor constant:chipVerticalSpacing],
+       [self.trailingAnchor constraintEqualToAnchor:self.titleLabel.trailingAnchor constant:chipHorizontalSpacing],
+       [self.bottomAnchor constraintEqualToAnchor:self.titleLabel.bottomAnchor constant:chipVerticalSpacing]
+       ]
+     ];
+}
+
+#pragma mark - Updates
+
+- (void)updateStyle {
+    if(self.selected) {
+        self.backgroundColor = BPKColor.blue500;
+        self.textColor = BPKColor.white;
+    } else {
+        self.backgroundColor = BPKColor.white;
+        self.textColor = BPKColor.gray700;
+    }
+
+    if(!self.enabled){
+        self.backgroundColor = BPKColor.white;
+        self.textColor = BPKColor.gray100;
+    }
+
+    [self updateTitle];
+}
+
+- (void)updateTitle {
+    if(self.title == nil){
+        return;
+    }
+
+    self.titleLabel.text = self.title;
+    [self.titleLabel setTextColor:self.textColor];
+}
+
+#pragma mark - Actions
+
+- (void)handleSingleTap:(UITapGestureRecognizer *)recognizer {
+    [self setSelected:!self.selected];
+}
+
+#pragma mark - Helpers
+
+- (CGFloat)chipHorizontalSpacing {
+    return BPKSpacingBase;
+}
+
+- (CGFloat)chipVerticalSpacing {
+    return BPKSpacingMd;
+}
+
+@end
+
+NS_ASSUME_NONNULL_END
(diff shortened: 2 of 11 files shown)


Backpack/Chip/Classes/BPKChip.h
```diff
@@ -0,0 +1,59 @@
...
+ */
+NS_SWIFT_NAME(Chip) IB_DESIGNABLE @interface BPKChip: UIControl
+
+/**
+ * The title to display inside the chip.
+ */
+@property(nonatomic, copy, nullable) NSString *title;
+
+/**
+ * Creates a `BPKChip` with a decoder (typically when creating form Storyboards)
```
k0nserv: Copy-and-paste error

Backpack/Chip/Classes/BPKChip.h
```diff
@@ -0,0 +1,59 @@
...
+ *
+ * @param enabled BOOL the new value
+ */
+- (void)setEnabled:(BOOL)enabled;
```
k0nserv: Why are we declaring this at all? It's inherited from `UIControl` right?

Backpack/Chip/Classes/BPKChip.m
```diff
@@ -0,0 +1,161 @@
...
+ */
+
+#import "BPKChip.h"
+
+#import <Backpack/Font.h>
+#import <Backpack/Label.h>
+#import <Backpack/Color.h>
+#import <Backpack/Shadow.h>
+#import <Backpack/Spacing.h>
+#import <Backpack/UIView+BPKRTL.h>
```
k0nserv: You need a dependency on `Backpack/RTLSupport` if you want to use this import

Backpack/Chip/Classes/BPKChip.m
```diff
@@ -0,0 +1,161 @@
...
+    }
+    
+    return self;
+}
+
+- (void)setup {
+    self.initializing = YES;
+    
+    self.titleLabel = [[BPKLabel alloc]initWithFrame:CGRectZero];
+    [self.titleLabel setFontStyle:BPKFontStyleTextSm];
```
k0nserv: Prefer dot syntax for properties

Backpack/Chip/Classes/BPKChip.h
shaundon:
```suggestion
 * Creates a `BPKChip` with a decoder (typically when creating from Storyboards)
```

Backpack/Chip/Classes/BPKChip.h
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * Creates a `BPKChip` with a decoder (typically when creating form Storyboards)
```
shaundon:
```suggestion
 * Creates a `BPKChip` with a decoder (typically when creating from Storyboards)
```

Backpack/Chip/Classes/BPKChip.m
```diff
@@ -0,0 +1,161 @@
...
+    [self addSubview:self.titleLabel];
+    
+    self.titleLabel.textAlignment = NSTextAlignmentCenter;
+    
+    UITapGestureRecognizer *singleFingerTap =
+    [[UITapGestureRecognizer alloc] initWithTarget:self
```
k0nserv: Is there are reason we aren't using `addTarget:action:for:UIControlEventTouchUpInside` here?

Backpack/Chip/Classes/BPKChip.m
```diff
@@ -0,0 +1,161 @@
...
+- (void)setHighlighted:(BOOL)highlighted {
+    [super setHighlighted:highlighted];
+}
+
+#pragma mark - Layout
+
+- (void)layoutSubviews {
+    [super layoutSubviews];
+    
+    [self setConstraints];
```
k0nserv: This will re create all constraints everytime the view is laid out. This is pretty bad for performance. Can we create the constriant in `setup` instead?

Backpack/Chip/Classes/BPKChip.m
```diff
@@ -0,0 +1,161 @@
...
+    self.layer.cornerRadius = self.bounds.size.height / 2;
+    
+    BPKShadow *shadow = [BPKShadow shadowSm];
```
k0nserv: We don't need to apply this on every layout

Backpack/Chip/Classes/BPKChip.m
```diff
@@ -0,0 +1,161 @@
...
+    [shadow applyToLayer:self.layer];
+}
+
+- (void)setConstraints {
```
k0nserv: Can we call this `setupConstraints` instead?

Backpack/Chip/Classes/BPKChip.m
```diff
@@ -0,0 +1,161 @@
...
+    self.translatesAutoresizingMaskIntoConstraints = NO;
+    self.titleLabel.translatesAutoresizingMaskIntoConstraints = NO;
+        [[self.titleLabel.centerXAnchor constraintEqualToAnchor:self.centerXAnchor] setActive:YES];
+        [[self.titleLabel.centerYAnchor constraintEqualToAnchor:self.centerYAnchor] setActive:YES];
+}
+
+- (CGSize)intrinsicContentSize {
```
k0nserv: What happens if we specificy the top, leading, trailing and bottom constraints and skip overriding this property?

Backpack/Chip/Classes/BPKChip.m
```diff
@@ -0,0 +1,161 @@
...
+    CGSize labelSize = self.titleLabel.bounds.size;
+    return CGSizeMake(labelSize.width + 2*BPKSpacingBase, labelSize.height + 2*BPKSpacingMd);
+}
+
+#pragma mark - Updates
+
+- (void)updateStyle {
+    if(self.selected) {
+        self.backgroundColor = [BPKColor blue500];
```
k0nserv: We've converted the colors to class properties so this can be written as

```objective-c
self.backgroundColor = BPKColor.blue500
```

Which I think is nicer, happy to discuss it though

Backpack/Chip/Classes/BPKChip.m
```diff
@@ -0,0 +1,161 @@
...
+        self.textColor = [BPKColor gray800];
+    }
+    
+    if(!self.enabled){
+        self.backgroundColor = [BPKColor white];
+        self.textColor = [BPKColor gray100];
+    }
+    
+    [self updateTitle];
+    [self setNeedsDisplay];
```
k0nserv: Is `setNeedsDisplay` required here? I would think that changing background + textColor would cause it to be set anyway?

Backpack/Chip/Classes/BPKChip.m
```diff
@@ -0,0 +1,161 @@
...
+}
+
+- (void)updateTitle {
+    if(self.title == nil){
+        return;
+    }
+    
+    self.titleLabel.text = self.title;
+    [self.titleLabel sizeToFit];
```
k0nserv: Can we drop this if we use the constraints setup described above?

Backpack/Chip/Classes/BPKChip.m
```diff
@@ -0,0 +1,161 @@
...
+    [self.titleLabel setTextColor:self.textColor];
+}
+
+- (void)handleSingleTap:(UITapGestureRecognizer *)recognizer
+{
```
k0nserv: This curly brace is in the wrong place

Backpack/Chip/Classes/BPKChip.m
```diff
@@ -0,0 +1,161 @@
...
+
+#pragma mark - Updates
+
+- (void)updateStyle {
+    if(self.selected) {
+        self.backgroundColor = [BPKColor blue500];
+        self.textColor = [BPKColor white];
+    } else {
+        self.backgroundColor = [BPKColor white];
+        self.textColor = [BPKColor gray800];
```
shaundon: Design for this says `gray700`.

Example/Backpack/View Controllers/ChipsViewController.swift
```diff
@@ -0,0 +1,32 @@
...
+
+class ChipsViewController: UIViewController {
+    @IBOutlet weak var regularChip: Chip!
+    @IBOutlet weak var selectedChip: Chip!
+    @IBOutlet weak var disabledChip: Chip!
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+//        selectedChip.title = "Test"
+//        disabledChip.setEnabled(true)
```
shaundon: We should remove these if they aren't needed.

Backpack/Chip/Classes/BPKChip.h
georgegillams: Yeah it comes from UIControl
If we don't declare this, I don't seem to be able to set `chip.enabled` or use `chip.setEnabled`. Am I doing something wrong?

Backpack/Chip/Classes/BPKChip.m
georgegillams: Yeah - don't need this

Backpack/Chip/Classes/BPKChip.h
k0nserv: Yeah that seems odd you should be able to still use those without it declared. I think the getter is `isEnabled` though however `chip.enabled = true` should work from Swift

Backpack/Chip/Classes/BPKChip.m
georgegillams: it works!

Backpack/Chip/Classes/BPKChip.m
georgegillams: Yes!

Backpack/Chip/Classes/BPKChip.h
georgegillams: When I remove this, it's not happy with `chip.enabled = true` but `chip.isEnabled = true` works

k0nserv: That sounds correct if it's fro Swift. The whole property has been renamed `isEnabled` in Swift

Backpack/Chip/Classes/BPKChip.m
georgegillams: See commit `discussion_r232276847`

Backpack/Chip/Classes/BPKChip.m
```diff
@@ -75,16 +74,27 @@ - (void)setup {
...
+#pragma mark - Layout overrides
+
+- (void)layoutSubviews {
+    self.layer.cornerRadius = self.bounds.size.height / 2;
+}
+
+- (CGSize)sizeThatFits:(CGSize)size {
+    [self.titleLabel sizeToFit];
+    float width = self.titleLabel.frame.size.width + 2 * BPKSpacingBase;
+    float height = self.titleLabel.frame.size.height + 2 * BPKSpacingMd;
```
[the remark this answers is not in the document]
shaundon: Not sure tbh - maybe try it and see how it looks?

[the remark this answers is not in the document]
k0nserv: Yeah I think that's good. They can't be constant because they aren't constant expressions. I've done something similar for other components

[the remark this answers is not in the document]
k0nserv: Here's an example https://github.com/Skyscanner/backpack-ios/blob/master/Backpack/Button/Classes/BPKButton.m#L518-L520

Backpack/Chip/Classes/BPKChip.m
```diff
@@ -0,0 +1,155 @@
...
+        [self setup];
+    }
+    
+    return self;
+}
+
+- (void)setup {
+    self.initializing = YES;
+    
+    self.titleLabel = [[BPKLabel alloc]initWithFrame:CGRectZero];
```
k0nserv: Missing space after `]`

Backpack/Chip/Classes/BPKChip.m
```diff
@@ -0,0 +1,155 @@
...
+    self.titleLabel.fontStyle = BPKFontStyleTextSm;
+    [self addSubview:self.titleLabel];
+    
+    self.titleLabel.textAlignment = NSTextAlignmentCenter;
+    
+    [self addTarget:self action:@selector(handleSingleTap:) forControlEvents:UIControlEventTouchDown];
```
k0nserv: This should be `UIControlEventTouchUpInside` I think

Backpack/Chip/Classes/BPKChip.m
```diff
@@ -0,0 +1,155 @@
...
+    self.initializing = NO;
+}
+
+#pragma mark - Layout overrides
+
+- (void)layoutSubviews {
+    self.layer.cornerRadius = self.bounds.size.height / 2;
+}
+
+- (CGSize)sizeThatFits:(CGSize)size {
```
k0nserv: Why do we need to override this in the first place?

Backpack/Chip/Classes/BPKChip.m
```diff
@@ -0,0 +1,155 @@
...
+    [super setEnabled:enabled];
+    [self updateStyle];
+}
+
+#pragma mark - Layout
+
+- (void)setupConstraints {
+    self.titleLabel.translatesAutoresizingMaskIntoConstraints = NO;
+    
+    [[self.titleLabel.leadingAnchor constraintEqualToAnchor:self.leadingAnchor constant:BPKSpacingBase] setActive:YES];
```
k0nserv: Can we use `NSLayoutConstraint activateConstraints:` instead?

Backpack/Chip/Classes/BPKChip.m
georgegillams: Because otherwise when used in a collection view (and probably other places) I can't get the correct size for setting cell size, as it the frame is `0x0`

Backpack/Chip/Classes/BPKChip.m
georgegillams: See commit `discussion_r232691305`

Backpack/Chip/Classes/BPKChip.m
georgegillams: See commit `discussion_r232690803`

Backpack/Chip/Classes/BPKChip.m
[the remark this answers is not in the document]
georgegillams: See commit `discussion_r232675275`

Backpack/Chip/Classes/BPKChip.m
```diff
@@ -0,0 +1,191 @@
...
+        [self setup];
+    }
+    
+    return self;
+}
+
+- (void)setup {
+    self.initializing = YES;
+    
+    self.tintLayer = [CALayer layer];
```
k0nserv: Is this layer approach based on discussions with Will?

georgegillams: So Will and I agreed that the chip should fade darker when pressed (to have the same visual langauge as the button)
The implementation was something I devised in response.
Open to alternative suggestions

Backpack/Chip/Classes/BPKChip.m
georgegillams: Yeah - but it doesn't seem to. Maybe I'm doing something wrong 🤷‍♂️

Backpack/Chip/Classes/BPKChip.m
```diff
@@ -0,0 +1,192 @@
...
+
+#import <Backpack/Font.h>
+#import <Backpack/Label.h>
+#import <Backpack/Color.h>
+#import <Backpack/Shadow.h>
+#import <Backpack/Spacing.h>
+
+NS_ASSUME_NONNULL_BEGIN
+@interface BPKChip()
+@property(nonatomic, getter=isInitializing) BOOL initializing;
```
k0nserv: It seems like this property is never used? Do we need it at all?

Backpack/Chip/Classes/BPKChip.m
```diff
@@ -0,0 +1,192 @@
...
+
+    [self updateStyle];
+    self.initializing = NO;
+}
+
+#pragma mark - Layout overrides
+
+- (void)layoutSubviews {
+    self.tintLayer.frame = self.bounds;
+    self.tintLayer.cornerRadius = self.bounds.size.height / 2;
```
k0nserv: Since the tintLayer is a subLayer of `self.layer` coyldn't we make `self.layer` `maskToBounds` and avoid this?

Backpack/Chip/Classes/BPKChip.m
```diff
@@ -0,0 +1,192 @@
...
+
+#pragma mark - Actions
+
+- (void)handleSingleTap:(UITapGestureRecognizer *)recognizer {
+    [self setSelected:!self.selected];
+}
+
+#pragma mark - Helpers
+
+- (void)animateTintLayer:(BOOL) tint {
```
k0nserv: Don't use a space after the parenthesis here

k0nserv: This method could do with a better name. I read `animateTintLayer:(BOOL)` as the bool controlling whether animation was used or not.

georgegillams: Actually - it doesn't need a name at all.
See commit `discussion_r232996079`

PR #113, opened 2018-11-13
First draft of the calendar API

Szabadember:
This pull requests adds a new calendar sub-pod together with the headers containing the signature of the proposed calendar component.


Backpack/Calendar/Classes/BPKCalendar.h
```diff
@@ -0,0 +1,79 @@
...
+};
+
+@protocol BPKCalendarDelegate
+
+/**
+ * Called when a date is selected.
+ *
+ * @param date The date selected.
+ */
+- (void)didSelectDate:(NSDate *)date;
```
tevelee: @Szabadember is the selection type is `BPKCalendarSelectionRange`, how can I access the currently selected range? or in case of `BPKCalendarSelectionMultiple `, the array of individual dates

Backpack/Calendar/Classes/BPKCalendar.h
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * Locale used for displaying name of days and months
+ */
+@property (copy, nonatomic) NSLocale *locale;
+
+/**
+ * First day of the week
+ */
+@property (assign, nonatomic) NSUInteger firstWeekday;
```
tevelee: you can access this information based on the provided `locale`:
```objc
NSCalendar *calendar = [[NSLocale currentLocale] objectForKey:NSLocaleCalendar];
NSUInteger firstDay = [calendar firstWeekday];
```

Backpack/Calendar/Classes/BPKCalendar.h
tevelee: should we have didSelectDateRange or similar delegate methods?

Backpack/Calendar/Classes/BPKCalendar.h
```diff
@@ -0,0 +1,79 @@
...
+/**
+ * Called when a date is deselected.
+ *
+ * @param date The date deselected.
+ */
+- (void)didDeselectDate:(NSDate *)date;
+
+@end
+
+@interface BPKCalendar : NSObject
```
tevelee: is should be a UIView subclass

Backpack/Calendar/Classes/BPKCalendar.h
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * First day of the week
+ */
+@property (assign, nonatomic) NSUInteger firstWeekday;
+
+/**
+ * Type of selection allowed
+ */
+@property (assign, nonatomic) BPKCalendarSelection selectionType;
```
tevelee: you don't need to specify `assign` and `strong`, because those are default values

Backpack/Calendar/Classes/BPKCalendar.h
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * The earliest date that the user is allowed to select
+ */
+@property NSDate *minDate;
```
tevelee: should these be `nonatomic` properties?

tevelee:
The designs contain weekday titles on the top of the view, which is part of the calendar view itself. Shall we inject those strings (as we do not want to depend on the localisation solution)?

tevelee:
Should we allow users to select past dates? (if we include a cal in backpack, that should support generic use-cases as well). In this case, there should be a minimum selectable date or an array of dates

Backpack/Calendar/Classes/BPKCalendarAppearance.h
```diff
@@ -0,0 +1,108 @@
...
+ */
+
+#import <Foundation/Foundation.h>
+
+@interface BPKCalendarAppearance : NSObject
+
+/**
+ * The font of the weekday text.
+ */
+@property (strong, nonatomic) UIFont   *weekdayFont;
```
tevelee: do you intentionally have these many whitespaces?

tevelee: as commented on another file, you don't need to specify `strong` for the property as it's the default one

Backpack/Calendar/Classes/BPKCalendar.h
Szabadember: 👍

Backpack/Calendar/Classes/BPKCalendarAppearance.h
Szabadember: 👍

Backpack/Calendar/Classes/BPKCalendar.h
Szabadember: 👍

Backpack/Calendar/Classes/BPKCalendar.h
Szabadember: 👍

Backpack/Calendar/Classes/BPKCalendar.h
```diff
@@ -0,0 +1,75 @@
...
+
+@class BPKCalendar;
+
+/**
+ * Enum values for specifying calendar selection type
+ */
+typedef NS_ENUM(NSUInteger, BPKCalendarSelection) {
+    BPKCalendarSelectionSingle = 0,
+    BPKCalendarSelectionRange = 1,
+    BPKCalendarSelectionMultiple = 2,
```
k0nserv: What does multiple mean? Multiple single selections or multiple ranges? Do we need to clarify that?

Backpack/Calendar/Classes/BPKCalendar.h
```diff
@@ -0,0 +1,75 @@
...
+
+@protocol BPKCalendarDelegate
+
+/**
+ * Called when a date selection changed.
+ *
+ * @param calendar The backpack calendar.
+ * @param dateList List of selected dates.
+ */
+- (void)calendar:(BPKCalendar *)calendar didChangeDateSelection:(NSArray<NSDate *> *)dateList;
```
k0nserv: How does this differentiated between range and single selection? Should we have a speific type for a range selection?

Backpack/Calendar/Classes/BPKCalendarAppearance.h
```diff
@@ -0,0 +1,108 @@
...
+ */
+
+#import <Foundation/Foundation.h>
+
+@interface BPKCalendarAppearance : NSObject
+
+/**
+ * The font of the weekday text.
+ */
+@property (nonatomic) UIFont *weekdayFont;
```
k0nserv: This will not be enough to configure the text display was we have switched all text rendering to `NSAttributedString`

Backpack/Calendar/Classes/BPKCalendarAppearance.h
```diff
@@ -0,0 +1,108 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#import <Foundation/Foundation.h>
+
+@interface BPKCalendarAppearance : NSObject
```
k0nserv: Do we really need to be able to customize the appearance this much? Couldn't we make the first version in a single style and then add something like this if required?

Backpack/Calendar/Classes/BPKCalendar.h
Szabadember: Multiple single selection

Backpack/Calendar/Classes/BPKCalendar.h
Szabadember: In single selection mode the array will always have 0/1 element, in range selection mode two elements, in multi selection mode 0+ elements

Do we need to have a specific type? My first plan was to use NSDateInterval but this is only available from iOS10.0 onwards

Backpack/Calendar/Classes/BPKCalendarAppearance.h
Szabadember: I am more than happy to cut scope. We could remove the whole BPKCalendarAppearance.h altogether.

Backpack/Calendar/Classes/BPKCalendarAppearance.h
Szabadember: See next answer

PR #121, opened 2018-11-20
[BPK-2077] Add docs examples to match Android

shaundon:
Added two new stories:

simulator screen shot - iphone 8 - 2018-11-20 at 14 05 04

I reused the Segue IDs so I didn't have to recreate the code, it's just two new table view cells.


georgegillams:
Why don't we remove `Normal` and `Delete` from the menu, as they are effectively there twice now?

PR #127, opened 2018-11-26
[No JIRA] Add npm ci to release task

shaundon:
https://github.com/Skyscanner/backpack-android/pull/129/files

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -51,6 +51,11 @@ def install_pods_in_example_project()
   $?.exitstatus == 0
 end
 
+def check_pristine()
+  changes = `git status --porcelain | grep -Ev "CHANGELOG|UNRELEASED"`.lines
+  changes.length == 0
+end
+
 task :analyze do
   sh "set -o pipefail && ! xcodebuild -workspace #{EXAMPLE_WORKSPACE} -scheme \"#{EXAMPLE_SCHEMA}\" -sdk #{BUILD_SDK} -destination \"platform=iOS Simulator,name=iPhone 8\" ONLY_ACTIVE_ARCH=NO analyze 2>&1 | xcpretty | grep -A 5 \"#{ANALYZE_FAIL_MESSAGE}\""
 end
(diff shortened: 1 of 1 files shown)


k0nserv:
Isn’t the idea with this that we can’t acciden release with outdated tokens? If so surely we also need to run the gulp build too?

shaundon:
Ah yeah good shout

georgegillams:
To achieve that, we could use a helper like this. Although this will return false if `CHANGELOG.md` has been update, so may need to be tweaked a little.
```
def check_pristine()
  changes = `git status --porcelain`.lines
  result = changes.length < 0
end
```

georgegillams:
This is what a use will see if `gulp` has made changes:
screenshot 2018-11-27 at 09 30 47

georgegillams:
The pristine check allows for changes in `CHANGELOG.md` and `UNRELESED.md` as these are expected when doing a release

Rakefile
```diff
@@ -68,11 +73,13 @@ task ci: [:lint, :analyze, :test]
 
 # task release: :test do
 task release: :ci do
-  sh "npm ci"
   abort red 'Must be on master branch' unless current_branch == 'master'
   abort red 'Must have push access to Backpack on CocoaPods trunk' unless has_trunk_push
   abort red 'Git branch is not up to date please pull' unless branch_up_to_date
 
+  sh "npm ci"
```
georgegillams: I moved this as we might as well get the fast checks above done and passed before we make the person releasing wait for `npm ci`

Rakefile
```diff
@@ -51,6 +51,11 @@ def install_pods_in_example_project()
   $?.exitstatus == 0
 end
 
+def check_pristine()
```
georgegillams: The pristine check allows for changes in `CHANGELOG.md` and `UNRELESED.md` as these are expected when doing a release

k0nserv: Podspec is done after this point right? So it doesn't have to be checked

Rakefile
```diff
@@ -51,6 +51,11 @@ def install_pods_in_example_project()
...
+  changes = `git status --porcelain | grep -Ev "CHANGELOG|UNRELEASED"`.lines
+  result = changes.length < 1
```
k0nserv: This can just be `changes.length == 0`

georgegillams: Good point

Rakefile
georgegillams: I don't know what your point is?

georgegillams: Do you mean this check should also exclude `Podspec` as it will be checked later anyway?

k0nserv: No I think this is fine since `Podspec` is modified later in the flow

PR #128, opened 2018-11-26
[BPK-2084] Change XXL and XXXL tokens

shaundon:


UNRELEASED.md
```diff
@@ -1,2 +1,6 @@
 # Unreleased
 
+**Breaking:**
+
+- Backpack/Font
+  - XXL and XXL font sizes are now 42pt and 36pt (previously 36pt and 30pt).
```
k0nserv: This is incorrect.  
```suggestion
  - XXXL and XXL font sizes are now 36pt and 30pt (previously 42pt and 36pt). The 42pt font size have been removed for the time being
```

package.json
```diff
@@ -36,7 +36,7 @@
   "homepage": "https://github.com/Skyscanner/backpack-ios#readme",
   "devDependencies": {
     "bpk-svgs": "^6.0.2",
-    "bpk-tokens": "^27.2.1",
+    "bpk-tokens": "^27.3.1-beta.11",
```
k0nserv: Are we planning to release like this or will we wait for a stable release of `bpk-tokens`?

UNRELEASED.md
shaundon: Damn I just got schooled

shaundon: Fixed

package.json
shaundon: I think we should wait for a stable release. We have pretty much everything we need:

- [x] Tokens source of truth
- [x] iOS (this PR)
- [x] RN
- [x] Android (https://github.com/Skyscanner/backpack-android/pull/128/files)

So once iOS and Android are merged we can do quick follow-up PRs to bump to stable tokens and release all repos at once.

k0nserv: Sounds good then

UNRELEASED.md
```diff
@@ -1,2 +1,6 @@
 # Unreleased
 
+**Breaking:**
+
+- Backpack/Font
+  - XXXL and XXL font sizes are now 36pt and 30pt (previously 42pt and 36pt). The 42pt font size has been removed for the time being.
```
georgegillams: Would this make more sense the other way round? ie
```
XXL and XXXL font sizes are now 30pt and 36pt (previously 36pt and 42pt). The ???? font size has been removed for the time being.
```

Also I think you've got copy-pasta on the 42pt size

PR #148, opened 2018-12-11
Snapshot tests for BPKCalendar

Szabadember:


Example/SnapshotTests/BPKCalendarSnapshotTest.m
```diff
@@ -0,0 +1,103 @@
...
+    
+    [self configureParentView:parentView forCalendar:bpkCalendar];
+    bpkCalendar.selectionType = BPKCalendarSelectionSingle;
+    bpkCalendar.selectedDates = @[ self.date1 ];
+    [bpkCalendar reloadData];
+    
+    FBSnapshotVerifyView(parentView, nil);
+}
+
+- (void)testCalendarWithRangeSelection {
```
tevelee: can you add another case for same-day range selection as well?

Example/SnapshotTests/BPKCalendarSnapshotTest.m
```diff
@@ -0,0 +1,103 @@
...
+    UIView *parentView = [[UIView alloc] initWithFrame:CGRectZero];
+    BPKCalendar *bpkCalendar = [[BPKCalendar alloc] initWithFrame:CGRectZero];
```
tevelee: Parent view and calendar init logic are there in all of the test cases. The same 2 lines of code. Maybe it's worth extracting them

shaundon:
It all looks good, but I'm a bit confused by the year pill looking weird in the snapshots:

screen shot 2018-12-13 at 13 52 55

Do you know why this is, and is it a problem?

shaundon:
Alright cool, sounds like you've mitigated it well :)

PR #150, opened 2018-12-18
[No JIRA] Bump deployment target to iOS 10 🙌

k0nserv:


UNRELEASED.md
```diff
@@ -1,2 +1,4 @@
 # Unreleased
 
+**Breaking:**
+  + Bumbed deployment target to iOS 10.0.
```
shaundon:
```suggestion
  + Bumped deployment target to iOS 10.0.
```

shaundon:
Looks good once CI is happy!

UNRELEASED.md
k0nserv: I've been making this mistake since I learned to write. It's a classic Hugo

PR #153, opened 2018-12-18
Improvements to the Calendar component

tevelee:
based on 
https://gojira.skyscanner.net/browse/STARK-1771


shaundon:
So sorry for the long delay with this, it wasn't our intention to leave this unreviewed for such a long time.

Couple of things:

1) I had to remove the `testCalendarWithoutSelection` snapshot as it would cause a build fail if it was run on a different day than when the test was last generated, because it always highlights the current day. So for example if the snapshot has '1' highlighted and we run it on the 3rd, the snapshot changes and the build fails. I wasn't sure how to fix it so I removed the test as a quick fix in https://github.com/Skyscanner/backpack-ios/pull/155. As this PR still contains that snapshot it's conflicting so it'll need to be removed (or you could change the test can be fixed to mock `calendar.today` 😇 )

2. The behaviour of the year pill feels a bit off to me. It doesn't appear until I stop scrolling, then stays on screen forever, then disappears again as soon as I stop scrolling. Could you check that this is working correctly? I've attached a video below.

[Kapture 2019-01-11 at 14.45.32.mp4.zip](https://github.com/Skyscanner/backpack-ios/files/2750047/Kapture.2019-01-11.at.14.45.32.mp4.zip)

shaundon:
@tevelee just making sure you've seen this ⬆️ 😇

tevelee:
I'll close this and let @Szabadember to continue calendar improvements

PR #161, opened 2019-01-22
[No JIRA] Fix build issue for framework changes

k0nserv:
Previously the new Xcode build system caused development pods not to
correctly rebuild when changes were made to the source. This required a
clean build for each change.
See: https://github.com/CocoaPods/CocoaPods/issues/8073

With this chagne we adopt a new CocoaPods option
`disable_input_output_paths: true` which resolves this issue while still
staying on the new build system.

See: https://github.com/CocoaPods/CocoaPods/pull/8105


georgegillams:
YOU LEGEND!

Thank you

PR #173, opened 2019-02-07
[NO-JIRA] Fix button disabled style

georgegillams:
Setting title after disabling the button causes the attributed to be set with the wrong text color.
I have modified `currentContentColor` to return the disabled content color if disabled.

I have also changed the background color for disabled link buttons as it should be invisible, but was white.

### DIFF
--- a/Backpack/Button/Classes/BPKButton.m
+++ b/Backpack/Button/Classes/BPKButton.m
@@ -47,52 +47,52 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (instancetype)initWithSize:(BPKButtonSize)size style:(BPKButtonStyle)style {
     self = [super initWithFrame:CGRectZero];
-
+    
     if (self) {
         [self setupWithSize:size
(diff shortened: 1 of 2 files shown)


Backpack/Button/Classes/BPKButton.m
```diff
@@ -509,7 +512,7 @@ - (void)setLinkStyleWithColor:(UIColor *)color {
 }
 
 - (void)setDisabledStyle {
-    UIColor *backgroundColor = (self.style == BPKButtonStyleLink || self.style == BPKButtonStyleOutline) ? BPKColor.white : BPKColor.gray100;
+    UIColor *backgroundColor = self.style == BPKButtonStyleLink ? [UIColor colorWithWhite:1 alpha:0] : (self.style == BPKButtonStyleOutline ? BPKColor.white : BPKColor.gray100);
```
k0nserv: Use `UIColor.clearColor` instead

PR #175, opened 2019-02-11
Add a style guide for Objective-c

k0nserv:


georgegillams:
Is this only enforced by hand? Seems like the sort of thing that could easily be passed over??

k0nserv:
There are no good tools to enforce this stuff automatically I believe. We can have a deeper look, but I think the state is just a bit 🤷‍♂️

PR #176, opened 2019-02-12
[BPK-1942] Update panel text to match other platforms

shaundon:
Tweaked the constraints to make them look ok with this increased text, too.


shaundon:
Good catch, now done. I'm getting the hang of applying constraints now!

PR #178, opened 2019-02-13
[BPK-2245] Add settings view to Example app

georgegillams:
## TODO in subsequent PR's
 - [ ] Implement the settings screen controls.

### DIFF
--- a/Example/Backpack/View Controllers/BPKRootListTableViewController.m
+++ b/Example/Backpack/View Controllers/BPKRootListTableViewController.m
@@ -17,12 +17,21 @@
  */
 
 #import "BPKRootListTableViewController.h"
+#import <Backpack/Icon.h>
+#import <Backpack/Color.h>
 
 @interface BPKRootListTableViewController ()
-
+@property (weak, nonatomic) IBOutlet UIBarButtonItem *settingsButton;
 @end
 
 @implementation BPKRootListTableViewController
+
+- (void) viewDidLoad {
+    [super viewDidLoad];
+
+    UIImage *largeSettingsIcon = [BPKIcon templateIconNamed:@"settings" size:BPKIconSizeLarge];
+    self.settingsButton.image = largeSettingsIcon;
+}
 
 - (void)prepareForSegue:(UIStoryboardSegue *)segue sender:(id)sender {
     // We cheat and use the same view controller for both fonts and labels
--- a/Example/Backpack/View Controllers/SettingsViewController.swift
+++ b/Example/Backpack/View Controllers/SettingsViewController.swift
@@ -1,7 +1,7 @@
 /*
  * Backpack - Skyscanner's Design System
  *
- * Copyright 2018 Skyscanner Ltd
+ * Copyright 2019 Skyscanner Ltd
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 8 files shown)


Backpack/Button/Classes/BPKButton.m
```diff
@@ -374,7 +374,8 @@ - (void)updateContentColor {
         [self setAttributedTitle:nil forState:UIControlStateNormal];
     }
     
-    
+
+    self.tintColor = self.currentContentColor;
```
georgegillams: This seems to be required to make the tint colour be correct when first rendered 🤔

shaundon:
Something about the UI here feels a bit off. I think settings screens generally appear as separate flows, like in a modal, rather than as a nested view off of the current hierarchy. See the attached movie for an example. 

[IMG_2339.TRIM.MOV.zip](https://github.com/Skyscanner/backpack-ios/files/2860479/IMG_2339.TRIM.MOV.zip)

How difficult would it be to do it this way? If it's a big job I'm happy to leave it be as I know it's a small point.

georgegillams:
@shaundon I made it like that.
https://github.com/Skyscanner/backpack-ios/pull/178#issuecomment-463196321

Have also implemented shake behaviour too.

shaundon:
This needs a bit of work:

screen shot 2019-02-14 at 10 52 43

1. It's using the wrong type of table view (should be the grouped style)
2. The theming button is non functional. I think this was maybe added by mistake as the original PR didn't include it.
3. We definitely shouldn't have the title and close button as part of the table.

I think it should look something like this:

screen shot 2019-02-14 at 11 04 39

Example/Backpack/View Controllers/BPKRootListTableViewController.m
```diff
@@ -17,13 +17,27 @@
  */
 
 #import "BPKRootListTableViewController.h"
```
k0nserv: Can we have an empty line after this import to group the imports?

Example/Backpack/View Controllers/BPKRootListTableViewController.m
```diff
@@ -17,13 +17,27 @@
...
+#import <Backpack/Button.h>
+#import <Backpack/Icon.h>
 
 @interface BPKRootListTableViewController ()
+@property (weak, nonatomic) IBOutlet BPKButton *settingsButton;
```
k0nserv: I am not sure how I feel about this being a custom button when we could just use as regular `UIBarButtonItem` and `- (instancetype)initWithImage:(UIImage *)image style:(UIBarButtonItemStyle)style target:(id)target action:(SEL)action;`

Example/Backpack/View Controllers/SettingsViewController.swift
```diff
@@ -0,0 +1,35 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2018 Skyscanner Ltd
```
k0nserv: This year is wrong

Example/Backpack/View Controllers/SettingsViewController.swift
```diff
@@ -0,0 +1,35 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import UIKit
+import Backpack
+
+class SettingsViewController: UITableViewController {
+    @IBOutlet weak var closeButton: Button!
```
k0nserv: Similar here do we really need to use a `Backpack.Button` over a regular system button with the appropraite `tintColor`?

georgegillams: Dogfooding man!

k0nserv: It doesn't feel appropriate to use a Backpack Button in the navigation bar in that case we should provide an API for configuring bar button items

georgegillams: But this is an example app, and it works...

Example/Backpack/View Controllers/BPKRootListTableViewController.m
georgegillams: I've now swapped the settings and close buttons out for system ones. Let me know what you think of it now 😃

Example/Backpack/View Controllers/BPKRootListTableViewController.m
```diff
@@ -35,8 +35,9 @@ - (void)viewDidLoad {
     UIImage *largeSettingsIcon = [BPKIcon templateIconNamed:@"settings" size:BPKIconSizeLarge];
     [self.settingsButton setImage:largeSettingsIcon];
```
k0nserv: This should be removed no?

georgegillams: Do you mean for it to look like this?
https://github.com/Skyscanner/backpack-ios/pull/178/commits/9238f3170600cb0a7eef9e962725fb4c183a561d

PR #183, opened 2019-02-21
Added main thread assertions

tevelee:
Ensuring at runtime that all Backpack UI components are being called from the main thread.

_If you are curious about how we review, please read through the [code review guidelines](./CODE_REVIEW_GUIDELINES.md)_

### DIFF
--- a/Backpack/Common.h
+++ b/Backpack/Common.h
@@ -20,13 +20,6 @@
 
 #import <pthread.h>
 
-#if !defined(NS_BLOCK_ASSERTIONS)
-#define BPK_ASSERTIONS_ENABLED 1
-#else
-#define BPK_ASSERTIONS_ENABLED 0
-#endif
-
-#define BPKAssert(condition, desc, ...) NSAssert(condition, desc, ##__VA_ARGS__)
-#define BPKAssertMainThread() BPKAssert(!BPK_ASSERTIONS_ENABLED || 0 != pthread_main_np(), @"This method must be called on the main thread")
+#define BPKAssertMainThread() NSAssert(0 != pthread_main_np(), @"This method must be called on the main thread")
 
 #endif
--- a/Backpack/Common.h
+++ b/Backpack/Common.h
@@ -15,39 +15,18 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+#ifndef BPK_Common_h
+#define BPK_Common_h
 
-#import "BPKLondonThemeContainer.h"
-#import <Backpack/Common.h>
+#import <pthread.h>
 
-NS_ASSUME_NONNULL_BEGIN
+#if !defined(NS_BLOCK_ASSERTIONS)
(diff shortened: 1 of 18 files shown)


Backpack/Common.h
```diff
@@ -0,0 +1,32 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef BPK_Common_h
+#define BPK_Common_h
+
+#import <pthread.h>
+
+#if !defined(NS_BLOCK_ASSERTIONS)
+#define BPK_ASSERTIONS_ENABLED 1
```
k0nserv: Do we need our own method for turning assertions on and off?

Backpack/Common.h
```diff
@@ -0,0 +1,32 @@
...
+#else
+#define BPK_ASSERTIONS_ENABLED 0
+#endif
+
+#define BPKAssert(condition, desc, ...) NSAssert(condition, desc, ##__VA_ARGS__)
```
k0nserv: This should also be guarded with a `BPK_ASSERTIONS_ENABLED` check if we do think we should have a custom assertion control?

Backpack/Panel/Classes/BPKPanel.m
```diff
@@ -64,6 +68,7 @@ - (void)setPadded:(BOOL)padded {
 }
 
 - (void)addSubview:(UIView *)view {
+    BPKAssertMainThread();
     NSAssert(self.subviews.count == 0, @"BPKPanel can only have a single subview");
```
k0nserv: Should we switch out `NSAssert`s for `BPKAssert`s?

tevelee:
Thanks @shaundon and @k0nserv for the reviews! I'll update the branch shortly, implementing Hugo's suggestions. 

Regarding Shaun's question, I don't think there's a better way than guarding public methods, manually. 
I took Texture (the old AsyncDisplayKit) as an example, they do it the same way: https://github.com/TextureGroup/Texture/blob/d4efe95ec5399def621096eca9016acdabefaf0b/Source/ASCellNode.mm#L208

shaundon:
Fair enough - thanks for clarifying!

Backpack/Common.h
```diff
@@ -0,0 +1,25 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef BPK_Common_h
+#define BPK_Common_h
+
+#import <pthread.h>
+
+#define BPKAssertMainThread() NSAssert(defined(NS_BLOCK_ASSERTIONS) || 0 != pthread_main_np(), @"This method must be called on the main thread")
```
k0nserv: Do we even need to check `NS_BLOCK_ASSERTIONS` ourselves? My impression is that `NSAssert` already does that

tevelee: Good point

PR #187, opened 2019-02-26
[BPK-2245] Add switch themeable attributes to readme

georgegillams:


Backpack/Switch/README.md
```diff
@@ -22,3 +22,7 @@ import Backpack
 let `switch` = Backpack.Switch();
 // Position switch with autolayout or other method
 ```
+
+### Appearance attributes
+`(UIColor)onTintColor`
```
georgegillams: We need to document the type here too, as a appearance attribute can be any object type - so it won't necessarily be obvious

PR #195, opened 2019-03-07
[No JIRA] Fix button layout bug

k0nserv:
When a button was laid out after having and image set and then the image
being set back to `nil` it would not lay out correctly. This was because
we were relying on `self.imageView.image` when determining if the button
had an image. When the image is set to `nil` via `setImage:forState` the
underlying `image` property on `UIImageView` still contains the image
and is thus not `nil`. With this change we use `currentImage` which
correctly determines if and image is set.


k0nserv:
I verified that even in highlighted state `currentImage` is non-nil since it has been set for the normal control state

PR #197, opened 2019-03-11
Exposing some of the underlying scrollview's properties as readonly props

Szabadember:
This will enable more complex interactions with the calendar. example: animated external header

### DIFF
--- a/Backpack/Calendar/Classes/BPKCalendar.h
+++ b/Backpack/Calendar/Classes/BPKCalendar.h
@@ -42,9 +42,9 @@ NS_SWIFT_NAME(CalendarDelegate) @protocol BPKCalendarDelegate <NSObject>
 /**
  * Called when the calendar was scrolled
  * @param calendar The backpack calendar.
- * @param contentOffset The content offset
+ * @param scrollView The scrollView of the calendar
  */
-- (void)calendar:(BPKCalendar *)calendar didScroll:(CGPoint)contentOffset;
+- (void)calendar:(BPKCalendar *)calendar didScroll:(UIScrollView *)scrollView;
 
 @end
 
--- a/Backpack/Calendar/Classes/BPKCalendar.h
+++ b/Backpack/Calendar/Classes/BPKCalendar.h
@@ -42,9 +42,9 @@ NS_SWIFT_NAME(CalendarDelegate) @protocol BPKCalendarDelegate <NSObject>
 /**
  * Called when the calendar was scrolled
  * @param calendar The backpack calendar.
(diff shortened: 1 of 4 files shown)


Backpack/Calendar/Classes/BPKCalendar.h
```diff
@@ -83,6 +83,36 @@ NS_SWIFT_NAME(Calendar) @interface BPKCalendar: UIView
...
+
+/**
+ * The underlying scrollView's isDecelerating
+ */
+@property (readonly) BOOL isDecelerating;
+
+/**
+ * The underlying scrollView's isTracking
+ */
+@property (readonly) BOOL isTracking;
```
k0nserv: Shouldn't this and the other `BOOL ` properties be `@property(readonly, getter=isTracking) BOOL tracking` to align with `UIScrollView`?

UNRELEASED.md
```diff
@@ -1,2 +1,4 @@
 # Unreleased
 
+- BPKCalendar:
```
k0nserv: Can we include the change category(Added) here and use `Backpack/Calendar` instead of `BPKCalendar`.

```suggestion
**Added:**
- Backapck/Calendar
  - Exposed some properties of the calendar's underlying scrollView.
```

For completeness wouldn't it be nice to list all properties?

Backpack/Calendar/Classes/BPKCalendar.h
Szabadember: What is the difference? Creating a readonly property with an overriden getter like I did will achieve the same.

k0nserv: It would make it easier to make it `readwrite` later is my thinking

Szabadember: No ivar is created by the compiler for that property because it is readonly and the getter is overridden.

Szabadember: I am not sure if we want to make it readwrite

Szabadember: These properties are readonly on UISCrollView: https://developer.apple.com/documentation/uikit/uiscrollview/1619413-istracking

Szabadember: contentInset and contentSize is not intented to be modified, I think the only contentOffset could be made readwrite to scroll programmatically

k0nserv: Fair enough let's merge

PR #200, opened 2019-03-14
[NO-JIRA] Add Doha theme

georgegillams:

### DIFF
--- a/Backpack/Theme/Classes/BPKDohaTheme.m
+++ b/Backpack/Theme/Classes/BPKDohaTheme.m
@@ -1,0 +1,36 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2018-2019 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


Backpack/Theme/Classes/BPKDohaTheme.m
```diff
@@ -0,0 +1,39 @@
...
+ * limitations under the License.
+ */
+
+#import "BPKDohaTheme.h"
+#import "BPKDohaThemeContainer.h"
+
+#import <Backpack/Button.h>
+#import <Backpack/Chip.h>
+#import <Backpack/Spinner.h>
+#import <Backpack/Switch.h>
```
shaundon: It looks like Button, Chip and Spinner are unused here?

georgegillams:
> How about making a generic / abstract theme that we can subclass, so then we could do things like we do in JS where we just have to pass in something like primaryColor, secondaryColor etc and it returns a correct theme?

Hmm... Yeah sounds good. Can we get this merged first? The API will be the same so we can refactor after this PR

PR #203, opened 2019-03-18
Introducing SimpleDate in BPKCalendar

zsoltvarnai:
The update changes the public interface of the BPKCalendar to surface the selected dates as SimpleDate, instead of NSDate to make it explicit that the values should be timezone independent and less error prone to accidental timezone conversions.

### DIFF
--- a/Backpack/Calendar/Classes/BPKCalendar.h
+++ b/Backpack/Calendar/Classes/BPKCalendar.h
@@ -17,6 +17,7 @@
  */
 
 @class BPKCalendar;
+@class BPKSimpleDate;
 
 /**
  * Enum values for specifying calendar selection type
@@ -35,7 +36,7 @@ NS_SWIFT_NAME(CalendarDelegate) @protocol BPKCalendarDelegate <NSObject>
  * @param calendar The backpack calendar.
  * @param dateList List of selected dates.
  */
-- (void)calendar:(BPKCalendar *)calendar didChangeDateSelection:(NSArray<NSDate *> *)dateList;
+- (void)calendar:(BPKCalendar *)calendar didChangeDateSelection:(NSArray<BPKSimpleDate *> *)dateList;
 
 @optional
 
@@ -71,17 +72,17 @@ NS_SWIFT_NAME(Calendar) @interface BPKCalendar: UIView
 /**
  * List of selected dates
  */
-@property (nonatomic) NSArray<NSDate *> *selectedDates;
+@property (nonatomic) NSArray<BPKSimpleDate *> *selectedDates;
 
 /**
  * The earliest date that the user is allowed to select
  */
-@property (nonatomic) NSDate *minDate;
+@property (nonatomic) BPKSimpleDate *minDate;
 
 /**
  * The latest date that the user is allowed to select
  */
-@property (nonatomic) NSDate *maxDate;
+@property (nonatomic) BPKSimpleDate *maxDate;
 
 /**
  * The underlying scrollView's content offset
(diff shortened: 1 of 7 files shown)


Backpack/Calendar/Classes/BPKCalendar.h
```diff
@@ -16,6 +16,8 @@
  * limitations under the License.
  */
 
+#import "BPKSimpleDate.h"
```
k0nserv: `BPKSimpleDate` can be forward declared in this file I believe

Backpack/Calendar/Classes/BPKCalendar.m
```diff
@@ -445,6 +445,40 @@ + (BOOL)date:(NSDate*)date isBetweenDate:(NSDate*)beginDate andDate:(NSDate*)end
     return YES;
 }
 
+- (BPKSimpleDate *)simpleDateFromDate:(NSDate *)date {
+    if(date == nil) {
```
k0nserv: Can we explicitly mark the return type and parameter as `nullable` or `nonnull` and either remove this check or keep it.

Backpack/Calendar/Classes/BPKCalendar.m
```diff
@@ -445,6 +445,40 @@ + (BOOL)date:(NSDate*)date isBetweenDate:(NSDate*)beginDate andDate:(NSDate*)end
...
+        return nil;
+    }
+    
+    NSDateComponents *components = [self.calendarView.gregorian components:NSCalendarUnitYear|NSCalendarUnitMonth|NSCalendarUnitDay
```
k0nserv: Can we add a few spaces between components and `|` here?

Backpack/Calendar/Classes/BPKCalendar.m
```diff
@@ -445,6 +445,40 @@ + (BOOL)date:(NSDate*)date isBetweenDate:(NSDate*)beginDate andDate:(NSDate*)end
...
+                                                                  fromDate:date];
+    
+    return [[BPKSimpleDate alloc] initWithYear:components.year month:components.month day:components.day];
+}
+
+- (NSArray<BPKSimpleDate *> *)simpleDatesFromDates:(NSArray<NSDate *> *)dates {
+    NSMutableArray *simpleDates = [NSMutableArray new];
```
k0nserv: We know the new capacity and can use `initWithCapacity:` here

Backpack/Calendar/Classes/BPKCalendar.m
```diff
@@ -445,6 +445,40 @@ + (BOOL)date:(NSDate*)date isBetweenDate:(NSDate*)beginDate andDate:(NSDate*)end
...
+    
+    for (NSDate *date in dates) {
+        [simpleDates addObject:[self simpleDateFromDate:date]];
+    }
+    
+    return [simpleDates copy];
+}
+
+-(NSDate *)dateFromSimpleDate:(BPKSimpleDate *)simpleDate {
+    if(simpleDate == nil) {
```
k0nserv: Can we make the same consideration about `nonnull` vs `nullable` here?

Backpack/Calendar/Classes/BPKCalendar.m
```diff
@@ -445,6 +445,40 @@ + (BOOL)date:(NSDate*)date isBetweenDate:(NSDate*)beginDate andDate:(NSDate*)end
...
+        return nil;
+    }
+    
+    NSDateComponents *components = [[NSDateComponents alloc] init];
```
k0nserv: `init` isn't documented are we sure that it correctly sets the timeZone property?

Backpack/Calendar/Classes/BPKSimpleDate.m
```diff
@@ -67,6 +66,10 @@ - (NSUInteger)hash {
     return self.year ^ self.month ^ self.day;
 }
 
+-(NSString *)description {
+    return [NSString stringWithFormat:@"%lu-%lu-%lu", (unsigned long)self.year, (unsigned long)self.month, (unsigned long)self.day];
```
k0nserv: Can we make this `<BPKSimpleDate %lu-%lu-%lu>`?

k0nserv:
The snapshot tests haven't been updated, that's why CI is failing. See `BPKCalendarSnapshotTests.m`

Backpack/Calendar/Classes/BPKCalendar.h
zsoltvarnai: I was going with that first, but then whoever uses the calendar needs to import one more thing on their own, is that preferred?

k0nserv: People who use the calendar should be importing `Backpack/Calendar.h` not `Backpack/BPKCalendar.h`. We should add `BPKSimpleDate.h` to `Calendar.h`

PR #204, opened 2019-03-18
[NO-JIRA] Refactor theme classes

georgegillams:

### DIFF
--- a/Backpack/Theme/Classes/BPKTheme.m
+++ b/Backpack/Theme/Classes/BPKTheme.m
@@ -16,22 +16,34 @@
  * limitations under the License.
  */
 
-#import "BPKHongKongTheme.h"
-#import "BPKHongKongThemeContainer.h"
+#import "BPKTheme.h"
+#import "BPKDohaThemeContainer.h"
 
-#import <Backpack/Button.h>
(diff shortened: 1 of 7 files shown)


Backpack/Theme/Classes/BPKTheme.m
```diff
@@ -0,0 +1,49 @@
...
+    [[BPKSwitch appearanceWhenContainedInInstancesOfClasses:@[class]] setOnTintColor:switchPrimaryColor];
+}
+
++ (UIColor *)getSwitchPrimaryColor {
+    // "BPKTheme should not be applied directly. Subclasses that override `getSwitchPrimaryColor` should be used instead.
+    assert(false);
+    return nil;
+}
+
++ (Class)getThemeContainerClass {
```
georgegillams: The way I've done this is to add overridable methods for retrieveing theme-specific values to the base theme class.
If these methods are invoked directly (ie they are not overridden) then they will cause assertion failures. This ensures that the theme being applied is completely defined.

georgegillams: (As it's only an assertion failure, if it gets into production it won't cause the app to crash.)

PR #206, opened 2019-03-19
[NO-JIRA] Repeat failing tests on CI

georgegillams:

Results from a similar approach I took using JS

| Attempt | Result |
| ----- | ----- |
| 1 | Passed (1st attempt) |
| 2 | Passed (1st attempt) |
| 3 | Passed (1st attempt) |
| 4 | Passed (1st attempt) |
| 5 | Passed (1st attempt) |
| 6 | Passed (1st attempt) |
| 7 | Passed (1st attempt) |
| 8 | Passed (**2nd** attempt) |
| 9 | Passed (**2nd** attempt) |
| 10 | Passed (1st attempt) |
| 11 | Passed (1st attempt) |

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -8,6 +8,7 @@ EXAMPLE_SCHEMA = 'Backpack Native'
 UNIT_TEST_SCHEMA = 'Backpack Native Unit Tests'
 VERSION_FORMAT = '%M.%m.%p%s%d'
 PODSPEC = 'Backpack.podspec'
+MAX_TEST_REPEATS = 3
 ANALYZE_FAIL_MESSAGE = '⚠️'
 
 def ask(question)
@@ -46,6 +47,21 @@ def file_is_dirty(filename)
   $?.exitstatus != 0
 end
 
+def repeat_on_fail(command, run_count = 1)
+  if run_count > MAX_TEST_REPEATS
+    puts "Process is still failing after #{MAX_TEST_REPEATS} attempts.\n Giving up."
+    abort "Tests failed."
+  end
+
+  puts "Running process. This is attempt #{run_count} of #{MAX_TEST_REPEATS}."
+  `#{command}`
+  if $?.exitstatus == 0
+    puts "Process succeeded on attempt #{run_count}."
+  else
+    repeat_on_fail(command, run_count + 1)
+  end
+end
+
 def install_pods_in_example_project()
   `(cd Example && bundle exec pod install)`
   $?.exitstatus == 0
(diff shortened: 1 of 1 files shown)


scripts/repeat-process.js
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env node
```
k0nserv: Can we write automation in this repo in Ruby? Just makes it confusing if we are using both JS and Ruby across the board. Happy to help with this too

georgegillams: > This is a POC only - to help me establish a theory...

georgegillams: But yes - if it turns out that this works I plan to reimpliment it if possible 👍

Rakefile
```diff
@@ -60,16 +77,21 @@ task :analyze do
   sh "set -o pipefail && ! xcodebuild -workspace #{EXAMPLE_WORKSPACE} -scheme \"#{EXAMPLE_SCHEMA}\" -sdk #{BUILD_SDK} -destination \"platform=iOS Simulator,name=iPhone 8\" ONLY_ACTIVE_ARCH=NO analyze 2>&1 | xcpretty | grep -A 5 \"#{ANALYZE_FAIL_MESSAGE}\""
 end
 
+task :eraseDevices do
```
georgegillams: This has been found to help in other cases documented online of XCUI tests being flaky in CI.
I didn't want to make it part of the test command as it might be annoying to have devices wiped locally each time.

Rakefile
```diff
@@ -60,16 +77,21 @@ task :analyze do
...
+  sh "xcrun simctl erase all"
+end
+
 task :test do
   only_testing = FULL_TESTS ? '' : '-only-testing:Backpack_Tests'
-  sh "set -o pipefail && xcodebuild test -enableCodeCoverage YES -workspace #{EXAMPLE_WORKSPACE} -scheme \"#{EXAMPLE_SCHEMA}\" -sdk #{BUILD_SDK} -destination \"#{DESTINATION}\" #{only_testing} ONLY_ACTIVE_ARCH=NO | xcpretty"
+  testCommand = "set -o pipefail && xcodebuild test -enableCodeCoverage YES -workspace #{EXAMPLE_WORKSPACE} -scheme \"#{EXAMPLE_SCHEMA}\" -sdk #{BUILD_SDK} -destination \"#{DESTINATION}\" #{only_testing} ONLY_ACTIVE_ARCH=NO | xcpretty"
+  repeat_on_fail(testCommand)
```
georgegillams: The difference here is that the test process will now be repeated up to 3 times on failure.

Rakefile
```diff
@@ -46,6 +47,22 @@ def file_is_dirty(filename)
   $?.exitstatus != 0
 end
 
+def repeat_on_fail(command, runCount = 1)
+  if runCount > MAX_TEST_REPEATS then
```
k0nserv: You don't need `then` in ifs

Rakefile
```diff
@@ -46,6 +47,22 @@ def file_is_dirty(filename)
   $?.exitstatus != 0
 end
 
+def repeat_on_fail(command, runCount = 1)
```
k0nserv: Use snake case for variables

Rakefile
```diff
@@ -46,6 +47,22 @@ def file_is_dirty(filename)
...
+  if runCount > MAX_TEST_REPEATS then
+    puts "Process is still failing after #{MAX_TEST_REPEATS} attempts.\n Giving up."
+    exit(1);
+  end
+
+  puts "Running process. This is attempt #{runCount} of #{MAX_TEST_REPEATS}."
+  `#{command}`
+  if $?.exitstatus == 0 then
+    puts "Process succeeded on attempt #{runCount}."
+    exit(0);
```
k0nserv: This is redundant Rake already handles this

Rakefile
```diff
@@ -46,6 +47,22 @@ def file_is_dirty(filename)
   $?.exitstatus != 0
 end
 
+def repeat_on_fail(command, runCount = 1)
+  if runCount > MAX_TEST_REPEATS then
+    puts "Process is still failing after #{MAX_TEST_REPEATS} attempts.\n Giving up."
+    exit(1);
```
k0nserv: Use `abort` instead

Rakefile
k0nserv: It would be a bit better if we only retried tests that we know can be flakey i.e. UI tests not unit test or snapshot tests

Rakefile
k0nserv: Can we use a snake case name?

georgegillams:
https://github.com/Skyscanner/backpack-ios/pull/206#discussion_r266950618

I looked into this - also only rerunning failed tests, but it's not trivial as you have to create a testing profile with the tests you want etc...

Rakefile
georgegillams: I looked into this - also only rerunning failed tests, but it's not trivial as you have to create a testing profile with the tests you want etc...

Rakefile
shaundon: > documented online

Do you have a link?

PR #218, opened 2019-03-20
Layout update issue fixed

Szabadember:
This change moves the update of the calendar's collectionview from the beginning of a size transition to the end which is the right time to reload the data. This fixes the split screen issues and some weird behaviour on screen rotation.


k0nserv:
image
This passes locally so I am merging and releasing

PR #222, opened 2019-03-21
[NO-JIRA] Fix debug view hierarchy

georgegillams:
Instead of hackily replacing the window to enable shake gesture, I've subclassed UIApplication and added the logic there.


Example/Backpack/main.m
```diff
@@ -22,6 +22,6 @@
 int main(int argc, char * argv[])
 {
     @autoreleasepool {
-        return UIApplicationMain(argc, argv, nil, NSStringFromClass([BPKAppDelegate class]));
+        return UIApplicationMain(argc, argv, @"BPKExampleApp", NSStringFromClass([BPKAppDelegate class]));
```
k0nserv: Can we use `NSStringFromClass([BPKExampleApp class])` please?

Example/Backpack/BPKExampleApp.m
```diff
@@ -0,0 +1,37 @@
...
+ * limitations under the License.
+ */
+
+#import "BPKExampleApp.h"
+#import <Backpack/Color.h>
+
+@implementation BPKExampleApp
+
+- (void)sendEvent:(UIEvent *)event
+{
```
k0nserv: Can we fix up the code style in this method?

Example/Backpack/BPKExampleApp.h
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#import <Foundation/Foundation.h>
+@import UIKit;
```
k0nserv: Can we stick to `#import <UIKit/UIKit.h>` for the time being? We might conclude that we should use the `@import` syntax more, but let's make that an explicit decision if so

Example/Backpack/BPKExampleApp.h
```diff
@@ -0,0 +1,23 @@
...
+
+@interface BPKExampleApp : UIApplication
```
k0nserv: This is missing `NS_ASUSME_NONNULL_BEGIN`/`NS_ASSUME_NONNULL_END` markers

Example/Backpack/BPKExampleApp.h
shaundon: I didn't even know about this syntax!

k0nserv: It was introduce a while back and is specific to frameworks(I think?). See https://stackoverflow.com/questions/18947516/import-vs-import-ios-7

PR #223, opened 2019-03-22
[BPK-2339] Refactor `BPKThemeContainerController`

k0nserv:
Refactored to be a proper container view controller that can wrap an
arbitrary controller hierarchy. See [Implementing a Container View
Controller](https://developer.apple.com/library/archive/featuredarticles/ViewControllerPGforiPhoneOS/ImplementingaContainerViewController.html)

## Todo

+ [x] Review best practices for implementing container view controller and ensure we haven't missed anything. The above link is a good start.
+ [x] Write docs


Example/Backpack/BPKAppDelegate.m
```diff
@@ -33,8 +35,15 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
 
     ShakeableWindow *shakeableWindow = [[ShakeableWindow alloc] initWithFrame:[[UIScreen mainScreen] bounds]];
     shakeableWindow.windowLevel = UIWindowLevelAlert;
-    shakeableWindow.rootViewController = self.window.rootViewController;
-    self.shakeableWindow = shakeableWindow;
+    [BPKLondonTheme apply];
```
georgegillams: Unrelated, but I wonder if the wording `apply` is misleading. Maybe the API should be `[BPKLondonTheme enable];`?

k0nserv: I think we should probably rethink this API a bit, but I'd suggest we do that in another PR. Having both `BPKXTheme` and `BPKXThemeCotainer` seems non-ideal to me. My thoughts right now is that it would be nicer if `BPKXTheme` had a method that returned an instance of a theme container for that theme. This can also be formalized in a protocol so consumers can define their own themes in a predictable way

shaundon:
Worth noting that if we merge this in its current state the theme will be enabled and there's currently no UI to disable it, which is probably undesirable.

k0nserv:
> Worth noting that if we merge this in its current state the theme will be enabled and there's currently no UI to disable it, which is probably undesirable.

George will pick that up this week so should be remedied quickly

PR #231, opened 2019-03-29
New selection mechanism where every selection after a range defined c…

Szabadember:
…lears the previously selected range

Selection mechanism changed to behave the same way it does on backpack-android. Every selection after a range defined clears the previously selected range.


Example/Podfile.lock
```diff
@@ -28,4 +28,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 4043bcb10dcbd63bc0cb3ee1dadfdc93629dafcd
 
-COCOAPODS: 1.6.0.beta.2
+COCOAPODS: 1.6.1
```
Szabadember: Why do you use a beta version if there is a stable one available?

PR #233, opened 2019-03-31
[BPK-2339] Programatically show settings from settings button

georgegillams:
We want to programatically show the settings UI when the settings button is clicked so that we can wrap it in a theme container.

### DIFF
--- a/Example/Backpack/Base.lproj/Main.storyboard
+++ b/Example/Backpack/Base.lproj/Main.storyboard
@@ -393,11 +393,7 @@
                         </connections>
                     </tableView>
                     <navigationItem key="navigationItem" title="Backpack" id="N7k-3Q-Qpj">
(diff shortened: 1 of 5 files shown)


Example/Backpack/Base.lproj/Main.storyboard
```diff
@@ -393,11 +393,7 @@
                         </connections>
                     </tableView>
                     <navigationItem key="navigationItem" title="Backpack" id="N7k-3Q-Qpj">
-                        <barButtonItem key="rightBarButtonItem" id="oJ6-zT-C2i">
-                            <connections>
-                                <segue destination="419-up-khy" kind="presentation" id="MgG-WK-vR0"/>
```
georgegillams: Basically just removing the segue as we'll do this programatically

PR #234, opened 2019-03-31
[BPK-2339] Add theme helpers and settings

georgegillams:


Example/Utils/Settings.swift
```diff
@@ -0,0 +1,47 @@
...
+
+        UserDefaults.standard.set(themeName, forKey: "theme")
+    }
+
+    @objc class func activeTheme() -> ThemeHelpers.ThemeName {
+        guard let currentTheme = UserDefaults.standard.value(forKey: "theme") as? String else {
+            return .none
+        }
+
+        switch currentTheme {
```
k0nserv: This isn't required instead we can use an enum with a a raw value of `String` in combination with 

```swift
ThemeName(rawValue: currentTheme) ?? .none
```

Example/Utils/Settings.swift
```diff
@@ -0,0 +1,47 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Foundation
+
+class Settings: NSObject {
```
k0nserv: Can we make this an instance that has the `UserDefaults` injected? That way we can write a few tests for it too. We can expose a singleton as `Settings.sharedSettings` that is instantiated with `UserDefaults.standard`

Example/Utils/Settings.swift
georgegillams: Yeah but Objective C 😢

Example/Utils/Settings.swift
```diff
@@ -0,0 +1,55 @@
...
+
+    var userDefaults: UserDefaults = UserDefaults.standard
+    @objc static let sharedSettings = Settings(userDefaults: UserDefaults.standard)
+
+    init (userDefaults: UserDefaults) {
+        super.init()
+            self.userDefaults = userDefaults
+    }
+
+    @objc func setActiveTheme(theme: ThemeHelpers.ThemeName) {
```
k0nserv: Can this be a property?

Example/Utils/Settings.swift
```diff
@@ -0,0 +1,55 @@
...
+ * limitations under the License.
+ */
+
+import Foundation
+
+let THEMEKEY = "theme"
+
+class Settings: NSObject {
+
+    var userDefaults: UserDefaults = UserDefaults.standard
```
k0nserv: We don't need to set this value here and it can be a `let` property

Example/Utils/Settings.swift
georgegillams: Once BPKAppDelegate has been rewritten we can do this fo sho!

PR #248, opened 2019-04-03
[BPK-2270] Add theme definition to container controller API

georgegillams:
So we realised that the ThemeContainer being agnostic to the ThemeDefinition was a bit of an oversight. This changes the API to take a themeDefinition. It then creates the appropriate container itself. (Previously consumers would create a themeContainer from the themeDefinition themselves, and pass the container in).

This simplifies the API, and also means we can expose the current themeDefinition.

### DIFF
--- a/Backpack/Theme/Classes/BPKThemeContainerController.h
+++ b/Backpack/Theme/Classes/BPKThemeContainerController.h
@@ -19,6 +19,8 @@
 #import <Foundation/Foundation.h>
 #import <UIKit/UIKit.h>
 
+#import "BPKThemeDefinition.h"
+
 NS_ASSUME_NONNULL_BEGIN
 
 /**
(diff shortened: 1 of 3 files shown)


Backpack/Theme/Classes/BPKThemeContainerController.h
```diff
@@ -56,12 +58,19 @@ NS_SWIFT_NAME(ThemeContainerController) @interface BPKThemeContainerController :
  */
 @property(nonatomic, assign, getter=isThemeActive) BOOL themeActive;
 
+/**
+ * The theme container that is currently being used if the theme is active.
+ * This value is readonly. In order to change the themeContainer, the
```
shaundon:
```suggestion
 * This value is read only. In order to change the theme container, the
```

or 

```suggestion
 * This value is `readonly`. In order to change the `themeContainer`, the
```

PR #249, opened 2019-04-03
[NO-JIRA] Add docs build to travis

georgegillams:
We realised that we weren't catching breaks to the docs building process in CI as it only gets built when a release is done. This PR adds docs building to the CI process (but not deploying) so that we catch errors earlier.


.travis.yml
```diff
@@ -34,6 +34,7 @@ script:
   - npm run build
   - 'if [[ "$MAIN_STAGE" = true ]]; then ./scripts/check-pristine-state package-lock.json; fi'
   - 'if [[ "$MAIN_STAGE" = true ]]; then Example/Pods/SwiftLint/swiftlint lint --strict --config Example/.swiftlint.yml; fi'
+  - 'if [[ "$MAIN_STAGE" = true && "$TRAVIS_TAG" ]]; then ./scripts/build-docs-ci && rm -rf docs; fi'
```
georgegillams: This check basically does the same as the line below 🤦‍♂️

PR #254, opened 2019-04-04
[BPK-2270] Primary color utility spike 2

georgegillams:
Food for thought:
 - We can traverse the view hierarchy of a given view to find out what theme is the most appropriate.
 - What should consumers do to ensure that changes to the view hierarchy result in the correct change to the primary color? Relying on `didChangeTheme` to trigger re-pulling the `primaryColor` will not be 100% reliable. Instead can they just respond to all view hierarchy changes?

 - Logic would probably be simpler if we attached themeDefinition to each ThemeContainerView. Then we could simply traverse the hierarchy to find an instance of BPKThemeContainer instead of needing to work out which ViewController is active.

### DIFF
--- a/Backpack/Theme/Classes/BPKDefaultThemeContainer.m
+++ b/Backpack/Theme/Classes/BPKDefaultThemeContainer.m
@@ -15,11 +15,18 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 #import "BPKDefaultThemeContainer.h"
+#import "BPKDefaultTheme.h"
 
 NS_ASSUME_NONNULL_BEGIN
 
 @implementation BPKDefaultThemeContainer
+
+- (id<BPKThemeDefinition>)themeDefinition {
+    return [BPKDefaultTheme new];
+}
+
 @end
 
 NS_ASSUME_NONNULL_END
--- a/Backpack/Theme/Classes/BPKTheme.m
+++ b/Backpack/Theme/Classes/BPKTheme.m
@@ -18,8 +18,11 @@
 
 #import "BPKTheme.h"
 #import "BPKDohaThemeContainer.h"
-#import "BPKThemeDefinition.h"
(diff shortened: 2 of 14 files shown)


Backpack/Theme/Classes/BPKDefaultThemeContainer.m
```diff
@@ -15,11 +15,22 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 #import "BPKDefaultThemeContainer.h"
+#import "BPKDefaultTheme.h"
 
 NS_ASSUME_NONNULL_BEGIN
 
+@interface BPKDefaultThemeContainer ()
```
georgegillams: Can we remove these empty blocks?

Backpack/Theme/Classes/BPKTheme.m
```diff
@@ -18,8 +18,11 @@
 
 #import "BPKTheme.h"
 #import "BPKDohaThemeContainer.h"
-#import "BPKThemeDefinition.h"
+#import "BPKThemeContainer.h"
+#import "BPKThemeContainerController.h"
```
georgegillams: If we remove this line, everything breaks. I don't know why

Backpack/Theme/Classes/UIView+BPKThemeContainer.h
```diff
@@ -0,0 +1,33 @@
...
+#import <UIKit/UIKit.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@class BPKThemeContainer;
+@interface UIView (BPKThemeContainer)
+
+/**
+ * The nearest ancestor in the view controller hierachy that is a theme container controller. This is `nil` if
+ * there are no theme container controllers among the receivers ancestors.
```
shaundon:
```suggestion
 * there are no theme container controllers among the receiver's ancestors.
```

PR #267, opened 2019-04-12
[BPK-2320] Forward selectors in Theme Container

k0nserv:
Some apps make assumptions about the selectors available on
`keyWindow.rootViewController`. `BPKThemeContainerController` now
handles this better in release builds by not crashing and forwarding the
selector. In debug builds this causes an assertion failure.

### DIFF
--- a/Backpack/Theme/Classes/BPKThemeContainerController.m
+++ b/Backpack/Theme/Classes/BPKThemeContainerController.m
@@ -127,13 +127,20 @@ NS_ASSUME_NONNULL_BEGIN
     return themeContainerController;
 }
 
-#pragma mark - Overriden methods
+#pragma mark - UIViewController methods
 
 - (nullable UIViewController *)childViewControllerForStatusBarStyle {
     return self.rootViewController;
 }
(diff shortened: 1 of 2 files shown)


zsoltvarnai:
I think this helps a lot to minimize risks of crashes and other weird behavior. Do you think we also need to notify the app about it (via delegate or something), so it can log an error event? That would help us catch all cases in production, not just the ones the developers go through during debug.

k0nserv:
> I think this helps a lot to minimize risks of crashes and other weird behavior. Do you think we also need to notify the app about it (via delegate or something), so it can log an error event? That would help us catch all cases in production, not just the ones the developers go through during debug.

In theory this should have no negative impact in release and we'll eventually find all cases where this happens via the assert, but if you wish we can add a delegate to achieve the logging in release too

zsoltvarnai:
Yeah, it might be fine to start with only this and see how many assertion failures and surprises do we have during testing.

k0nserv:
Let's go for that 👍

PR #268, opened 2019-04-12
[BPK-2320] Fix race condition in Switch theming

k0nserv:


Backpack/Switch/Classes/BPKSwitch.h
```diff
@@ -42,6 +42,16 @@ NS_SWIFT_NAME(Switch) IB_DESIGNABLE @interface BPKSwitch : UISwitch
  */
 - (instancetype)initWithFrame:(CGRect)frame NS_DESIGNATED_INITIALIZER;
 
+
+/**
+ * The primary color of the reciever. This is used to control the
+ * `onTintColor`.
+ *
+ * @warning This is not intended to be used directly, it exsists to support theming only.
```
shaundon:
```suggestion
 * @warning This is not intended to be used directly, it exists to support theming only.
```

PR #273, opened 2019-04-18
[BPK-2380] Add theming support to chip

georgegillams:

Screenshot 2019-04-18 at 11 34 11

Screenshot 2019-04-18 at 11 34 05

### DIFF
--- a/Backpack/Chip/Classes/BPKChip.m
+++ b/Backpack/Chip/Classes/BPKChip.m
@@ -40,6 +40,8 @@ NS_ASSUME_NONNULL_BEGIN
     self = [super initWithFrame:frame];
 
     if (self) {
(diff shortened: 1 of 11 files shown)


Backpack/Chip/Classes/BPKChip.m
```diff
@@ -155,7 +168,7 @@ - (void)setUpConstraints {
 
 - (void)updateStyle {
     if (self.selected) {
-        self.backgroundColor = BPKColor.blue500;
+        self.backgroundColor = self.primaryColor;
```
k0nserv: Since it's `nullable` maybe we should also have a default here?

georgegillams: I'll just make them non-nullable as they shouldn't be anyway 👍

PR #277, opened 2019-04-22
[NO-JIRA] Fix calendar header background to be transparent

jorgedc93:
Remove the white background in the Calendar header to allow it to be transparent


k0nserv:
It looks good to me, @Szabadember since you wrote a lot of this code can you weigh in?

PR #280, opened 2019-04-29
[No JIRA] Update jazzy version

k0nserv:

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -75,7 +75,7 @@ GEM
     redcarpet (3.4.0)
     rouge (2.0.7)
     ruby-macho (1.4.0)
-    sass (3.7.3)
+    sass (3.7.4)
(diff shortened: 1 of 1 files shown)


Gemfile.lock
```diff
@@ -75,7 +75,7 @@ GEM
     redcarpet (3.4.0)
     rouge (2.0.7)
     ruby-macho (1.4.0)
-    sass (3.7.3)
+    sass (3.7.4)
```
georgegillams: Couldn't this just be `~> 3.7`?

k0nserv: This is in `Gemfile.lock`. Remember the purpose of the lock file is to ensure all versions are pinned exactly so it doesn't support semver patterns

georgegillams: Oh yeah - that makes sense. File names are helpful 😆

PR #283, opened 2019-04-30
[BPK-2102] Fix button heights and font weight

georgegillams:

### DIFF
--- a/Backpack/Button/Classes/BPKButton.m
+++ b/Backpack/Button/Classes/BPKButton.m
@@ -88,6 +88,11 @@ NS_ASSUME_NONNULL_BEGIN
     self.titleLabel.textAlignment = NSTextAlignmentCenter;
     self.titleLabel.tintAdjustmentMode = UIViewTintAdjustmentModeNormal;
 
+    // Use this to get the title value if one has been set in storyboard.
+    if (self.titleLabel != nil && self.titleLabel.text != nil) {
+        _title = self.titleLabel.text;
+    }
+
     self.size = size;
     self.style = style;
     self.imagePosition = BPKButtonImagePositionTrailing;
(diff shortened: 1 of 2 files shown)


georgegillams:
An alternative (better?) way to fix this might be to override `setTitle: forState` and `setAttributedTitle: forState` to set the `title` value, but this gets a bit complicated because we already rely on those methods in our code, so we end up with some weird side-effects.

I think this is therefore the cleanest solution.
@k0nserv and @shaundon wdyt?

georgegillams:
Screenshot 2019-04-30 at 15 05 39

Backpack/Button/Classes/BPKButton.m
```diff
@@ -88,6 +88,11 @@ - (void)setupWithSize:(BPKButtonSize)size style:(BPKButtonStyle)style {
     self.titleLabel.textAlignment = NSTextAlignmentCenter;
     self.titleLabel.tintAdjustmentMode = UIViewTintAdjustmentModeNormal;
 
+    // Use this to get the title value if one has been set in storyboard.
+    if (self.titleLabel != nil && self.titleLabel.text != nil) {
```
k0nserv: I am confused as to how this resolve the problem 🤔 Can you expand a bit?

georgegillams: Yeah - it's because we set the title of the disabled ones in the interface-builder. That results in calling `setTitle: forState` or `setAttributedStyle: forState`.
But we don't support those, so our button ignores them. As a result, our button behaves as if it's icon-only and doesn't properly format the text.

k0nserv: Right and the reason it thinks it's icon only is because `_title` is `nil`?

georgegillams: Yeah

georgegillams: @k0nserv any other suggestions of how we could fix this? Or do you think this is a reasonable approach?

k0nserv: Nah I think it sounds reasonable

PR #285, opened 2019-05-01
[BPK-2427] Add direction in primary gradient

k0nserv:
All `BPKGradient` already contain a builtin in direction component in
their definiton. However when we render the primary gradient we
sometimes do so with an alternative direction. We want to be able to
theme primary gradients, while still supporting direction overrides for
specific cases.


Backpack/Gradient/Classes/BPKGradient.m
```diff
@@ -184,6 +194,7 @@ + (CGPoint)endPointForDirection:(BPKGradientDirection)direction {
     case BPKGradientDirectionBottomRight:
         return CGPointMake(1.0, 1.0);
     default:
+        NSAssert(NO, @"Invalid `BPKGradientDirection` value %lu", (unsigned long)direction);
```
shaundon: `%lu` is 'unsigned long' to save others a google 😇

Backpack/Gradient/Classes/BPKGradient.m
```diff
@@ -61,6 +61,15 @@ - (instancetype)initWithColors:(NSArray<UIColor *> *)colors
     return self;
 }
 
+- (instancetype)cloneWithNewDirection:(BPKGradientDirection)newDirection {
```
georgegillams: Why do we need `cloneWithNewDirection`. Would it make more sense for consumers to just clone a gradient and set the direction on the resulting view?

k0nserv: I am confused, can you share some code examples?

georgegillams: For example, why can't a user do something like this:
```
let gradient = Gradient();
let someOtherGradient = gradient.clone()
someOtherGradient.directionOverride = GradientDirectionTopLeft
```

georgegillams: I would expect the API to allow me to just set the direction on the gradient instance

k0nserv: This lets you do that, but it prefers a readonly API which is much nicer than a mutable API as you suggested

PR #290, opened 2019-05-08
[NO-JIRA] Add assertion that primary colour view is in hierarchy

georgegillams:

### DIFF
--- a/Backpack/Theme/Classes/UIView+BPKThemeContainer.m
+++ b/Backpack/Theme/Classes/UIView+BPKThemeContainer.m
@@ -24,6 +24,8 @@ NS_ASSUME_NONNULL_BEGIN
 @implementation UIView (BPKThemeContainer)
 
 - (nullable BPKThemeContainer *)themeContainer {
+    NSAssert(self.window != nil, @"To get a theme container from a view, the view must be within the view hierarchy.");
+
     UIView *parent = self;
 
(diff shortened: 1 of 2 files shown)


Backpack/Theme/Classes/UIView+BPKThemeContainer.m
```diff
@@ -26,6 +26,9 @@ @implementation UIView (BPKThemeContainer)
 - (nullable BPKThemeContainer *)themeContainer {
     UIView *parent = self;
 
+    NSAssert(parent.window != nil,
```
georgegillams: hmmm... this doesn't behave as I expected.....

Example/Backpack/Views/BPKTableViewSelectableCell.swift
```diff
@@ -41,7 +41,9 @@ class BPKTableViewSelectableCell: UITableViewCell {
     }
 
     override func didMoveToWindow() {
-        tickIcon.tintColor = Theme.primaryColor(for: self)
+        if self.window != nil {
```
georgegillams: I'm not sure why this check is needed. It appears that they are only outside the window when I change theme...

georgegillams: Overriding `didMoveToSuperview` has the same downside....

PR #294, opened 2019-05-10
Bpk 2398 add bpk font theming support 2

georgegillams:


Backpack/Button/Classes/BPKButton.m
```diff
@@ -57,6 +57,10 @@ - (instancetype)initWithSize:(BPKButtonSize)size style:(BPKButtonStyle)style {
     return self;
 }
 
+- (void)didMoveToWindow {
```
georgegillams: Instead of this hack we should use `setFontFamily`. This should work, as placing this into a different view-hierarchy will cause the `setter` to fire.

Backpack/Label/Classes/BPKLabel.m
```diff
@@ -59,28 +59,49 @@ - (instancetype)initWithFrame:(CGRect)frame {
...
-        attributedString = [BPKFont attributedStringWithFontStyle:self.fontStyle content:text];
+        attributedString = [BPKFont attributedStringWithFontStyle:self.fontStyle content:_textV forView:self];
     }
     super.attributedText = attributedString;
 }
 
+- (void)setText:(NSString *_Nullable)text {
+    BPKAssertMainThread();
+    if (text != _textV) {
+        _textV = text;
```
georgegillams: Rename `_textV` to something more intuitive

PR #297, opened 2019-05-16
[BPK-2480] Gray theming utility

georgegillams:
Screenshot 2019-05-17 at 11 21 51

Screenshot 2019-05-17 at 11 21 45


georgegillams:
The only `BPKTheme` method I have added requires a number and view. ie `gray(number:50 for:view)`
Do you think I should add some convenience methods for each gray? ie `gray50(for:view)`, `gray300(for:view)` etc

PR #302, opened 2019-05-23
[NO-JIRA] Enable calendar gray theming

georgegillams:

Screenshot with an outrageous set of grays applied:
image


georgegillams:
@k0nserv and @shaundon - wdyt about theming grays like this? Feels better to me to stick to our existing convention of one UI_APPEARANCE_SELECTOR per property, but then we're gonna have twice as many selectros 🤔

k0nserv:
I'd rather expose the invidual properties I think

PR #303, opened 2019-05-24
[NO-JIRA] Make dialog icon optional

georgegillams:
Make dialog icon and background color optional. Because why not 🤷‍♂️

image


Backpack/Dialog/Classes/BPKDialogView.m
```diff
@@ -45,6 +45,8 @@ @interface BPKDialogView ()
 @property(nonatomic, strong) BPKLabel *titleLabel;
 @property(nonatomic, strong) BPKLabel *descriptionLabel;
 @property(nonatomic, strong) UIStackView *buttonStackView;
+@property(nonatomic, strong) UIColor *iconBackgroundColorValue;
```
georgegillams: I'm not sure if using these underlying properties is the best way to respond to them being set... 🤔

phelgo: Agreed. If I understand correctly we should be able to call `setIconBackgroundColor` directly without adding state.

Backpack/Dialog/Classes/BPKDialogView.m
```diff
@@ -219,20 +256,26 @@ - (BPKDialogButtonAction *_Nullable)actionForButton:(BPKButton *)button {
 
 - (void)setIconBackgroundColor:(UIColor *_Nullable)iconBackgroundColor {
     BPKAssertMainThread();
-    self.iconView.iconBackgroundColor = iconBackgroundColor;
+    if (_iconBackgroundColorValue != iconBackgroundColor) {
+        _iconBackgroundColorValue = iconBackgroundColor;
+        [self layoutIconView];
```
phelgo: Do we need to call `layoutIconView` if we are just updating the color?

Backpack/Dialog/Classes/BPKDialogView.m
georgegillams: Think I'll refactor this to use another object, so that consumers must provide neither or both 👍

Backpack/Dialog/Classes/BPKDialogController.h
```diff
@@ -100,8 +100,8 @@ NS_SWIFT_NAME(DialogController)
 + (instancetype)dialogControllerWithTitle:(NSString *)title
                                   message:(NSString *)message
                                     style:(BPKDialogControllerStyle)style
-                      iconBackgroundColor:(UIColor *)iconBackgroundColor
-                                iconImage:(UIImage *)iconImage;
+                      iconBackgroundColor:(UIColor *_Nullable)iconBackgroundColor
```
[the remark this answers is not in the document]
georgegillams: Think I'll refactor this to use another object, so that consumers must provide neither or both 👍

Backpack/Dialog/Classes/BPKDialogIconView.m
```diff
@@ -49,19 +49,16 @@ - (CGSize)intrinsicContentSize {
     return [[self class] viewSize];
 }
 
-- (void)setIconBackgroundColor:(UIColor *_Nullable)iconBackgroundColor {
-    self.iconContainerView.backgroundColor = iconBackgroundColor;
+- (void)setIconDefinition:(BPKDialogIconDefinition *)iconDefinition {
+//    self.iconContainerView.backgroundColor = iconDefinition.iconBackgroundColor;
```
georgegillams: For somereason it blows up on these lines, but I have no idea why 😢 
Feel like I've been looking at it too long to see the wood for the trees. Probably something simple that I'm missing @k0nserv @phelgo ?

georgegillams: @k0nserv solved it for me 👍

georgegillams:
Think this might be good to roll now @k0nserv and @phelgo 🎉

Backpack/Dialog/Classes/BPKDialogIconDefinition.h
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * The background color behind the icon
+ */
+@property(nonatomic, strong) UIColor *iconBackgroundColor;
+
+/**
+ * The icon to show at the top of the dialog
+ */
+@property(nonatomic, strong) UIImage *iconImage;
```
k0nserv: These should be `readonly` imo. Immutable objects are much simpler to reason about.

Backpack/Dialog/Classes/BPKDialogIconView.m
```diff
@@ -49,19 +49,16 @@ - (CGSize)intrinsicContentSize {
     return [[self class] viewSize];
 }
 
-- (void)setIconBackgroundColor:(UIColor *_Nullable)iconBackgroundColor {
-    self.iconContainerView.backgroundColor = iconBackgroundColor;
+- (void)setIconDefinition:(BPKDialogIconDefinition *)iconDefinition {
+    self.iconContainerView.backgroundColor = iconDefinition.iconBackgroundColor;
```
k0nserv: This needs to update the local variable `_iconDefintion` too and it should have the regular  

```objective-c
if (iconDefintion != _iconDefintion) {
}
```

guards

Backpack/Dialog/Classes/BPKDialogIconView.h
```diff
@@ -15,13 +15,13 @@
...
  */
+#import "BPKDialogIconDefinition.h"
 #import <UIKit/UIKit.h>
 
 NS_ASSUME_NONNULL_BEGIN
 
 @interface BPKDialogIconView : UIView
-@property(nullable, nonatomic, strong) UIColor *iconBackgroundColor;
-@property(nullable, nonatomic, strong) UIImage *iconImage;
+@property(nonatomic, strong) BPKDialogIconDefinition *iconDefinition;
```
k0nserv: This should be `nullable`

Backpack/Dialog/Classes/BPKDialogIconView.m
```diff
@@ -49,19 +49,16 @@ - (CGSize)intrinsicContentSize {
...
+    self.iconImageView.image = iconDefinition.iconImage;
 }
 
-- (UIColor *_Nullable)iconBackgroundColor {
+- (UIColor *)iconBackgroundColor {
```
k0nserv: This is a lie as this value can still be `nil`. Lying is naughty

Backpack/Dialog/Classes/BPKDialogIconView.m
```diff
@@ -49,19 +49,16 @@ - (CGSize)intrinsicContentSize {
...
     return self.iconContainerView.backgroundColor;
 }
 
-- (void)setIconImage:(UIImage *_Nullable)iconImage {
-    self.iconImageView.image = iconImage;
-}
-
-- (UIImage *_Nullable)iconImage {
+- (UIImage *)iconImage {
```
k0nserv: This is also a lie

Backpack/Dialog/Classes/BPKDialogIconView.m
```diff
@@ -49,19 +49,16 @@ - (CGSize)intrinsicContentSize {
     return [[self class] viewSize];
 }
 
-- (void)setIconBackgroundColor:(UIColor *_Nullable)iconBackgroundColor {
-    self.iconContainerView.backgroundColor = iconBackgroundColor;
+- (void)setIconDefinition:(BPKDialogIconDefinition *)iconDefinition {
```
k0nserv: This argument should be `_Nullable`. If the user sets it to `nil` after it was not `nil` we need to ensure the correct thing happens

k0nserv:
I am sorry it took so long for me to get to this :/

Backpack/Dialog/Classes/BPKDialogIconView.h
georgegillams: I'm not sure it should... That would mean we allow an icon view to exist without an icon definition.

Think it should be non-null, set in the constructor

Backpack/Dialog/Classes/BPKDialogIconDefinition.m
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+#import "BPKDialogIconDefinition.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@implementation BPKDialogIconDefinition
+
+- (instancetype)initWithIcon:(UIImage *)iconImage iconBackgroundColor:(UIColor *)iconBackgroundColor {
+    _iconImage = iconImage;
```
k0nserv: This isn't calling super or checking the value of `self`.

Backpack/Dialog/Classes/BPKDialogIconView.m
```diff
@@ -49,18 +49,18 @@ - (CGSize)intrinsicContentSize {
     return [[self class] viewSize];
 }
 
-- (void)setIconBackgroundColor:(UIColor *_Nullable)iconBackgroundColor {
-    self.iconContainerView.backgroundColor = iconBackgroundColor;
+- (void)setIconDefinition:(BPKDialogIconDefinition *_Nullable)iconDefinition {
+    // TODO handle iconDefinition nil
```
k0nserv: This still isn't done

Backpack/Dialog/Classes/BPKDialogView.m
```diff
@@ -177,16 +171,52 @@ - (void)setupConstraints {
...
+            // blends with the contentView background. Using BPKSpacingLg makes it look unaligned with the BPKSpacingLg
+            // below. The border is BPKSpacingMd wide which when added to BPKSpacingBase makes it BPKSpacingLg. This
+            // ligns up nicely with the BPKSpacingLg between descriptionLabel and buttonStackView
+            [self.titleLabel.topAnchor constraintGreaterThanOrEqualToAnchor:self.iconView.bottomAnchor
+                                                                   constant:BPKSpacingBase]
+        ]];
+    }
+
+    if (!self.hasIcon && self.iconView != nil) {
+        [self.iconView removeFromSuperview];
```
k0nserv: This is a bit surprising in a method called `setupIconView` since it does the opposite of setting up

Backpack/Dialog/Classes/BPKDialogIconView.m
georgegillams: Shouldn't need to handle `nil` case as the `iconView` should always have an `iconDefinition`!

Backpack/Dialog/Classes/BPKDialogIconView.h
k0nserv:
```swift
let iconImage = BPKDialogIconView() // calls `init`
// this will crash because the type in swift is incorrectly `DialogIconDefinition`
// when it should be `DialogIconDefinition?`
iconImage.iconDefinition.iconImage 
```

Backpack/Dialog/Classes/BPKDialogView.m
```diff
@@ -53,21 +54,23 @@ @implementation BPKDialogView
...
         self.message = message;
-        self.iconBackgroundColor = iconBackgroundColor;
-        self.iconImage = iconImage;
+        self.iconDefinition = iconDefinition;
     }
 
     return self;
 }
 
+- (Boolean)hasIcon {
```
k0nserv: BOOL

k0nserv: Shouldn't this be delcared as a readonly property too?

georgegillams: Tempted to setup a danger rule for this as it always catches me out!

Backpack/Dialog/Classes/BPKDialogIconView.m
```diff
@@ -34,7 +34,7 @@ + (CGSize)iconContainerSize;
 
 @implementation BPKDialogIconView
 
-- (instancetype)init {
+- (instancetype)initWithIconDefinition:(BPKDialogIconDefinition *)iconDefinition {
```
k0nserv: `iconDefinition` is never used here

georgegillams: Excellent catch!

Backpack/Dialog/Classes/BPKDialogIconView.m
```diff
@@ -34,12 +34,13 @@ + (CGSize)iconContainerSize;
...
     CGSize size = [[self class] viewSize];
     self = [super initWithFrame:CGRectMake(0, 0, size.width, size.height)];
 
     if (self) {
         [self setupViews];
+        self.iconDefinition = iconDefinition;
```
k0nserv: Wouldn't it be better to set this without using the setter i.e. the `ivar` before calling `setupViews` so that `setupViews` can rely on the value being present?

Backpack/Dialog/Classes/BPKDialogIconView.m
georgegillams: Fixed 👍

Backpack/Dialog/Classes/BPKDialogIconView.m
georgegillams: Done

PR #336, opened 2019-07-02
[NO-JIRA] Label dark mode support

georgegillams:
Screenshot 2019-07-02 at 18 39 43

Screenshot 2019-07-02 at 18 39 48

+ [x] ~`UNRELEASED.md`~ -> We shouldn't shout about this yet!
+ [x] ~`README.md`~ -> ⬆️  Likewise
+ [x] ~Docs~ -> ⬆️  Likewise
+ [x] ~Tests~ -> Not really possible until we're using XCode 11

### DIFF
--- a/Backpack/Color/Classes/Generated/BPKColor.m
+++ b/Backpack/Color/Classes/Generated/BPKColor.m
@@ -17,6 +17,7 @@
  * limitations under the License.
  */
 #import "BPKColor.h"
+#import <Backpack/DarkMode.h>
 
 @implementation BPKColor
 
@@ -178,6 +179,21 @@
 
 + (UIColor *)clear {
   return UIColor.clearColor;
+}
+
++ (UIColor *) label {
+    return [BPKColor colorForUserInterfaceStyleWithLightVariant:BPKColor.skyGray darkVariant:BPKColor.white];
+}
+
++ (UIColor *)colorForUserInterfaceStyleWithLightVariant:(UIColor *)lightColor darkVariant:(UIColor*)darkColor {
+#if __BPK_DARK_MODE_SUPPORTED
+    if (@available(iOS 13.0, *)) {
+        if (UITraitCollection.currentTraitCollection.userInterfaceStyle == UIUserInterfaceStyleDark){
+            return darkColor;
+        }
+    }
+#endif
+    return lightColor;
 }
 
 + (UIColor *)blend:(UIColor*)firstColor with:(UIColor*)secondColor weight:(double)weight {
--- a/Backpack/Label/Classes/BPKLabel.m
+++ b/Backpack/Label/Classes/BPKLabel.m
@@ -20,6 +20,7 @@
 
 #import <Backpack/Color.h>
 #import <Backpack/Common.h>
+#import <Backpack/DarkMode.h>
 
 #import "BPKTextDefinition.h"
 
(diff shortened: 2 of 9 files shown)


Backpack/Label/Classes/BPKLabel.m
```diff
@@ -100,11 +100,22 @@ - (void)setFontMapping:(BPKFontMapping *_Nullable)fontMapping {
     }
 }
 
+- (void)traitCollectionDidChange:(UITraitCollection *_Nullable)previousTraitCollection {
+    [super traitCollectionDidChange:previousTraitCollection];
```
georgegillams: This code is also gonna be used a lot. Might be worth creating a `Delegate` that components can adhere to to get notified of when they actually need to update appearance.

Example/Backpack/Labels.storyboard
```diff
@@ -1,11 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<document type="com.apple.InterfaceBuilder3.CocoaTouch.Storyboard.XIB" version="3.0" toolsVersion="14313.18" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" useSafeAreas="YES" colorMatched="YES" initialViewController="Gdg-nX-jAF">
-    <device id="retina4_7" orientation="portrait">
-        <adaptation id="fullscreen"/>
-    </device>
+<document type="com.apple.InterfaceBuilder3.CocoaTouch.Storyboard.XIB" version="3.0" toolsVersion="14810.12" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" useSafeAreas="YES" colorMatched="YES" initialViewController="Gdg-nX-jAF">
```
georgegillams: These changes are entirely backwards-compatible!

Backpack/Color/Classes/Generated/BPKColor.m
```diff
@@ -292,6 +292,15 @@ + (UIColor *)clear {
   return UIColor.clearColor;
 }
 
++ (UIColor *) label {
+if (@available(iOS 13.0, *)) {
```
georgegillams: This code is gonna be used a lot. Probably worth having a helper method that takes the two colours and spits back the right one.

Backpack/Color/Classes/Generated/BPKColor.m
```diff
@@ -292,6 +292,17 @@ + (UIColor *)clear {
   return UIColor.clearColor;
 }
 
++ (UIColor *) label {
+#if __IPHONE_OS_VERSION_MAX_ALLOWED >= 130000
```
georgegillams: This code is gonna be used a lot. Probably worth having a helper method that takes the two colours and spits back the right one.

georgegillams: What this logic basically says is:
```
 - macro: if I'm compiling for iOS 13 then include the following code:
   - code if ios version is 13 then:
     - check whether dark mode is on or off 
   - end code if
 - end macro if
```

shaundon: Do you know why it's `130000` out of interest and not `13`?

Backpack/Label/Classes/BPKLabel.m
```diff
@@ -100,11 +100,24 @@ - (void)setFontMapping:(BPKFontMapping *_Nullable)fontMapping {
...
+
+#if __IPHONE_OS_VERSION_MAX_ALLOWED >= 130000
+    if (@available(iOS 13.0, *)) {
+        UITraitCollection *traitCollection = [UITraitCollection currentTraitCollection];
```
k0nserv: We should use `self.traitCollection` instead since it's specific to the current instance.

Backpack/Color/Classes/Generated/BPKColor.m
k0nserv: Can we create a macro for this?

georgegillams: The version here is actually not the ios version number. It's the number of kitten videos that were watched by Apple engineers whilst building iOS. When the iOS 13 beta came out, the number of kitty videos had just passed 130,000.

georgegillams: You mean to create a macro for just for line 196, right?

Backpack/Label/Classes/BPKLabel.m
```diff
@@ -100,11 +101,24 @@ - (void)setFontMapping:(BPKFontMapping *_Nullable)fontMapping {
...
+    }
+#endif
+}
+
 #pragma mark - Private
 
 - (void)setupWithStyle:(BPKFontStyle)style {
     self.fontStyle = style;
-    self.textColor = BPKColor.gray700;
+    self.textColor = BPKColor.label;
```
georgegillams: If the consumer has set the colour themselves then we should respect that here instead of changing to the default colour for the current appearance...

PR #338, opened 2019-07-03
[BPK-2596] Fix Tappable Label protocol naming

georgegillams:

+ [-] `UNRELEASED.md`
+ [x] `README.md`
+ [-] Tests
+ [-] Docs (either update [backpack-docs](https://github.com/Skyscanner/backpack-docs) now, or create a follow up ticket)

### Still todo:
 - [x] Fix issue where text is cut off

### DIFF
--- a/Backpack/TappableLinkLabel/Classes/BPKTappableLinkLabel.m
+++ b/Backpack/TappableLinkLabel/Classes/BPKTappableLinkLabel.m
@@ -261,11 +261,19 @@ NS_ASSUME_NONNULL_BEGIN
     if ([self.delegate respondsToSelector:@selector(attributedLabel:didSelectLinkWithURL:)]) {
         [self.delegate attributedLabel:label didSelectLinkWithURL:url];
     }
+
+    if ([self.delegate respondsToSelector:@selector(tappableLabel:didSelectLinkWithURL:)]) {
+        [self.delegate tappableLabel:label didSelectLinkWithURL:url];
+    }
 }
 
 - (void)attributedLabel:(BPKTappableLinkLabel *)label didSelectLinkWithTransitInformation:(NSDictionary *)components {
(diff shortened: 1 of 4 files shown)


Backpack/TappableLinkLabel/Classes/BPKTappableLinkLabelDelegate.h
```diff
@@ -25,6 +25,8 @@ NS_ASSUME_NONNULL_BEGIN
 @optional
 - (void)attributedLabel:(BPKTappableLinkLabel *)label didSelectLinkWithURL:(NSURL *)url;
 - (void)attributedLabel:(BPKTappableLinkLabel *)label didSelectLinkWithTransitInformation:(NSDictionary *)components;
```
k0nserv: Can we add a comment about these two being deprecated?

Backpack/TappableLinkLabel/Classes/BPKTappableLinkLabel.m
```diff
@@ -261,12 +261,20 @@ - (void)attributedLabel:(BPKTappableLinkLabel *)label didSelectLinkWithURL:(NSUR
     if ([self.delegate respondsToSelector:@selector(attributedLabel:didSelectLinkWithURL:)]) {
         [self.delegate attributedLabel:label didSelectLinkWithURL:url];
     }
+
+    if ([self.delegate respondsToSelector:@selector(tappableLabel:didSelectLinkWithURL:)]) {
```
k0nserv: I feel like we should only call one of these with a pereference for `tappableLabel:didSelectLinkWithURL:`

georgegillams: Like this?
```
if (new_one_is_supported) {
  do new_one
} else if (old_one_is_supported) {
  do old_one
}
```

k0nserv: Yeah that's what I am thinking

PR #344, opened 2019-07-11
[BPK-2460] Add Horizontal Navigation component

georgegillams:
image

+ [x] `UNRELEASED.md`
+ [ ] `README.md`
+ [ ] Tests

Follow up PR's to add:
 - [x] Boolean property to optionally show underline
 - [ ] A version that supports icons too

### DIFF
--- a/Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigation.h
+++ b/Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigation.h
@@ -1,0 +1,75 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2018-2019 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#import "BPKHorizontalNavigationDelegate.h"
+#import <Backpack/Font.h>
+#import <Foundation/Foundation.h>
+#import <UIKit/UIKit.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ * `BPKHorizontalNavigation` is a subclass of `UIControl`.
+ * The `createHorizontalNavigationItem` method can be overridden to create cutom looking cells.
+ * By default it will just return a new instance of `BPKHorizontalNavigationItem`.
+ */
+@class BPKHorizontalNavigationItem;
+@class BPKHorizontalNavigationOption;
+NS_SWIFT_NAME(HorizontalNavigation) IB_DESIGNABLE @interface BPKHorizontalNavigation : UIControl
+
+@property(nullable, nonatomic, strong) UIColor *selectedColor UI_APPEARANCE_SELECTOR;
+@property(nullable, nonatomic, strong) BPKFontMapping *fontMapping UI_APPEARANCE_SELECTOR;
+
+/**
+ * The options to display within the navigation.
+ *
+ * see BPKHorizontalNavigationOption
+ */
+@property(nonatomic, copy) NSArray<BPKHorizontalNavigationOption *> *options;
+
+/**
+ * Whether to show a bar under the selected item
+ */
+@property(nonatomic) Boolean showSelectedBar;
+
+/**
+ * The index of the currently selected item.
+ * By default the first item is selected.
+ * If the index is out of bounds for the options available (eg -1) then no item will be selected.
+ */
+@property(nonatomic) NSInteger selectedItemIndex;
+
+/**
+ * A delegate that can be informed of changes to the selection.
+ */
+@property(nonatomic, weak) id<BPKHorizontalNavigationDelegate> delegate;
+
+/**
+ * Create a `BPKHorizontalNavigation` with a set of options and an optionaly selected option.
+ *
+ * @param options NSArray<NSString> the options available to the useR
+ * @param selectedItemIndex NSString the initially selected item
+ */
+- (instancetype)initWithOptions:(NSArray<NSString *> *)options
+                       selected:(NSInteger)selectedItemIndex NS_DESIGNATED_INITIALIZER;
+
+- (instancetype)initWithCoder:(NSCoder *)aDecoder NS_DESIGNATED_INITIALIZER;
+- (instancetype)initWithFrame:(CGRect)frame NS_DESIGNATED_INITIALIZER;
+
+@end
+NS_ASSUME_NONNULL_END
--- a/Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigation.h
+++ b/Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigation.h
@@ -16,20 +16,20 @@
  * limitations under the License.
  */
 
-#import "BPKHorizontalNavigationDelegate.h"
-#import <Backpack/Font.h>
 #import <Foundation/Foundation.h>
 #import <UIKit/UIKit.h>
+#import <Backpack/Font.h>
+
+#import "BPKHorizontalNavigationDelegate.h"
 
 NS_ASSUME_NONNULL_BEGIN
 
-/**
- * `BPKHorizontalNavigation` is a subclass of `UIControl`.
- * The `createHorizontalNavigationItem` method can be overridden to create cutom looking cells.
- * By default it will just return a new instance of `BPKHorizontalNavigationItem`.
- */
 @class BPKHorizontalNavigationItem;
 @class BPKHorizontalNavigationOption;
+
+/**
+ * A `BPKHorizontalNavigation` is a control comprising of multiple segments, where each acts as a discrete button.
+ */
 NS_SWIFT_NAME(HorizontalNavigation) IB_DESIGNABLE @interface BPKHorizontalNavigation : UIControl
 
 @property(nullable, nonatomic, strong) UIColor *selectedColor UI_APPEARANCE_SELECTOR;
@@ -45,7 +45,7 @@ NS_SWIFT_NAME(HorizontalNavigation) IB_DESIGNABLE @interface BPKHorizontalNaviga
 /**
  * Whether to show a bar under the selected item
  */
-@property(nonatomic) Boolean showSelectedBar;
+@property(nonatomic) BOOL showsSelectedBar;
 
 /**
  * The index of the currently selected item.
--- a/Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigation.h
+++ b/Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigation.h
@@ -18,6 +18,7 @@
 
 #import <Foundation/Foundation.h>
 #import <UIKit/UIKit.h>
+
 #import <Backpack/Font.h>
 
 #import "BPKHorizontalNavigationDelegate.h"
--- a/Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigation.h
+++ b/Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigation.h
@@ -60,10 +60,10 @@ NS_SWIFT_NAME(HorizontalNavigation) IB_DESIGNABLE @interface BPKHorizontalNaviga
 @property(nonatomic, weak) id<BPKHorizontalNavigationDelegate> delegate;
 
 /**
- * Create a `BPKHorizontalNavigation` with a set of options and an optionaly selected option.
+ * Create a `BPKHorizontalNavigation` with a set of options and an initally selected option.
  *
  * @param options NSArray<NSString> the options available to the useR
- * @param selectedItemIndex NSString the initially selected item
+ * @param selectedItemIndex NSInteger the initially selected item
  */
 - (instancetype)initWithOptions:(NSArray<NSString *> *)options
                        selected:(NSInteger)selectedItemIndex NS_DESIGNATED_INITIALIZER;
--- a/Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigation.m
+++ b/Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigation.m
@@ -78,11 +78,13 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (void)setSelectedItemIndex:(NSInteger)selectedItemIndex {
     if (self.options.count > 0) {
-        NSString *assertionErrorMessage =
+#if DEBUG
+        NSString *errorMessage =
             [NSString stringWithFormat:@"selectedItemIndex must be within range of the number of options available. "
                                        @"The number of options is %lu but the index selected was %ld",
                                        (unsigned long)self.options.count, (long)selectedItemIndex];
-        NSAssert(selectedItemIndex >= 0 && selectedItemIndex < self.options.count, assertionErrorMessage);
+        NSAssert(selectedItemIndex >= 0 && selectedItemIndex < self.options.count, errorMessage);
+#endif
     }
 
     if (_selectedItemIndex != selectedItemIndex) {
--- a/Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigation.m
+++ b/Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigation.m
@@ -191,11 +191,6 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (void)setupWithOptions:(NSArray<BPKHorizontalNavigationOption *> *)options selected:(NSInteger)selectedItemIndex {
 
-    self.barView = [UIView new];
-    [self updateBarColor];
-    self.barView.translatesAutoresizingMaskIntoConstraints = NO;
-    [self addSubview:self.barView];
-
     [NSLayoutConstraint activateConstraints:@[
         [self.barView.topAnchor constraintEqualToAnchor:self.bottomAnchor constant:-BPKSpacingSm / 2],
         [self.barView.leadingAnchor constraintGreaterThanOrEqualToAnchor:self.leadingAnchor],
@@ -212,6 +207,11 @@ NS_ASSUME_NONNULL_BEGIN
     self.stackView = [UIStackView new];
     self.stackView.translatesAutoresizingMaskIntoConstraints = NO;
     [self addSubview:self.stackView];
+
+    self.barView = [UIView new];
+    [self updateBarColor];
+    self.barView.translatesAutoresizingMaskIntoConstraints = NO;
+    [self addSubview:self.barView];
 
     [NSLayoutConstraint activateConstraints:@[
         [self.stackView.topAnchor constraintEqualToAnchor:self.topAnchor constant:-BPKSpacingSm],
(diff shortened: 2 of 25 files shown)


Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigation.h
```diff
@@ -0,0 +1,53 @@
...
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ * `BPKHorizontalNavigation` is a subclass of `UIStackView`.
+ * dequeueReusableCell method can be overridden to create cutom looking cells. By default it will just return a new
+ * instance of BPKHorizontalNavigationCell
+ */
+@class BPKHorizontalNavigationCell;
+@class BPKHorizontalNavigationDelegate;
+NS_SWIFT_NAME(HorizontalNavigation) IB_DESIGNABLE @interface BPKHorizontalNavigation : UIStackView
```
k0nserv: I kind of feel like this should be a `UIControl` subclass like `UISegmentedControl` is. At the very least making it a `UIStackView` subclass feels like an abstraction leak so if it isn't `UIControl` it should be `UIView`

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigation.h
```diff
@@ -0,0 +1,53 @@
...
+
+@property(nullable, nonatomic, strong) BPKFontMapping *fontMapping UI_APPEARANCE_SELECTOR;
+@property(nullable, nonatomic, strong) UIColor *selectedColor UI_APPEARANCE_SELECTOR;
+
+@property(nullable, nonatomic, strong) NSArray<NSString *> *options;
```
k0nserv: This should be `copy`

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigation.h
```diff
@@ -0,0 +1,53 @@
...
+@property(nullable, nonatomic, strong) NSString *selectedItem;
```
k0nserv: An indexed based approach for the selected option feels nicer imo

georgegillams: Yeah that seems reasonable 👍

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigationDelegate.h
```diff
@@ -0,0 +1,37 @@
...
+
+@optional
+
+/*
+ * Tells the delegate that the user has selected one of the horizontal navigation options.
+ *
+ * @param horizontalNavigation The `BPKHorizontalNavigation` which has been interacted with.
+ * @param item The option selected by the user.
+ */
+- (void)horizontalNavigation:(BPKHorizontalNavigation *)horizontalNavigation didSelectItem:(NSInteger *)item;
```
k0nserv: Can we rename this to include the term `index` too?

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigationItem.h
```diff
@@ -0,0 +1,54 @@
...
+#import <Backpack/Button.h>
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ * `BPKHorizontalNavigationItem` is a subclass of `UICollectionViewCell`.
+ */
+@class BPKHorizontalNavigationOption;
+NS_SWIFT_NAME(HorizontalNavigationItem) IB_DESIGNABLE @interface BPKHorizontalNavigationItem : BPKButton
```
k0nserv: Not sure about inherting from `BPKButton` here, it feels like much more potential hassel than it's worth

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigationItem.h
```diff
@@ -0,0 +1,54 @@
...
+
+@property(nullable, nonatomic, strong) UIColor *selectedColor UI_APPEARANCE_SELECTOR;
+
+/**
+ * The definition of the option, which will be displayed as a button in the component.
+ *
+ * see BPKHorizontalNavigationOption
+ */
+@property(nullable, nonatomic, strong) BPKHorizontalNavigationOption *definition;
```
k0nserv: Any reason this isn't readonly or even not saved?

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigationItem.m
```diff
@@ -0,0 +1,104 @@
...
+    if (_selectedColor != selectedColor) {
+        _selectedColor = selectedColor;
+
+        [self updateStyle];
+    }
+}
+
+- (void)setupWithDefinition:(BPKHorizontalNavigationOption *_Nullable)definition {
+    self.style = BPKButtonStyleLink;
+    _definition = definition;
```
k0nserv: Use the property rather than the variable

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigationItem.m
```diff
@@ -0,0 +1,104 @@
...
+
+    self.title = definition.name;
+    self.translatesAutoresizingMaskIntoConstraints = NO;
```
k0nserv: I don't believe this is required

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigationOption.h
```diff
@@ -0,0 +1,34 @@
...
+
+#import <Foundation/Foundation.h>
+#import <Backpack/Icon.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface BPKHorizontalNavigationOption : NSObject
+
+@property(readonly, strong, nullable) BPKIconName iconName;
+@property(readonly, strong) NSString *name;
```
k0nserv: This should be copy

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigationOption.m
```diff
@@ -0,0 +1,51 @@
...
+
+NS_ASSUME_NONNULL_BEGIN
+
+@implementation BPKHorizontalNavigationOption
+
+- (instancetype)initWithName:(NSString *)name {
+    self = [super init];
+
+    if (self) {
+        _name = name;
```
k0nserv: I would suggest we redefine the property in an anonymous interface and use the property instead of the ivar

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigation.m
```diff
@@ -0,0 +1,173 @@
...
+
+    return self;
+}
+
+- (instancetype)initWithCoder:(NSCoder *)aDecoder {
+    BPKAssertMainThread();
+    self = [super initWithCoder:aDecoder];
+
+    if (self) {
+        [self setupWithOptions:@[] selected:-1];
```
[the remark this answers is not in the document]
georgegillams: I think at the least we should make 0 the default

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigationDelegate.h
georgegillams: `didSelectItem:(NSInteger *)index`
or
`didSelectItemIndex:(NSInteger *)itemIndex`
?

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigationOption.m
georgegillams: I've tried that but it feels a bit nasty tbh

georgegillams:
Just added the bar into this PR too 🎉

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigationItem.h
georgegillams: Do you reckon we should go for a custom `UIButton` then?

k0nserv: Yeah that sounds reasonable

georgegillams: tbh I kinda feel like using our button might make sense given that we want to support icons too...

georgegillams: I'll give it a go and see where I get to!

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigationItem.h
```diff
@@ -0,0 +1,59 @@
...
+ */
+
+#import "BPKHorizontalNavigation.h"
+#import <Backpack/Button.h>
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ * `BPKHorizontalNavigationItem` is a subclass of `UICollectionViewCell`.
```
k0nserv: These docs are incorrect and should be placed immediately adjacent to the `@interface` declaration. Can we improve the doc string a bit too?

k0nserv: Although arguably this is not public since it will not be exported to the users directly so maybe we should leave it undocumented, at least the top level doc comment which will hide it from Jazzy

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigationItem.h
```diff
@@ -0,0 +1,59 @@
...
+ * see BPKFontMapping
+ */
+@property(nonatomic) BPKFontMapping *fontMapping;
+
+/**
+ * The definition of the option, which will be displayed as a button in the component.
+ *
+ * see BPKHorizontalNavigationOption
+ */
+@property(readonly, nullable, nonatomic, strong) BPKHorizontalNavigationOption *definition;
```
k0nserv: Does this need to be retained at all? Can't it just be used to construct an instance or even just to extract the instance to call a relevant initializer here?

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigationItem.m
```diff
@@ -0,0 +1,118 @@
...
+    [self setAttributedTitle:titleString forState:UIControlStateNormal];
+
+    if (self.selected) {
+        self.accessibilityTraits = UIAccessibilityTraitButton | UIAccessibilityTraitSelected;
+    } else {
+        self.accessibilityTraits = UIAccessibilityTraitButton;
+    }
+}
+
+- (UIColor *)getContentColor {
```
k0nserv: Can we make this a private `readonly` property instead? The `get` prefix stands out like a sore thumb in Objective-C

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigationItem.m
```diff
@@ -0,0 +1,118 @@
...
+    BPKFontStyle fontStyle = BPKFontStyleTextBase;
+
+    NSAttributedString *titleString = [BPKFont attributedStringWithFontStyle:fontStyle
+                                                                     content:self.definition.name
+                                                                   textColor:[self getContentColor]
+                                                                 fontMapping:self.fontMapping];
+    [self setAttributedTitle:titleString forState:UIControlStateNormal];
+
+    if (self.selected) {
+        self.accessibilityTraits = UIAccessibilityTraitButton | UIAccessibilityTraitSelected;
```
k0nserv: Does `UIButton`/`UIControl` not already handle this when we call `[super setSelected:selected]`?

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigationOption.h
```diff
@@ -0,0 +1,34 @@
...
+NS_ASSUME_NONNULL_BEGIN
+
+@interface BPKHorizontalNavigationOption : NSObject
+
+@property(readonly, strong, nullable) BPKIconName iconName;
+@property(readonly, copy) NSString *name;
+@property(readonly) BOOL hasIcon;
+
+- (instancetype)initWithName:(NSString *)name;
+- (instancetype)initWithName:(NSString *)name iconName:(BPKIconName _Nullable)iconName;
```
k0nserv: Is there a point in making the second argument `_Nullable` here? That's what the first initializer is for after all

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigationOption.m
```diff
@@ -0,0 +1,57 @@
...
+
+    return self;
+}
+
+- (instancetype)initWithName:(NSString *)name iconName:(BPKIconName _Nullable)iconName {
+    self = [super init];
+
+    if (self) {
+        self.name = name;
+        _iconName = iconName;
```
k0nserv: Why are we being inconsistent in the use of ivar and property here?

Example/Backpack/View Controllers/HorizontalNavSelectorViewController.swift
```diff
@@ -0,0 +1,40 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import UIKit
+import Backpack
+
+enum HorizontalNavSegueIdentifier: String {
+    case def = "Def"
```
k0nserv: Is this default? If so can we name it as such, `def = "Def"` needlessly obfuscates this

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigation.h
```diff
@@ -0,0 +1,75 @@
...
+
+/**
+ * Whether to show a bar under the selected item
+ */
+@property(nonatomic) Boolean showSelectedBar;
+
+/**
+ * The index of the currently selected item.
+ * By default the first item is selected.
+ * If the index is out of bounds for the options available (eg -1) then no item will be selected.
```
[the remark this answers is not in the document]
georgegillams: It is

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigation.h
```diff
@@ -0,0 +1,75 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#import "BPKHorizontalNavigationDelegate.h"
+#import <Backpack/Font.h>
+#import <Foundation/Foundation.h>
+#import <UIKit/UIKit.h>
```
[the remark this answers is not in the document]
georgegillams: Clang places them in this order, and for me consistency (ie what clang-format can provide) is more important.

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigation.h
```diff
@@ -0,0 +1,75 @@
...
+ */
+@property(nonatomic) NSInteger selectedItemIndex;
+
+/**
+ * A delegate that can be informed of changes to the selection.
+ */
+@property(nonatomic, weak) id<BPKHorizontalNavigationDelegate> delegate;
+
+/**
+ * Create a `BPKHorizontalNavigation` with a set of options and an optionaly selected option.
```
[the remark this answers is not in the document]
georgegillams: I guess it could be out of range? So nothing is selected as per this.
https://github.com/Skyscanner/backpack-ios/pull/344#discussion_r305971861

We need to decide on whether this is allowed or not

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigation.m
```diff
@@ -0,0 +1,273 @@
...
+    for (int i = 0; i < self.stackView.arrangedSubviews.count; i += 1) {
+        NSAssert([self.stackView.arrangedSubviews[i] isKindOfClass:[BPKHorizontalNavigationItem class]],
+                 @"HorizontalNav subview is not of type BPKHorizontalNavigationItem as expected.");
+        if (![self.stackView.arrangedSubviews[i] isKindOfClass:[BPKHorizontalNavigationItem class]]) {
+            continue;
+        }
+
+        BPKHorizontalNavigationItem *navigationCell = (BPKHorizontalNavigationItem *)self.stackView.arrangedSubviews[i];
+
+        if (navigationCell != nil) {
```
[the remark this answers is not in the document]
georgegillams: it wouldn't

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigation.m
```diff
@@ -0,0 +1,273 @@
...
+        return self.barView.backgroundColor = BPKColor.blue500;
+    }
+}
+
+- (void)setupWithOptions:(NSArray<BPKHorizontalNavigationOption *> *)options selected:(NSInteger)selectedItemIndex {
+
+    self.barView = [UIView new];
+    [self updateBarColor];
+    self.barView.translatesAutoresizingMaskIntoConstraints = NO;
+    [self addSubview:self.barView];
```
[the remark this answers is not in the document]
georgegillams: I have no idea. Will ensure that this is checked when I write tests

Example/Backpack/View Controllers/HorizontalNavSelectorViewController.swift
georgegillams: `default` is a keyword

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigationItem.m
georgegillams: Yeah I think this is no longer needed since I stopped using BPKButton (when I'd implemented my own variation of `selected` because BPKButton disallows it!)

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigation.h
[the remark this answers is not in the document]
k0nserv: > I've fixed them for now but going forward if we want a specific ordering (other than what Clang does) we should look into tooling that can enforce this

Won't Clang just override this orderign anyway? I don't fully know what clang format does here, if we group them logically with new lines I am happy with clang format formatting those groups

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigation.h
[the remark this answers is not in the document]
k0nserv: But the `selectedItemIndex` is still not optional here, in fact it can't be since it's not a pointer type

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigation.m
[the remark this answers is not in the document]
k0nserv: I would just do this, it could cause subtle and hard to debug issues down the line

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigation.h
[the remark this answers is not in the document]
georgegillams: Clang just bunches them together in alphabetical order. There don't seem to be any clang options to configure this behaviour.

> Won't Clang just override this orderign anyway?
Not if I don't run it 😛

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigation.m
```diff
@@ -0,0 +1,277 @@
...
+
+    [NSLayoutConstraint activateConstraints:@[
+        [self.barView.topAnchor constraintEqualToAnchor:self.bottomAnchor constant:-BPKSpacingSm / 2],
+        [self.barView.leadingAnchor constraintGreaterThanOrEqualToAnchor:self.leadingAnchor],
```
k0nserv: Can we keep the bar hidden while it's not being displayed so that these aren't required to resolve ambigious layout?

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigation.m
```diff
@@ -0,0 +1,277 @@
...
+- (void)setOptions:(NSArray<BPKHorizontalNavigationOption *> *)options {
+    if (_options != options) {
+        _options = [options copy];
+
+        [self repopulateStackview];
+    }
+}
+
+- (void)setItemSelectionStates {
+    for (int i = 0; i < self.stackView.arrangedSubviews.count; i += 1) {
```
k0nserv: Can we use the `forEach` helper here?

PR #354, opened 2019-07-24
[BPK-2745] Replace 'trips' icon

shaundon:
Simulator Screen Shot - iPhone Xʀ - 2019-07-24 at 10 05 11


shaundon:
I don't think it's breaking - it's just changing the look of an icon. I'd consider design changes to be breaking if they required user intervention, like if the size of something changed.

Also, Chris vouched for it:

> This replaces the existing 'trips' icon, which was used on the old web-trips feature. This change can be made with no issues.

Worth noting that the old trips icon is still available as 'baggage' so if anybody has issues there's a simple fix.

k0nserv:
Okay cool

PR #363, opened 2019-07-31
Add tag property to horizontal navigation

georgegillams:

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigation.m
+++ b/Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigation.m
@@ -132,9 +132,10 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)updateSelection:(UIButton *)sender {
     NSInteger newIndex = [self.stackView.arrangedSubviews indexOfObject:sender];
     [self setSelectedItemIndex:newIndex];
+    BPKHorizontalNavigationItem *navItem = (BPKHorizontalNavigationItem *)sender;
 
-    if ([self.delegate respondsToSelector:@selector(horizontalNavigation:didSelectItem:)]) {
-        [self.delegate horizontalNavigation:self didSelectItem:newIndex];
+    if ([self.delegate respondsToSelector:@selector(horizontalNavigation:didSelectItem:withTag:)]) {
+        [self.delegate horizontalNavigation:self didSelectItem:newIndex withTag:navItem.definition.tag];
     }
 }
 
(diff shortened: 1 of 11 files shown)


Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigation.m
```diff
@@ -132,9 +132,10 @@ - (void)adjustBarConstraintsForView:(UIView *)view {
 - (void)updateSelection:(UIButton *)sender {
     NSInteger newIndex = [self.stackView.arrangedSubviews indexOfObject:sender];
     [self setSelectedItemIndex:newIndex];
+    BPKHorizontalNavigationItem *navItem = (BPKHorizontalNavigationItem *)sender;
```
georgegillams: Should be done more safely!

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigationDelegate.h
```diff
@@ -31,7 +31,9 @@ NS_ASSUME_NONNULL_BEGIN
  * @param horizontalNavigation The `BPKHorizontalNavigation` which has been interacted with.
  * @param item The option selected by the user.
  */
-- (void)horizontalNavigation:(BPKHorizontalNavigation *)horizontalNavigation didSelectItem:(NSInteger)index;
```
k0nserv: The docs are out of date

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigationOption.h
```diff
@@ -25,9 +25,12 @@ NS_ASSUME_NONNULL_BEGIN
 @interface BPKHorizontalNavigationOption : NSObject
 
 @property(nonatomic, readonly, strong, nullable) BPKIconName iconName;
+@property(nonatomic, readonly) NSInteger tag;
```
k0nserv: What's the default value for `tag` if you don't use one of the constructors that sets it. Can we use `0` to align with `UIView.tag`?

georgegillams: The default is `0` for a UIView that hasn't had tag set 👍

k0nserv: But we set it unconditionally when creating the horizontal item. We never set the value of `tag` in `BPKHorizontalNavigationOption` when a constructor without tag is used

georgegillams: Yeah but the tag value that is handed back comes from the UIView.

georgegillams: And the default value for an NSInteger is `0` anyway

k0nserv: > And the default value for an NSInteger is 0 anyway
This is not true, `NSIntger` in fact any primitive type is not guranteed to have a stable or deterministic value

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigationDelegate.h
k0nserv: The `item` param is outdated in the doc string still

PR #382, opened 2019-08-21
Investigate auto updating bpk dependencies

georgegillams:

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


package.json
```diff
@@ -54,6 +55,7 @@
     "lodash": "^4.17.11",
     "markdown-spellcheck": "^1.3.1",
     "merge2": "^1.2.3",
+    "npm-check-updates": "^3.1.21",
```
shaundon: This sounds good, but isn't that the point of Greenkeeper? I looked at [the package](https://www.npmjs.com/package/npm-check-updates) and it looks like it serves the same purpose.

georgegillams: Yes, but it doesn't serve the purpose because it's too slow 😆

shaundon: Is it though? When we release a new `bpk-tokens` Greenkeeper sends us updates within the day

PR #383, opened 2019-08-21
[BPK 2834] dev iOS make dialog title optional

imchristian:
## Approach
I've made the title property nullable and with that change I was able to evaluate if we have or not a title to set the correct top anchor with the icon image.
Originally, if we didn't add a title (e.g. " " string) the description label was misplaced (see image bellow).

## Other discarded approach
One of the ideas was basically `removeFromSuperview` the `titleLabel` if it was `nil` but that's not so good for the runtime performance, instead, we just set the correct anchor values if we have an `iconDefinition`

## Future considerations
In the future would be great if we can make a refactor to encapsulate all the values (e.g. `title`, `description`, `icon`, etc) in one Model to be injected through the class initializer in the `BPKDialogView.h`.
I think that this approach could give us a more clean code and eliminate the possibility that someone else can change the properties values in runtime after that the dialog has been presented.

Group 2
Group 3
Group

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


UNRELEASED.md
```diff
@@ -1,5 +1,9 @@
 # Unreleased
 > Place your changes below this line.
+**Added:**
+
+- Backpack/Dialogs:
```
georgegillams:
```suggestion
- Backpack/Dialog:
```

Backpack/Dialog/Classes/BPKDialogView.m
```diff
@@ -305,5 +300,20 @@ - (void)addButtonAction:(BPKDialogButtonAction *)action {
     [self setLayoutMargins];
 }
 
+#pragma mark - Private methods
+
+// This method helps us to set up the correct top anchor margin depending if we have a title or not
+- (void)updateTopAnchorConstraint {
+    [NSLayoutConstraint activateConstraints:@[
+        [[self hasTitle] ? self.titleLabel.topAnchor : self.descriptionLabel.topAnchor
```
georgegillams: We tend to prefer using property syntax over method syntax
```suggestion
        [self.hasTitle ? self.titleLabel.topAnchor : self.descriptionLabel.topAnchor
```

Backpack/Dialog/Classes/BPKDialogView.m
```diff
@@ -192,14 +192,9 @@ - (void)updateIconView {
 
             [self.contentView.topAnchor constraintGreaterThanOrEqualToAnchor:self.iconView.bottomAnchor
                                                                     constant:-(iconViewSize.height / 2.0)],
-
-            // NOTE: The constant here is explicitly not BPKSpacingLg because the icon view has a white border which
```
georgegillams: @k0nserv I personally think this comment can go, but want to check with you first.

k0nserv: I'd wanna keep it in content where we add the constraint

Backpack/Dialog/Classes/BPKDialogView.m
k0nserv: Will this not cause a bug if you unset and set the title multiple times since we never cleanup constraints? Also with this approach the `self.descriptionLabel.topAnchor` would have two competing constraints to other anchors? May I suggest that instead of this approach we keep the constraints as is, but add a height constriant that forces `titleLabel` to be `0pt` tall. With that we might also want to adjust the descriptionLabel top constraint with a negative constant

imchristian: I had different thoughts about this approach because I don't know why we have setters as I said in the PR comment in **Future Considerations**. 
Usually, those properties don't have to change once the Dialog is already presented I think or at least I don't remember that kind of requirements for a UI component in my experience, because if we expose those properties instead of injecting them through the designated initializer we need to react as you said to all the changes and run autolayout again and that also is not so cost-effective in terms of run time processing 

For now I can improve my current approach and experiment with your suggestion but I think that in the future a refactor would be better to avoid entirely this kind of situations.

k0nserv: Cool, let's keep using this approach then. I'd like to make sure we aren't ending up with two competing constraints and ensure that we don't add hundreds of constraints if people keep setting and unsetting the title. 

I appreciate that this is unlikely to happen, but it feels like a subtle bug that we'd have to fight with down the line

imchristian: Also, the height is automatically set to 0 to the `titleLabel` if it doesn't have content or is nil so I think that if we change the descriptionLabel top constraint should be enough, I'll try it out

georgegillams:
Test

georgegillams:
Test

PR #400, opened 2019-08-30
Enable running the app with no theme enabled

georgegillams:
When the example app starts, it immediately applies a theme (even if that theme is `Default`).
This means we have no way of testing our components when un-themed.

As a result - bugs like this cannot be found:
Simulator Screen Shot - iPhone 8 - 2019-08-30 at 19 41 20

This change ensures that, if no theme is enabled on app start, no theme wrapper will be applied until theming is enabled.

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


k0nserv:
I think a better solution than this might be no including the theme container controller at al for this configuration. Possibly by using a different xcode schema

PR #410, opened 2019-09-09
[BPK-2928] Fix BPKCard background color bug

georgegillams:

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


Backpack/Card/Classes/BPKCard.m
```diff
@@ -108,7 +109,10 @@ - (void)setHighlighted:(BOOL)highlighted {
     BPKAssertMainThread();
     [super setHighlighted:highlighted];
 
-    self.backgroundColor = highlighted ? BPKColor.gray100 : BPKColor.white;
+    [CATransaction begin];
+    [CATransaction setAnimationDuration:highlighted ? .2 : 0];
+    self.tintLayer.opacity = highlighted ? 0.2 : 0;
```
antalindisguise: In the Skyscanner app we generally use a 0.75 alpha value on top, so should we align that and use 0.25 here instead?

Backpack/Card/Classes/BPKCard.m
```diff
@@ -108,7 +109,10 @@ - (void)setHighlighted:(BOOL)highlighted {
     BPKAssertMainThread();
     [super setHighlighted:highlighted];
 
-    self.backgroundColor = highlighted ? BPKColor.gray100 : BPKColor.white;
+    [CATransaction begin];
+    [CATransaction setAnimationDuration:highlighted ? .2 : 0];
```
antalindisguise: minor, but is there any hard code style for using either `.2` or `0.2`?

Backpack/Card/Classes/BPKCard.m
georgegillams: What colour do you use to begin with. Is it gray900 at 0.75?

We should definitely create a reusable property for this...

PR #435, opened 2019-09-25
API changes in Rating component

imchristian:

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/Backpack/Rating/Classes/BPKRating.m
+++ b/Backpack/Rating/Classes/BPKRating.m
@@ -60,8 +60,8 @@ NS_ASSUME_NONNULL_BEGIN
 }
 
 - (instancetype)initWithRatingValue:(CGFloat)ratingValue
-                              title:(NSString *)title
-                           subtitle:(NSString *_Nullable)subtitle {
+                              title:(BPKRatingTextDefinition *)title
+                           subtitle:(BPKRatingTextDefinition *_Nullable)subtitle {
     self = [super initWithFrame:CGRectZero];
 
     if (self) {
@@ -99,7 +99,7 @@ NS_ASSUME_NONNULL_BEGIN
     if (_lowRatingColor != lowRatingColor) {
         _lowRatingColor = lowRatingColor;
 
-        self.ratingBubble.lowRatingColor = lowRatingColor;
+        [self updateStyle];
     }
 }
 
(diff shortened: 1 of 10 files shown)


Backpack/Rating/Classes/BPKRating.m
```diff
@@ -250,6 +249,16 @@ - (CGSize)systemLayoutSizeFittingSize:(CGSize)targetSize {
 
 - (void)updateStyle {
     self.ratingBubble.ratingValue = self.ratingValue;
+    if (self.ratingValue < 6.00) {
+        self.textWrapper.title = self.title.lowRatingText;
+        self.textWrapper.subtitle = self.subtitle.lowRatingText;
```
georgegillams: I wonder if we should move the colouring logic of the RatingBubble to here too, so that it's all in one place.
For example
```
if (self.ratingValue < 6.00) {
  ...
  self.ratingBubble.backgroundColor = BPKColor.panjin;
} else .....
```

georgegillams: wdyt?

Example/Backpack/View Controllers/RatingsViewController.swift
```diff
@@ -25,18 +25,48 @@ class RatingsViewController: UIViewController {
     var layout: BPKRatingLayout = .horizontal
 
     fileprivate static var ratingData = [
-        (ratingValue: 3.0, title: "Low title", subtitle: "Low subtitle",
+        (ratingValue: 3.0, titleDefinition: RatingTextDefinition(highRatingText: "",
```
georgegillams: Instead of these can we just define some variables `titleDefinition` and `subtitle definition` and use them for each of these?

Example/SnapshotTests/BPKRatingSnapshotTest.m
```diff
@@ -33,7 +33,7 @@ - (void)setUp {
 
 - (void)testRatingOutOfRangeLow {
     BPKRating *bpkRating = [[BPKRating alloc] initWithFrame:CGRectZero];
-    bpkRating.title = @"Capped";
+    bpkRating.title = [[BPKRatingTextDefinition alloc] initWithHighRatingText:@"" mediumRatingText:@"" lowRatingText:@"Capped"];
```
georgegillams: similar to my comment above, could we have a couple of pre-defined definitions that we use for most of these?

Backpack/Rating/Classes/BPKRating.m
imchristian: You're right I totally forgot that refactor

PR #455, opened 2019-10-15
Refactor dark mode

imchristian:
This is PR is to make our backpack example app full dark mode.
After this, we should start the design review of each component to make all the necessaries adjustments.

darkmodeRefactor

+ [ ] `UNRELEASED.md`
+ [ ] `README.md`
+ [ ] Tests


Example/Backpack/View Controllers/TappableLinkLabelsViewController.swift
```diff
@@ -48,7 +48,7 @@ class TappableLinkLabelsViewController: UIViewController, BPKTappableLinkLabelDe
     override func viewWillAppear(_ animated: Bool) {
         super.viewWillAppear(animated)
 
-        if self.style == .alternate {
+        if isAlternateStyleWithLightMode() {
```
georgegillams: Now that we support dark mode, maybe when we enter this view we should just call `self.overrideUserInterfaceStyle = UIUserInterfaceStyleDark;`

wdyt?

imchristian: I thought that we were using that other background color for other reasons but if it was just to always try on dark scenarios then yeah we can definitely delete this code and always adopt dark mode for this particular screen. 
I'll do the change 👍

georgegillams: Nice 👍

imchristian: This is no longer valid because our example app still needs to work on iOS prev to 13

PR #515, opened 2019-11-22
Improve Dialog in extreme cases

k0nserv:
Improve the dialog in extreme cases such as:

+ Large viewports with little content displayed.
+ Small viewports with a lot of content displayed.

## Small Viewport

<details>
<summary>Before</summary>

Simulator Screen Shot - iPhone SE - 2019-11-19 at 14 11 06

</details>

<details>
<summary>After</summary>

Simulator Screen Shot - iPhone SE - 2019-11-21 at 15 43 31
</details>

## Large Viewport

<details>
<summary>Before</summary>

Simulator Screen Shot - iPad Pro (12 9-inch) (3rd generation) - 2019-11-22 at 11 45 39
</details>

<details>
<summary>After</summary>

Simulator Screen Shot - iPad Pro (12 9-inch) (3rd generation) - 2019-11-22 at 11 48 32
</details>

## Snapshot changes

These are changes in our snapshot tests which I think we should accept because the component works more like we want now.

### Increased default flare height

The previous flare height was set to be between 200 and 300 pt tall however this was an ambiguous layout so the flare view didn't end up being 300pt. With these changes the flare will try to be 300pt tall until it has to shrink due to space constraints 

<details>
<summary>Before</summary>

reference_testDialogViewWithLargeCornerStyleAndFlareViewAndButtons@2x
reference_testDialogViewWithLargeCornerStyleAndFlareViewAndSmallButtons@2x

</details>

<details>
<summary>After</summary>

failed_testDialogViewWithLargeCornerStyleAndFlareViewAndButtons@2x
failed_testDialogViewWithLargeCornerStyleAndFlareViewAndSmallButtons@2x
</details>

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [ ] Tests


k0nserv:
When testing this please do two things:

1. Look for constraint violations in the console. I haven't spotted any.
2. Pause the view debugger and make sure there are no ambiguous layouts.

tiagoengel:
My only concern with this is that it goes agains what was implemented for Android where the dialog expands to fullscreen when the copy is too large.

k0nserv:
Yeah it's a good point @tiagoengel, however this implementation also restricts how large the dialog can grow when the viewport is large, see iPad screenshots. How does Android currently handle this?

Backpack/Dialog/Classes/BPKDialogContentView.h
```diff
@@ -0,0 +1,79 @@
...
+ */
+@property(nonatomic, readonly, assign) BOOL hasButtonActions;
+
+/**
+ * Whether the reiever has a non-empty title set.
+ */
+@property(nonatomic, readonly, assign) BOOL hasTitle;
+
+- (instancetype)initWithFrame:(CGRect)frame NS_DESIGNATED_INITIALIZER;
+- (nullable instancetype)initWithCoder:(NSCoder *)aDecoder NS_DESIGNATED_INITIALIZER;
```
georgegillams: `nullable instancetype`?

Backpack/Dialog/Classes/BPKDialogController.m
```diff
@@ -128,24 +128,36 @@ - (void)setupViews {
 }
 
 - (void)setupConstraints {
+    NSLayoutConstraint *lowerWidthConstraint = [self.dialogView.widthAnchor constraintGreaterThanOrEqualToAnchor:self.view.widthAnchor multiplier:0.8];
+    lowerWidthConstraint.priority = UILayoutPriorityDefaultHigh;
```
georgegillams: Does this need to be high? Can't see any other constraint that this needs to take precedence over?

georgegillams:
I think this is just failing snapshots due to the new, larger flare size?

Backpack/Dialog/Classes/BPKDialogContentView.h
k0nserv: That's what it says yes

Backpack/Dialog/Classes/BPKDialogController.m
k0nserv: Otherwise if you put very short content in the dialog it will be tiny. For example the title "A" and the message "A" results in a dialog that's like 50pt wide which looks daft

k0nserv:
> I think this is just failing snapshots due to the new, larger flare size?
Yes

imchristian:
@k0nserv @georgegillams can we merge this pr or we are waiting for some additional changes?

k0nserv:
Nope

PR #516, opened 2019-11-22
New NavigationBar buttons

imchristian:
NavigationBarButtons

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


Backpack/NavigationBar/Classes/BPKNavigationBar.m
```diff
@@ -259,6 +261,25 @@ - (UIView *)borderView {
     return _borderView;
 }
 
+- (BPKNavigationBarButton *)leftButton {
+    if (!_leftButton) {
+        _leftButton = [[BPKNavigationBarButton alloc] init];
+        _leftButton.translatesAutoresizingMaskIntoConstraints = NO;
+        _leftButton.imagePosition = BPKButtonImagePositionLeading;
+        _leftButton.hidden = YES;
```
georgegillams: Should this have a chevron too?

imchristian: We don't have currently the appropriate asset with the correct inner padding inside the `BPKIcon` source. Although in the app there's an asset called `backArrow` which could be injected inside the `BPKNavigationButton`, We need to include this eventually in the `BPKIcon` too.

georgegillams: Cool. Will leave it as is for now then.
Do you mind if I push a change so that we have an example with buttons and another without?

imchristian: That would be great thanks!

Backpack/NavigationBar/Classes/BPKNavigationBarButton.h
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+#import <Backpack/Button.h>
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface BPKNavigationBarButton : BPKButton
+
+@property(nonatomic, strong) UIColor *contentColor UI_APPEARANCE_SELECTOR;
```
georgegillams: Don't think this needs to be a UIAppearanceSelector

Backpack/NavigationBar/Classes/BPKNavigationBarTitleView.h
```diff
@@ -30,7 +30,6 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (instancetype)init __attribute__((unavailable("use initWithFrame: instead")));
 - (instancetype)initWithCoder:(NSCoder *)aDecoder __attribute__((unavailable("use initWithFrame: instead")));
-;
```
georgegillams: NICE! 👏

PR #529, opened 2019-11-28
[AMT-3954] Fix animation glitch when updating `HorizontalNavigation`

fedetrim:
This PR fixes an animation glitch that happens when updating the options in the `HorizontalNavigation`.

48e7d780-0aea-11ea-894b-300787413134

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [ ] Tests

### DIFF
--- a/Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigation.m
+++ b/Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigation.m
@@ -108,7 +108,8 @@ NS_ASSUME_NONNULL_BEGIN
         if (UIAccessibilityIsReduceMotionEnabled()) {
             animationDuration = 0.0;
         }
-
+        
+        [self layoutIfNeeded];
         [UIView animateWithDuration:animationDuration
                          animations:^{
                            [self adjustBarConstraintsForView:selectedButton];
@@ -257,6 +258,7 @@ NS_ASSUME_NONNULL_BEGIN
     // Remove all current arranged views:
     for (UIView *subView in self.stackView.arrangedSubviews) {
         [self.stackView removeArrangedSubview:subView];
+        [subView removeFromSuperview];
     }
 
     for (BPKHorizontalNavigationOption *option in self.options) {
@@ -274,7 +276,7 @@ NS_ASSUME_NONNULL_BEGIN
 }
 
 - (void)setOptions:(NSArray<BPKHorizontalNavigationOption *> *)options {
-    if (_options != options) {
+    if (![_options isEqualToArray:options]) {
         _options = [options copy];
 
         [self repopulateStackview];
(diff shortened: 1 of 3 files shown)


Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigation.m
```diff
@@ -108,7 +108,9 @@ - (void)updateBarAppearance {
         if (UIAccessibilityIsReduceMotionEnabled()) {
             animationDuration = 0.0;
         }
-
+        
+        [self setNeedsLayout];
+        [self layoutIfNeeded];
```
fedetrim: There might be pending changes in the layout that we need to do before executing the animation, otherwise these pending changes will be animated as well.

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigation.m
```diff
@@ -257,6 +259,7 @@ - (void)repopulateStackview {
     // Remove all current arranged views:
     for (UIView *subView in self.stackView.arrangedSubviews) {
         [self.stackView removeArrangedSubview:subView];
+        [subView removeFromSuperview];
```
fedetrim: There are two steps needed to remove a subview from a `UIStackView`, this one was missing

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigation.m
```diff
@@ -274,7 +277,7 @@ - (void)repopulateStackview {
 }
 
 - (void)setOptions:(NSArray<BPKHorizontalNavigationOption *> *)options {
-    if (_options != options) {
+    if (![_options isEqualToArray:options]) {
```
fedetrim: If the contents of the array it's the same, don't update the layout

imchristian:
@fedetrim are you sure that you're rendering the views in the correct lifecycle method? because if you look at the navBar is making a weird layout after the `viewWillAppear` too and is not related to the `HorizontalNavigation` component 🤔

fedetrim:
> @fedetrim are you sure that you're rendering the views in the correct lifecycle method? because if you look at the navBar is making a weird layout after the `viewWillAppear` too and is not related to the `HorizontalNavigation` component 🤔

@imchristian I thought the same thing as you, so I tried moving the rendering of the views to `viewDidAppear()` and the glitch still happened (later, of course). Then I made these changes in Backpack, and the issue was fixed for both the horizontal navigation and the titles of the screen.

The animation block (`UIView.animate...`) had a `view.layoutIfNeeded()` inside, so the titles had changes pending, which were being picked up by the animation block in the horizontal nav.

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigation.m
imchristian: I think that here you only need the `layoutIfNeeded` that's the synchronous call that tells the system you want a layout and redraw of a view and its subviews, and also you want it done immediately without waiting for the update cycle.
Have you try it this solution without the `setNeedsLayout` here?

fedetrim: You are absolutely right. Just deleted `setNeedsLayout` and it works fine.

PR #530, opened 2019-12-03
[No JIRA] Expose large title text alignment

k0nserv:

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


k0nserv:
> @k0nserv what happens with the self.titleView.titleLabel.text here? do we also need to update the alignment for that one right?
It's [centered](https://github.com/Skyscanner/backpack-ios/blob/master/Backpack/NavigationBar/Classes/BPKNavigationBarTitleView.m#L59) so no we don't

PR #535, opened 2019-12-05
[NO-JIRA] Remove old snapshots

georgegillams:
For some reason there were a load of unused 3x snapshots checked in alongside the 2x ones


k0nserv:
Can we prevent this from happening again in the future?

georgegillams:
> Can we prevent this from happening again in the future?

Yes, but I think it would be quite a lot of effort and probably not worth it?

k0nserv:
Wouldn't a CI check or something suffice? Danger for example

PR #548, opened 2019-12-16
[BPK-3415] Calendar month heading RTL alignment

georgegillams:

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/Backpack/Calendar/Classes/BPKCalendarStickyHeader.m
+++ b/Backpack/Calendar/Classes/BPKCalendarStickyHeader.m
@@ -37,7 +37,6 @@
     self = [super initWithFrame:frame];
 
     if (self) {
-        self.titleLabel.textAlignment = NSTextAlignmentLeft;
         [self.weekdayView removeFromSuperview];
         [self.bottomBorder removeFromSuperview];
     }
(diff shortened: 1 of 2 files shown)


Backpack/Calendar/Classes/BPKCalendarStickyHeader.m
```diff
@@ -37,7 +37,7 @@ - (instancetype)initWithFrame:(CGRect)frame {
     self = [super initWithFrame:frame];
 
     if (self) {
-        self.titleLabel.textAlignment = NSTextAlignmentLeft;
+        self.titleLabel.textAlignment = NSTextAlignmentNatural;
```
georgegillams: Before this was causing the text to appear on the left even for RTL languages
Now it appears on the left for LTR and the right for RTL

georgegillams:
> Nice catch 🙏 `NSTextAlignmentNatural` is the default for `UILabel` and `BPKLabel` so it can just be omitted

That's what I thought, but then it goes to the center

georgegillams:
> Nice catch 🙏 `NSTextAlignmentNatural` is the default for `UILabel` and `BPKLabel` so it can just be omitted

That's what I thought, but then it goes to the center

Simulator Screen Shot - iPhone 8 - 2019-12-16 at 10 12 09

k0nserv:
It's inherited behaviour from the library as it turns out. Can we use the property syntax consistently namely in `configureAppearance`?

PR #563, opened 2019-12-24
[BPK-3348] Screenshot automation with Fastlane

georgegillams:
Try it!
`rake take_screenshots`


Example/Backpack/View Controllers/ToastViewController.swift
```diff
@@ -44,14 +44,14 @@ class ToastViewController: UIViewController {
         toast.mode = .indeterminate
         toast.labelText = "This is the title"
         toast.detailsLabelText = "Details label text"
-        toast.hide(true, afterDelay: 1.2)
+        toast.hide(true, afterDelay: 5)
```
georgegillams: I had to change these to ensure they are visible when the screenshot is captured

Rakefile
```diff
@@ -103,6 +105,16 @@ task :lint do
   sh "bundle exec pod lib lint --allow-warnings"
 end
 
+task :take_screenshots do
+  # Remove existing screenshots
+  FileUtils.rm_rf(Dir.glob('screenshots/*'))
```
k0nserv: Do we have to glob here? `rm_rf` already removes recursively so

```ruby
FileUtils.rm_rf('screenshots')
```

should be enough

georgegillams: But we don't want to delete `screenshots`, just everything inside it. Hence the `glob` and the `*`

k0nserv: Isn't everything in `screenshots` generated though? Seems cleaner to just remove the whole thing

georgegillams: Actually it still works. I guess `fastlane` recreates the directory so 🤷‍♀️

PR #565, opened 2019-12-27
[No JIRA] Reinstate `UIFont` apis in `BPKFont`

k0nserv:
These APIs were removed because our type scale required us to do all
text rendering via attributed string. This is no longer the case so we
can reinstate these APIs.

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/Example/Tests/BPKFontTest.m
+++ b/Example/Tests/BPKFontTest.m
@@ -102,26 +102,36 @@ NS_ASSUME_NONNULL_BEGIN
 }
 
 - (void)testFontWithStyle {
-    /*
-    XCTAssertEqualObjects([BPKFont fontWithStyle:BPKFontStyleTextXs], [BPKFont textXs]);
-    XCTAssertEqualObjects([BPKFont fontWithStyle:BPKFontStyleTextXsEmphasized], [BPKFont textXsEmphasized]);
+    BPKFontStyle styles[] = {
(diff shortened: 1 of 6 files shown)


Example/Tests/BPKFontTest.m
```diff
@@ -102,26 +102,36 @@ - (void)testAttributedStringWithCustomFontFaces {
...
+        BPKFontStyleTextXsEmphasized,
+        BPKFontStyleTextXxl,
+        BPKFontStyleTextXxlEmphasized,
+        BPKFontStyleTextXxlHeavy,
+        BPKFontStyleTextXxxl,
+        BPKFontStyleTextXxxlEmphasized,
+        BPKFontStyleTextXxxlHeavy
+    };
+
+    for (NSUInteger i = 0; i < sizeof(styles) / sizeof(styles[0]); i++) {
```
georgegillams: Could we just use `for (BPKFontStyle style in styles) { ... }`?

k0nserv: Not with C arrays no

georgegillams: Why is it a C array tho?

k0nserv: Because otherwise we'd have to box it in `NSNumber` to put them in an `NSArray`

PR #567, opened 2019-12-30
[NO-JIRA] Kill xcode 9 builds

georgegillams:
This is what travis provides in the xcode 11.2 image:
https://docs.travis-ci.com/user/reference/osx/#xcode-1121

If we want to build using different SDKs, I think we need to keep using separate xcode images as before.

TL;DR: Don't merge unless @k0nserv is happy to.


k0nserv:
Different SDKs is not too important.

PR #582, opened 2020-01-21
[NO-JIRA] Fix bottom sheet shadow for apps codebase

georgegillams:


Backpack/BottomSheet/Classes/BackpackFloatingPanelController.swift
```diff
@@ -45,14 +45,12 @@ final class BackpackFloatingPanelController: FloatingPanelController {
     }
 
     override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
-        #if swift(>=4.2)
```
georgegillams: For some reason this was not being satisfied at runtime 🤷‍♀️
We don't actually need this check anymore though, so 🤪

PR #601, opened 2020-03-06
[No JIRA] Improve UI tests

k0nserv:
The UI tests were running with animation at the default speeds which
caused them to take longer to run. Because we don't care about
animations or their durations in UI tests we can disable animations for
UI tests and increase the speed of those animations that aren't disabled
by this.

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


k0nserv:
`cmd + u` locally goes from 1:51 to 1:19 with this change

PR #604, opened 2020-03-09
[BPKR-121] [AMT-4360] Add API for Bottom Sheet with intrinsic size

fedetrim:
This PR adds an API that allows instantiating and presenting a Bottom Sheet without scrollable content. The height of the bottom sheet will be automatically calculated based on its intrinsic height.

Screen-Recording-2020-03-09-at-15 57 48

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/Backpack/BottomSheet/Classes/BottomSheet.swift
+++ b/Backpack/BottomSheet/Classes/BottomSheet.swift
@@ -107,6 +107,8 @@ public final class BottomSheet: NSObject {
     
     /// Instantiates a `BottomSheet` with a non-scrollable content. Height of the bottom sheet will be
     /// calculated based on the content.
+    /// If the content height might not fit the screen, then `init(contentViewController: _, scrollViewToTrack: _)`
+    /// should be used instead.
     /// - Parameter contentViewController: Content of the bottom sheet.
     public init(contentViewController: UIViewController) {
         super.init()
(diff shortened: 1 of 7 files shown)


georgegillams:
Sorry for the delay getting back to you. What should happen if the content is too large for the screen? Should the scroll behaviour be preserved if the content is too large?

fedetrim:
@georgegillams 
>Sorry for the delay getting back to you. What should happen if the content is too large for the screen? Should the scroll behaviour be preserved if the content is too large?

Thanks for your question. IMHO if there's a possibility that the content won't fit, then it's responsibility the user (of the API) to use the other init instead (the one that takes a scroll view as a param).

Otherwise, if Backpack takes that responsibility, we would have detect this case programmatically and then embed the provided content VC into a scroll view so that it can be scrolled and fallback to the other behaviour. I think it is too much and it might hurt maintainability in the future.

What do you think?

georgegillams:
> @georgegillams
> 
> > Sorry for the delay getting back to you. What should happen if the content is too large for the screen? Should the scroll behaviour be preserved if the content is too large?
> 
> Thanks for your question. IMHO if there's a possibility that the content won't fit, then it's responsibility of the user (of the API) to use the other init instead (the one that takes a scroll view as a param).
> 
> Otherwise, if Backpack takes that responsibility, we would have detect this case programmatically and then embed the provided content VC into a scroll view so that it can be scrolled and fallback to the other behaviour. I think it is too much and it might hurt maintainability in the future.
> 
> What do you think?

Seems reasonable to pass this responsibility to the consumer 👍
Maybe we should add a note to the new initialiser stating that it should only we used if the content will definitely fit on screen?

fedetrim:
@georgegillams makes sense, just pushed the improvement

PR #605, opened 2020-03-10
[BPK-3747] Customize horizontal navigation items

k0nserv:
Here's the suggested evolution to the current `BPKHorizontalNavigation`
API. It ends up being a breaking change unfortunately but I think in
practice no one relies on the aspects that will break, mainly the change
in type of `options`.

## Badge Example

To implement the new badge type we'd create a new option type
`BPKHorizontalNavigationOptionBadge` which adopts the
`BPKHorizontalNavigationOptionType` protocol and returns a new
`UIControl` subclass, `BPKHorizontalNavigationItemWithBadge` from `makeItem`.

+ [ ] `UNRELEASED.md`
+ [ ] `README.md`
+ [ ] Tests

### DIFF
--- a/Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigation.m
+++ b/Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigation.m
@@ -24,7 +24,8 @@
 #import <Backpack/Spacing.h>
 
 #import "BPKHorizontalNavigationItem.h"
-#import "BPKHorizontalNavigationOption.h"
+#import "BPKHorizontalNavigationItemDefault.h"
+
 
 NS_ASSUME_NONNULL_BEGIN
 
@@ -40,7 +41,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 @implementation BPKHorizontalNavigation
 
-- (instancetype)initWithOptions:(NSArray<BPKHorizontalNavigationOption *> *)options
+- (instancetype)initWithOptions:(NSArray<id<BPKHorizontalNavigationOptionType>> *)options
                        selected:(NSInteger)selectedItemIndex {
     BPKAssertMainThread();
     self = [super initWithFrame:CGRectZero];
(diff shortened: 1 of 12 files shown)


Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigationItem.m
```diff
@@ -37,6 +37,19 @@ @interface BPKHorizontalNavigationItem ()
 
 @implementation BPKHorizontalNavigationItem
 
+- (instancetype)initWithOption:(BPKHorizontalNavigationOption *)option {
```
georgegillams: Should this be `id<BPKHorizontalNavigationOptionType>`?

k0nserv: Strictly speaking yes but this whole class doens't work if it's not `BPKHorizontalNavigationOption` so it should probably throw

k0nserv:
The size thing is proving to be complex too

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigationOptionType.h
```diff
@@ -0,0 +1,51 @@
...
+@property(nonatomic, readonly) NSInteger tag;
+
+/**
+ * A `UIControl` sublclass to use for rendering this item
+ * in a `BPKHorizontalNavigation`. In addition to being a `UIControl`
+ * subclass it should implement the initializer
+ *
+ *      - (instancetype)initWithOption:(id<BPKHorizontalNavigationOptionType>)option;
+ */
+@property(class, nonatomic, readonly) Class itemClass;
```
imchristian: Can we make this a `UIControl` instead of `Class` to ensure that at least is inheriting from `UIControl`?
In addition, we can move the initializer `- (instancetype)initWithOption:(BPKHorizontalNavigationOption *)option` to a protocol to force this property to be a `UIControl<someProtocolWithTheInitializer>`

k0nserv: That doens't work because it's not returning an instance it's just returning a `Class` which we later intialize. I don't believe there's any way to constrain the type of a `Class`

k0nserv: But maybe it's just cleaner to return an instance rather than a class here anyway

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigation.m
```diff
@@ -149,34 +150,36 @@ - (void)updateSelection:(UIButton *)sender {
...
     return newItem;
 }
 
-- (void)forEachNavigationItem:(void (^)(BPKHorizontalNavigationItem *))callback {
+- (void)forEachNavigationItem:(void (^)(UIControl<BPKHorizontalNavigationItem> *))callback {
     for (UIView *subView in self.stackView.arrangedSubviews) {
-        NSAssert([subView isKindOfClass:[BPKHorizontalNavigationItem class]],
-                 @"HorizontalNav subview is not of type BPKHorizontalNavigationItem as expected.");
-        if (![subView isKindOfClass:[BPKHorizontalNavigationItem class]]) {
+        NSAssert([subView isKindOfClass:[UIControl class]] &&
```
imchristian: Do we still need to check for this inheritance and protocol? can we have some scenario in this current implementation where you can have one of the subviews not being a `UIControl` or conforming the `BPKHorizontalNavigationItem` protocol?

k0nserv: It's because we are fishing them out of the stack views arranged subviews here. In theory there should only be things of this type in the stack view but it doesn't hurt to have the check just in case?

imchristian: the thing is that AFAIK the `stackView` is not exposed in the `BPKHorizontalNavigation` so I think that there's not a direct way to jump that rule and do what you've mentioned right?

k0nserv: Future us can make a mistake

Backpack/HorizontalNavigation/README.md
```diff
@@ -33,6 +33,10 @@ let horizontalNavigation = Backpack.HorizontalNavigation(options: options, selec
 horizontalNavigation.showsSelectedBar = false
 ```
 
+## Custom Horizontal Navigation Segments
+
+The Horizontal Navigation can use custom segments. To achieve this implement a custom type that conforms to `BPKHorizontalNavigationOptionType` and return your own view(a subclass of `UIControl` that implements the protocol `BPKHorizontalNavigationItem`) for `makeItem`. Make sure the height of your view is similar to that of those rendered when `BPKHorizontalNavigationOption` is used directly for all the sizes that the component supports.
```
georgegillams:
```suggestion
Horizontal Navigation can use custom segments. To achieve this, implement a custom type that conforms to `BPKHorizontalNavigationOptionType` and return your own view (a subclass of `UIControl` that implements the protocol `BPKHorizontalNavigationItem`) for `makeItem`. Ensure the height of your view is similar to that of those rendered when `BPKHorizontalNavigationOption` is used directly for all the sizes that the component supports.
```

PR #609, opened 2020-03-13
Horizontal Nav with dot notification

imchristian:
Now the horizontal navigation options allow specifying whether we need to show a notification dot or not for a given section.

DotNotificationScreenshot

+ [ ] `UNRELEASED.md`
+ [ ] `README.md`
+ [ ] Tests


Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigationItemDefault.m
```diff
@@ -32,18 +32,27 @@ @interface BPKHorizontalNavigationItemDefault ()
 @property(readonly) UIColor *contentColor;
 @property(nonatomic) CGFloat horizontalSpacing;
 @property(nonatomic) CGFloat verticalSpacing;
+@property(nonatomic) BOOL showNotificationDot;
+@property(nonatomic, strong) UIView *notificationDot;
```
k0nserv: This should be `readonly` right?

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigationItemDefault.m
```diff
@@ -186,5 +197,30 @@ - (CGFloat)horizontalSpacing {
...
+
+        [self addSubview:_notificationDot];
+        [NSLayoutConstraint activateConstraints:@[
+            [_notificationDot.trailingAnchor constraintEqualToAnchor:self.trailingAnchor constant:-self.horizontalSpacing],
+            [_notificationDot.topAnchor constraintEqualToAnchor:self.topAnchor],
+            [_notificationDot.heightAnchor constraintEqualToConstant:BPKSpacingMd],
+            [_notificationDot.widthAnchor constraintEqualToConstant:BPKSpacingMd]
+        ]];
+
+        _notificationDot.layer.cornerRadius = BPKSpacingMd/2;
```
k0nserv: Nitpic: Spacing around `/`

Example/Backpack/View Controllers/HorizontalNavViewController.swift
```diff
@@ -27,15 +27,25 @@ class HorizontalNavViewController: UIViewController, BPKTappableLinkLabelDelegat
...
 
         if showIcons {
             horizontalNav.options = [
-                BPKHorizontalNavigationOption(name: "Flights", iconName: .flight, tag: 0),
-                BPKHorizontalNavigationOption(name: "Hotels", iconName: .hotels, tag: 1),
-                BPKHorizontalNavigationOption(name: "Car hire", iconName: .cars, tag: 2)
+                BPKHorizontalNavigationOption(name: "Flights",
+                                              iconName: .flight,
+                                              tag: 0,
+                                              showNotificationDot: showNotificationDot),
```
k0nserv: Shouldn't only one of the tabs show the dot? Ideally the first one that's not selected i.e. `Hotels`?

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigationItemDefault.m
```diff
@@ -176,6 +186,7 @@ - (void)setup {
     self.titleLabel.lineBreakMode = NSLineBreakByTruncatingTail;
     [self updateStyle];
     [self updateIconStyle];
+    [self showNotificationDot:self.showNotificationDot];
```
georgegillams: this name seems confusing as it could be used to hide the dot. Maybe 
```suggestion
    [self updateNotificationDot:self.showNotificationDot];
```

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigationOption.m
```diff
@@ -77,6 +78,23 @@ - (instancetype)initWithName:(NSString *)name iconName:(BPKIconName)iconName tag
     return self;
 }
 
+- (instancetype)initWithName:(NSString *)name
```
georgegillams: Could `initWithName:(NSString *)name iconName:(BPKIconName)iconName tag:(NSInteger)tag` just call this method now?

georgegillams:
Is it possible to update the dot at runtime without changing the options?
If it is, we should probably have a story for that in the example app

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigationItemDefault.m
imchristian: It's in the .m file so 🤷‍♂️

k0nserv: It's still helpful though? It expresses the intent better for anyone reading the code even if it has no impact on consumers?

imchristian: Yeah that's fair I'll do the change 👍

imchristian:
@georgegillams as we've discussed yesterday through slack there's not actually a way to update an option without replacing it entirely and since the feature to update the dot icon without replacing the whole option is not a requirement now we decided to not implement it.

imchristian:
@k0nserv I've changed the constraint logic, I didn't try to constraint it to the `UILabel` before because I thought that I'll be unreliable but it's actually working pretty well so good catch 👍 
The snapshot tests are still showing changes even though I deleted the `sizeToFit` but as far as I can see there isn't any visual change between both screenshots.

k0nserv:
There is, look at `testNarrow_light-mode@2x.png` for example

imchristian:
@k0nserv you're right, I'll investigate more because I don't know what could cause that

imchristian:
Ok, so here are the learnings:
The first approach was, of course, follow with the `UIEdgeInsetsMake` + `CGRect` usage to positioning the dot subview inside the `UIButton`.
This approach was extremely unreliable and it wasn't always in the best position (see @k0nserv [suggestion](https://github.com/Skyscanner/backpack-ios/pull/609#pullrequestreview-374299107))

After that, I tried with auto layout to positioning the dot subview, it's working perfectly to options without icons since the auto layout is renderer properly. The problems came up when you add a dot notification using also the inner `UILabel` and `UIImageView` in the `UIButton` 🤔
It was adding an extra space (16pt) between the icon and label, why? well I went to the books to find out this black magic problem:

**"Autolayout is implemented through the superview chain, so if a view uses autolayout, then automatically so do all its superviews"**

Basically when we add with autolayout the dot subview all the `UIButton` start to render their own constraints.
Those are not cancellable using `autoresizingMaskIntoConstraint` 🤷‍♂️ and because of that we cannot render the whole `UIButton` with our own constraints.

It's not a problem for the use case that we need now which is the section with title and dot notification with no icons, so the icon-label padding is not happening.
To avoid that we've limited through the init the dot notification to be used only with title and without icons.

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigationItemDefault.h
```diff
@@ -55,7 +55,17 @@ NS_SWIFT_NAME(HorizontalNavigationItemDefault) IB_DESIGNABLE @interface BPKHoriz
...
+- (instancetype)initWithName:(NSString *)name iconName:(BPKIconName)iconName;
+
+/**
+ * Create a `BPKHorizontalNavigationItem` with a set of options and an optionaly selected option.
+ *
+ * @param name NSString the name for the navigation item
+ * @param iconName BPKIconName the icon for the navigation item
+ * @param showDot send YES to show a red notification dot
+ */
+- (instancetype)initWithName:(NSString *)name iconName:(BPKIconName)iconName showNotificationDot:(BOOL)showDot NS_DESIGNATED_INITIALIZER;
```
georgegillams: I think the public interface should not have a method for initialising with an icon and dot if this is unsupported. It might still be good to have this method privately for the others to call?

k0nserv: I think that's supported? What's not supported it usin git with a badge?

k0nserv:
> It's not a problem for the use case that we need now which is the section with title and dot notification with no icons, so the icon-label padding is not happening.
To avoid that we've limited through the init the dot notification to be used only with title and without icons.

Can we not  use manual layout to position the dot to get around this?

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigationItemDefault.h
imchristian: Well, this class is [not declared](https://github.com/Skyscanner/backpack-ios/blob/master/Backpack/HorizontalNavigation/Classes/HorizontalNavigation.h) to be used directly so that's why I didn't think twice this initializer but we can still discuss if it's necessary anyway to do that change

imchristian:
@k0nserv by the manual layout you mean not to specify the positions using `UIEdgeInsetsMake` for the option with icon and title?
If that's what you mean I think it's not the right path to follow since Apple recommends to use `self.imageEdgeInsets` and `self.titleEdgeInsets` to change a native `UIButton` layout and as I said on my comment if we start adding constraints we can't deactivate the ones added by Apple with `translatesAutoresizingMaskIntoConstraints`.

k0nserv:
Well use `imageEdgeInsets` and `titleEdgeInsets` to position the title and icon, then use `layoutSubviews` to layout the dot only

imchristian:
@k0nserv I think that I'm not following you perhaps we should move to a pair session to tackle this down FYI @georgegillams

k0nserv:
Sounds good to me

imchristian:
Screenshots updated with the latest changes FYI @georgegillams @k0nserv 

Simulator Screen Shot - iPhone 8 - 2020-03-16 at 12 05 38
Simulator Screen Shot - iPhone 8 - 2020-03-16 at 12 06 02

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigationItemDefault.m
```diff
@@ -32,19 +32,32 @@ @interface BPKHorizontalNavigationItemDefault ()
 @property(readonly) UIColor *contentColor;
 @property(nonatomic) CGFloat horizontalSpacing;
 @property(nonatomic) CGFloat verticalSpacing;
+@property(nonatomic, readonly) UIView *notificationDot;
 
 @end
 
 @implementation BPKHorizontalNavigationItemDefault
 
+@synthesize notificationDot = _notificationDot;
```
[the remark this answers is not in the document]
k0nserv: I checked locally by removing it and everything is fine

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigationOption.h
```diff
@@ -88,5 +88,30 @@ NS_ASSUME_NONNULL_BEGIN
 */
 - (instancetype)initWithName:(NSString *)name iconName:(BPKIconName)iconName tag:(NSInteger)tag;
 
+/**
+* Intialize an option with a given name, tag, and icon.
+*
+* @param name The name of the option.
+* @param tag The tag of the option.
+* @param showDot send YES to show a red notification dot
+* @return A named option with an icon and a tag.
```
k0nserv: This is outdated

PR #628, opened 2020-03-31
Add appearance option to Horizontal Navigation

ofalvai:
Adding optional alternate style to `BPKHorizontalNavigation`. This will allow using the view over a dark background in light mode, just like the [Android component](https://backpack.github.io/components/horizontal-nav/?platform=android).

Added `BPKHorizontalNavigationAppearance` with values `Normal` and `Alternate`. This follows the [Android component's naming convention](https://github.com/Skyscanner/backpack-android/blob/master/Backpack/src/main/java/net/skyscanner/backpack/horisontalnav/BpkHorizontalNav.kt#L36). `Alternate` style causes the text and icon colour to be `BPKColor.skyGrayTint07` (instead of `textPrimaryColor` which resolves to a dark grey in light mode) allowing good contrast on dark backgrounds.

|Light|Dark|
|---|---|
|image|image|

+ [X] Check this if you have read and followed the [contributing guidelines](https://github.com/Skyscanner/backpack-ios/blob/master/CONTRIBUTING.md)

+ [X] `UNRELEASED.md`
+ [X] `README.md`
+ [X] Tests


k0nserv:
Looks like a good start. I think we need to make changes to `BPKHorizontalNavigationOptionType` too. So the horizontal nav can inform each of the options what style to use

ofalvai:
@k0nserv I think `BPKHorizontalNavigationItem` would be a better place for this data.

k0nserv:
Yes indeed, you are correct of course

ofalvai:
Added the appearance option handling to the various navigation item classes. I decided to do the actual color mapping in the `contentColor` method to apply the alternate colors not only to the text but to the icons as well. For the alternate color I used `BPKColor.skyGrayTint07` for consistency with the Android component, but let me know if there's a better value.

If you think the changes are good, then I'll continue with adding tests and documentation.

k0nserv:
Yeah this looks good but it needs to consider Dark Mode too. You'll run into that when you do the snapshot tests anyway

ofalvai:
Not sure what you mean by considering dark mode. My assumption is that the background will be dark in dark mode as well, so the alternate text colour should be the same as in light mode.

Added some tests, snapshot tests and a new section to the example app. What do you think?

ofalvai:
Also I had to update the snapshots. I think something is broken the way light and dark mode colors are handled in all the test cases.

Backpack/HorizontalNavigation/README.md
```diff
@@ -39,3 +41,4 @@ Horizontal Navigation can use custom segments. To achieve this, implement a cust
 
 ### Appearance attributes
 `(UIColor)contentColor`
+`(BPKHorizontalNavigationAppearance)appearance`
```
georgegillams, line 44: This isn't a `UI_APPEARANCE_SELECTOR` so should be removed from this section

PR #634, opened 2020-04-02
[BPK-3795] Add tests and docs

georgegillams:


imchristian:
@georgegillams can we improve the test and example showing some titles and `BPKIcons` maybe?

PR #636, opened 2020-04-03
[NO-JIRA] Update snapshots

georgegillams:


georgegillams:
The calendar change was deliberate
The toast change is just where it's fading out when the snapshot was taken, so no big deal

PR #643, opened 2020-04-08
[BPK-3769] Improve loading for text only buttons

k0nserv:
<details>
<summary>Screenshot</summary>

Simulator Screen Shot - iPhone 8 - 2020-04-08 at 14 23 25

</details>

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


Backpack/Button/Classes/BPKButton.m
```diff
@@ -255,7 +276,7 @@ - (void)layoutSubviews {
     }
     CGFloat buttonTitleIconSpacing = [[self class] buttonTitleIconSpacing];
 
-    if (self.isTextAndIcon) {
+    if (self.isTextAndIcon || (self.isLoading && self.isTextOnly)) {
```
georgegillams, line 279: Would this be cleaner?
```suggestion
    if (self.isTextAndIcon || self.isLoading) {
```

Backpack/Button/Classes/BPKButton.m
```diff
@@ -130,16 +132,26 @@ - (BOOL)hasTitle {
     return self.titleLabel.text.length > 0 && self.currentAttributedTitle != nil;
 }
 
+- (BOOL)hasImage {
```
georgegillams, line 135: I think it would be cleaner to have a property `hasIcon` which returns whether the icon property is set.
In my mind, `hasImage` would always be true when `loading` is true

Backpack/Button/Classes/BPKButton.m
k0nserv: No because `isTextAndIcon` is false when we use the dummy images

Backpack/Button/Classes/BPKButton.m
k0nserv: `hasImage` is independant of `loading`. For example a text and icon button still has an image if it's loading. However I'm open to calling this `hasIcon`

georgegillams: Yeah maybe `hasIcon` would clear things up 👍

Backpack/Button/Classes/BPKButton.m
georgegillams: 👍

PR #644, opened 2020-04-10
[BPK-3812] Enable overriding flare size

georgegillams:

I'm not gonna put this in the readme or changelog because I don't think we want to encourage overriding.

### DIFF
--- a/Backpack/FlareView/Classes/BPKFlarePath.h
+++ b/Backpack/FlareView/Classes/BPKFlarePath.h
@@ -47,8 +47,7 @@ IB_DESIGNABLE @interface BPKFlarePath : NSObject
  */
 + (UIBezierPath *)flareViewPathForSize:(CGSize)size
                              flareSize:(CGSize)flareSize
-                         flarePosition:(BPKFlarePosition)flarePosition NS_SWIFT_NAME(makeFlarePath(size:flareSize:
-                                                                                                   flarePosition:));
+                         flarePosition:(BPKFlarePosition)flarePosition;
 @end
 
 NS_ASSUME_NONNULL_END
--- a/Backpack/FlareView/Classes/BPKFlarePath.h
+++ b/Backpack/FlareView/Classes/BPKFlarePath.h
@@ -42,12 +42,13 @@ IB_DESIGNABLE @interface BPKFlarePath : NSObject
  * Creates and returns a UIBeizerPath of the flare mask for a given size.
  *
  * @param size CGSize of the path required.
- * @param flareSize CGSize of the path required.
+ * @param flareHeight CGFloat height of the flare required.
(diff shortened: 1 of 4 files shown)


Backpack/FlareView/Classes/BPKFlarePath.h
```diff
@@ -23,16 +23,6 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-/**
- * The height at which the BPKFlare should be rendered
- */
-extern CGFloat const BPKFlareHeight;
```
georgegillams, line 29: these are only used in one place now, so no point exposing them

georgegillams: This file is not exposed to consumers so not breaking

Backpack/FlareView/Classes/BPKFlarePath.h
```diff
@@ -52,11 +42,12 @@ IB_DESIGNABLE @interface BPKFlarePath : NSObject
  * Creates and returns a UIBeizerPath of the flare mask for a given size.
  *
  * @param size CGSize of the path required.
+ * @param flareSize CGSize of the path required.
+ * @param flarePosition BPKFlarePosition of the path required.
  */
 + (UIBezierPath *)flareViewPathForSize:(CGSize)size
-                         flarePosition:(BPKFlarePosition)flarePosition
-    NS_SWIFT_NAME(makeFlarePath(size:flarePosition:));
```
georgegillams, line 58: removed the NSSwiftName because this class is not exposed to consumers, so it'll never be called from swift anyway

Backpack/FlareView/Classes/BPKFlareView.h
```diff
@@ -62,5 +62,7 @@ IB_DESIGNABLE @interface BPKFlareView : UIView
  */
 - (instancetype)initWithFrame:(CGRect)frame NS_DESIGNATED_INITIALIZER;
 
+- (CGSize)flareSize;
```
georgegillams, line 65: Want this to be undocumented as we don't want consumers to override it

Backpack/FlareView/Classes/BPKFlarePath.h
k0nserv: Doesn't it make sense to have it anyway? It does no harm and if we eventually expose it we won't forget to add a swift name

Backpack/FlareView/Classes/BPKFlareView.h
k0nserv: Since the aspect ratio is always fixed would it be better to expose a way to set one of `height` or `width` and have the other one be calculated? This would mean it's impossible to render the flare distorted at least

PR #647, opened 2020-04-13
[BPK-3812] add bar chart bar

georgegillams:

+ [ ] `UNRELEASED.md`
+ [ ] `README.md`
+ [x] Tests


Backpack/BarChart/Classes/BarChartBar.swift
```diff
@@ -0,0 +1,181 @@
...
+ */
+
+import UIKit
+
+@objcMembers
+@objc(BPKBarChartBar)
+public final class BPKBarChartBar: UIControl {
+
+    let backgroundView: UIView = UIView()
+    let barView: UIView = UIView()
```
k0nserv, line 26: I was thinking we'd do this with layers instead of dedicated views. It would make for a simpler view hierarchy

Backpack/BarChart/Classes/BarChartBar.swift
```diff
@@ -0,0 +1,181 @@
...
+    let miniFlareView: BPKMiniFlareView = BPKMiniFlareView()
+    let titleLabel: BPKLabel = BPKLabel(fontStyle: .textBaseEmphasized)
```
k0nserv, line 28: Can I suggest we use lazy getters for these in which we do all the setup required i.e.

```swift
    lazy var titleLabel: BPKLabel = {
        let label = BPKLabel(fontStyle: .textBaseEmphasized)
        label.translatesAutoresizingMaskIntoConstraints = false

        return label
    }()
```

Backpack/BarChart/Classes/BarChartBar.swift
```diff
@@ -0,0 +1,181 @@
...
+    let subtitleLabel: BPKLabel = BPKLabel(fontStyle: .textSm)
+    var barViewHeightAnchor: NSLayoutConstraint?
+    var miniFlareViewPositionAnchor: NSLayoutConstraint?
+    let backgroundViewColor = BPKColor.dynamicColor(withLightVariant: BPKColor.skyGrayTint06,
```
k0nserv, line 32: This is only used in a single place so might be worth just inlining directly there? As it stands it's an instance property so it will needlessly add to the size of this class in memory even though the value is static

Backpack/BarChart/Classes/BarChartBar.swift
```diff
@@ -0,0 +1,181 @@
...
+                                                    darkVariant: BPKColor.blackTint03)
+    let noPriceColor = BPKColor.skyGrayTint03
+    let priceColor = BPKColor.primaryColor
+    let selectedColor = BPKColor.dynamicColor(withLightVariant: BPKColor.monteverde, darkVariant: BPKColor.glencoe)
```
k0nserv, line 36: Same with these three

Backpack/BarChart/Classes/BarChartBar.swift
```diff
@@ -0,0 +1,181 @@
...
+        self.setupViews()
+    }
+
+    public required init?(coder: NSCoder) {
+        super.init(coder: coder)
+        self.setupViews()
+    }
+
+    func setupViews() {
+        self.backgroundView.backgroundColor = self.backgroundViewColor
```
k0nserv, line 49: I don't think we should use `self` unless required to disambiguate. That's more swifty imo

Backpack/BarChart/Classes/BarChartBar.swift
```diff
@@ -0,0 +1,181 @@
...
+
+        self.miniFlareViewPositionAnchor = self.miniFlareView.bottomAnchor.constraint(equalTo: self.barView.topAnchor,
+                                                                                      constant: -BPKSpacingSm)
+
+        self.miniFlareViewPositionAnchor!.isActive = true
+
+        self.miniFlareView.isHidden = self.realValue == nil || !self.isSelected
+    }
+
+    public override var isSelected: Bool {
```
k0nserv, line 131: Can we keep public variables at the top of the file? These also need to be documented

Backpack/BarChart/Classes/BarChartBar.swift
```diff
@@ -0,0 +1,181 @@
...
+            return super.isSelected
+        }
+        set {
+            super.isSelected = newValue
+            self.updateBarColor()
+            self.updateminiFlareViewAppearance()
+        }
+    }
+
+    private var privateFillValue: CGFloat?
```
k0nserv, line 142: I think you want `fileprivate`

k0nserv: In general I think we should use `fileprivate` for everything that's not public by default

Backpack/BarChart/Classes/MiniFlareView.swift
```diff
@@ -0,0 +1,68 @@
...
+    public var text: String? {
+        get {
+            return self.label.text
+        }
+        set {
+            self.label.text = newValue
+        }
+    }
+
+    override public func flareHeight() -> CGFloat {
```
k0nserv, line 65: Note: I feel like this should be a readonly property in `BPKFlareView` and the same here. In Objective-C there is no distinction because a readonly property is just a method but I don't think using a method allows us to override it with a readonly property in Swift.

Backpack/BarChart/Classes/BarChartBar.swift
imchristian: In addition to that, part of the code included here should live into the lazy getters for the properties.

Backpack/BarChart/Classes/BarChartBar.swift
imchristian: I agree with that, if we make things private for usages outside of the Backpack module should be enough in most of the cases.

Backpack/BarChart/Classes/MiniFlareView.swift
georgegillams: So in FlareView.h it would be
```
@property(readonly, nonatomic) CGFloat flareHeight;
```

And FlareView.m would be unchanged
```
- (CGFloat)flareHeight {
    return BPKFlareHeight;
}
```

Then in MiniFlareView.swift we'd do
```
    override public var flareHeight: CGFloat {
        get {
            return BPKSpacingMd
        }
        set {
            return
        }
    }
```

Backpack/BarChart/Classes/BarChartBar.swift
georgegillams: That seems like it would involve some horrible calculations in code?

Backpack/BarChart/Classes/MiniFlareView.swift
k0nserv: Yeah except in Swift it just needs to be

```swift
override public var flareHeight: CGFloat {
    BPKSpacingMd
}
```

Because of implicit returns in getters and the ability to omit `get` when implemnting a read-only property

Backpack/BarChart/Classes/BarChartBar.swift
k0nserv: A single percentage calculation?

Backpack/BarChart/Classes/BarChartBar.swift
```diff
@@ -0,0 +1,202 @@
...
+            titleLabel.textColor = BPKBarChartBar.selectedColor
+            subtitleLabel.textColor = BPKBarChartBar.selectedColor
+            return
+        }
+
+        titleLabel.textColor = BPKColor.textPrimaryColor
+        subtitleLabel.textColor = BPKColor.textPrimaryColor
+
+        if fillValue == nil {
+            barView.backgroundColor = BPKBarChartBar.noValueColor
```
k0nserv, line 169: This can be written as `barView.backgroundColor = fillValue.map { BPKBarChartBar.valueColor } ?? BPKBarChartBar.noValueColor`

which is slightly nicer

Backpack/BarChart/Classes/BarChartBar.swift
```diff
@@ -0,0 +1,202 @@
...
+        }
+
+        if fillValue == nil {
+            barViewHeightAnchor = barView.heightAnchor.constraint(equalToConstant: BPKSpacingLg)
+        } else {
+            barViewHeightAnchor = barView.heightAnchor.constraint(equalTo: backgroundView.heightAnchor,
+                                                                  multiplier: CGFloat(exactly: fillValue!)!)
+        }
+
+        barViewHeightAnchor!.isActive = true
```
k0nserv, line 187: I'd probably use question mark here anyway

PR #648, opened 2020-04-14
[BPK-3839] add icon support to chips

georgegillams:

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


Backpack/Chip/Classes/BPKChip.m
```diff
@@ -30,8 +30,9 @@
 @interface BPKChip ()
 @property(nonatomic) BPKLabel *titleLabel;
 @property(nonatomic) CALayer *tintLayer;
-
-@property(nonatomic) UIColor *textColor;
+@property(nonatomic) UIColor *contentColor;
+@property(nullable, nonatomic) UIView *iconView;
```
k0nserv, line 34: Why isn't this a `BPKIconView` here?

Backpack/Chip/Classes/BPKChip.h
```diff
@@ -77,6 +79,11 @@ IB_DESIGNABLE @interface BPKChip : UIControl
  */
 @property(nullable, nonatomic) UIColor *backgroundTint;
 
+/**
+ * An optional icon
+ */
+@property(nullable, nonatomic) BPKIconName iconName;
```
k0nserv, line 85: This should be copy

georgegillams: Fixed

Backpack/Chip/Classes/BPKChip.m
georgegillams: We don't do anything to it that a UIView can't do, so it seems unecessary to constrain it to a `BPKIconView`.

k0nserv: But it is a `BPIconView`. It's not about what we do it, it's about helping the person reading the code understand. Reading `UIView` I started assuming that we might sometimes use a `BPKIconView` and other times some other type which hurts readability

georgegillams: Fixed

PR #649, opened 2020-04-15
[BPK-3746] Improve BPKSnackbar

k0nserv:
The snackbar wasn't quite done according to the design. Notably it was
missing the `text` property which I have added. To make the layout code
simpler I changed the approach to using a `UIStackView` instead of raw
Autolayout code.

As the above ends up being a breaking change I fixed the incorrect name
of the snackbar's delegate which is also breaking.

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [ ] Tests


k0nserv:
Just need to add tests and ensure the docs demonstrate this properly.

Example/Backpack/View Controllers/SnackbarSelectorViewController.swift
```diff
@@ -35,6 +35,9 @@ class SnackBarSelectorViewController: UITableViewController {
         setupDestinationController(destinationController: destinationController, segue: segue)
     }
 
+    // Cyclomatic complexity might be high in this function but it's actually really
```
georgegillams, line 38: This comment seems a bit yagni 🤷‍♂️

k0nserv: I feel like it's a good idea to include a comment when disabling lint rules, that's why I added it

k0nserv:
> I guess it's correct that a screen reader will read out the title and text separately?
Yeah that seems reasonable, they are two elements. Maybe we should give the title an accesisibility trait?

Backpack/Snackbar/Classes/BPKSnackbarButton.h
```diff
@@ -0,0 +1,94 @@
...
++ (instancetype)buttonWithTitle:(NSString *)title;
+
+/**
+ * A button with an icon and visible text.
+ *
+ * @param title The title to use.
+ * @param icon The icon to use.
+ * @return An instance that will display a text + icon button.
+ */
++ (instancetype)buttonWithIcon:(UIImage *)icon title:(NSString *)title;
```
tiagoengel, line 65: Is this part of the design, is not supported on Android/RN

k0nserv: Dunno but this isn't actually new, the previous API also supported it. By encapsulating this in an object we've only restricted the API to prevent creating icon only buttons that aren't visible to VoiceOver.

Backpack/Snackbar/Classes/BPKSnackbar.m
```diff
@@ -148,39 +177,36 @@ - (id)init {
...
+                        button:(BPKSnackbarButton *)button
                       leftIcon:(UIImage *_Nullable)leftIcon
                       duration:(BPKSnackbarDuration)duration
                 viewController:(UIViewController *)viewController
-                      delegate:(id<BPKSnackbarProtocol> _Nullable)delegate {
+                      delegate:(id<BPKSnackbarDelegate> _Nullable)delegate {
     self.titleLabel.text = title;
+    self.textLabel.text = text;
     self.delegate = delegate;
     self.snackbarView.backgroundColor = BPKColor.skyBlueShade02;
```
imchristian, line 189: Can we make this change to match the Android implementation for DM?

k0nserv: I'm sunnure what you mean

imchristian: The snackbar background should be white on DM, right? or we should change Android's implementation to match ours.

k0nserv: Yeah I guess we should figure that out but I'd suggest we call that out of scope for this PR

k0nserv:
I [removed the text + icon button](https://github.com/Skyscanner/backpack-ios/pull/649/commits/1f64d4d7aa32ef802fcedbcde39dc8e140c26bf4) since it's not supported in design and Android. We can always add it back if needed

PR #655, opened 2020-04-21
[BPK-3812] Add BPKBarChartCollectionView

georgegillams:

+ [ ] `UNRELEASED.md`
+ [ ] `README.md`
+ [x] Tests


Backpack/BarChart/Classes/BarChartCollectionView.swift
```diff
@@ -0,0 +1,149 @@
...
+    public func collectionView(
+        _ collectionView: UICollectionView, cellForItemAt indexPath: IndexPath
+    ) -> UICollectionViewCell {
+        guard let cell = dequeueReusableCell(
+            withReuseIdentifier: BPKBarChartCollectionView.cellIdentifier,
+            for: indexPath) as? BPKBarChartCollectionViewCell
+            else {
+                fatalError("No cell registered for reuse with identifier \(BPKBarChartCollectionView.cellIdentifier)")
+        }
+        cell.barChartBar.title = barChartDataSource?.titleForBar(barChartCollectionView: self,
```
k0nserv, line 129: We need to set `cell.isSelected` here too

Backpack/BarChart/Classes/BarChartCollectionViewCell.swift
```diff
@@ -0,0 +1,63 @@
...
+    lazy public var barChartBar: BPKBarChartBar = {
+        let view: BPKBarChartBar = BPKBarChartBar()
+        view.translatesAutoresizingMaskIntoConstraints = false
+        view.isUserInteractionEnabled = false
+        return view
+    }()
+
+    public override var isSelected: Bool {
+        get {
+            return barChartBar.isSelected
```
k0nserv, line 35: The cell already has and manages it's own selection state let's not override that instead we can use `didSet` to update the bar's state

Backpack/BarChart/Classes/BarChartCollectionViewDataSource.swift
```diff
@@ -0,0 +1,38 @@
...
+public protocol BPKBarChartCollectionViewDataSource {
+
+    /// The number of bars to display in the chart
+    func numberOfBarsInChart(barChartCollectionView: BPKBarChartCollectionView) -> Int
+
+    /// The title to show underneath a specific bar
+    func titleForBar(barChartCollectionView: BPKBarChartCollectionView, atIndex: Int) -> String
+
+    /// The subtitle to show under the title of a specific bar
+    func subtitleForBar(barChartCollectionView: BPKBarChartCollectionView, atIndex: Int) -> String
```
k0nserv, line 31: This should be optional right? You don't have to provide a subtitle after all

Backpack/BarChart/Classes/BarChartCollectionViewDataSource.swift
```diff
@@ -0,0 +1,38 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import UIKit
+
+@objc(BPKBarChartCollectionViewDataSource)
+public protocol BPKBarChartCollectionViewDataSource {
+
+    /// The number of bars to display in the chart
```
k0nserv, line 24: These doc comments should be expanded to describe the arguments and return values

Backpack/BarChart/Classes/BarChartCollectionView.swift
```diff
@@ -0,0 +1,149 @@
...
+        setupViews()
+    }
+    
+    required init?(coder: NSCoder) {
+        super.init(coder: coder)
+        setupViews()
+    }
+    
+    func setupViews() {
+        backgroundView = UIView() // <-- This is a bit hacky!
```
k0nserv, line 52: I'll say, this view is never added to the view hierarchy and it's not positioned. I think we need to find a better method than this

georgegillams: Yeah I agree. It's strange that this works. Maybe the green bar should be rendered by the parent component 🤷

k0nserv: I think we should explore constraining it to the collection view some more

k0nserv: I would try [`frameLayoutGuide`](https://developer.apple.com/documentation/uikit/uiscrollview/2865772-framelayoutguide)

Backpack/BarChart/Classes/BarChartCollectionView.swift
georgegillams: Why? When a cell is created it shouldn't be selected by default 🤔

Backpack/BarChart/Classes/BarChartCollectionView.swift
georgegillams: YOU ARE A GENIUS!!!!!

Backpack/BarChart/Classes/BarChartCollectionView.swift
k0nserv: Yeah but what if it is the selected cell? I believe because you aren't setting this here you have issues when scrolling. Keep in my whenever a cell is created you need to ensure that it relfects the underlying data, this includes whether it's selected or not

k0nserv: i.e. we want something like `cell.isSelected = selectedIndexPath == indexPath`

Backpack/BarChart/Classes/BarChartCollectionViewDataSource.swift
georgegillams: For now, I assumed you would for simplicity. I guess we can relax it in the future if there's a use case?

k0nserv: Alright

Backpack/BarChart/Classes/BarChartCollectionView.swift
imchristian: @k0nserv you're too smart to be here, get out. 😆

Backpack/BarChart/Classes/BarChartCollectionView.swift
```diff
@@ -0,0 +1,149 @@
...
+        view.backgroundColor = BPKColor.monteverde
+        view.isHidden = true
+        view.translatesAutoresizingMaskIntoConstraints = false
+        return view
+    }()
+    
+    fileprivate func updateSelectedMarkerPosition() {
+        if selectedMarkerBottomConstraint != nil {
+            selectedMarkerBottomConstraint?.isActive = false
+            selectedMarkerBottomConstraint = nil
```
k0nserv, line 89: Do we need to set this value to `nil` can't we just deactivate it and hide the view instead? This might be able save us from using a force unwrapped optional for the value?

Backpack/BarChart/Classes/BarChartCollectionView.swift
```diff
@@ -0,0 +1,149 @@
...
+        }
+        cell.barChartBar.title = barChartDataSource?.titleForBar(barChartCollectionView: self,
+                                                                 atIndex: indexPath.item)
+        cell.barChartBar.subtitle = barChartDataSource?.subtitleForBar(barChartCollectionView: self,
+                                                                       atIndex: indexPath.item)
+        cell.barChartBar.fillValue = barChartDataSource?.fillValueForBar(barChartCollectionView: self,
+                                                                         atIndex: indexPath.item)
+        cell.barChartBar.valueDescription = barChartDataSource?.valueDescriptionForBar(barChartCollectionView: self,
+                                                                                       atIndex: indexPath.item)
+        cell.isSelected = selectedIndexPath == indexPath
```
k0nserv, line 136: @georgegillams Did this help with the scrolling issue?

Backpack/BarChart/Classes/BarChartCollectionView.swift
```diff
@@ -0,0 +1,149 @@
...
+        return cell
+    }
+}
+
+extension BPKBarChartCollectionView: UICollectionViewDelegate {
+    public func collectionView(_ collectionView: UICollectionView, didSelectItemAt indexPath: IndexPath) {
+        selectedIndexPath = indexPath
+        
+        if barChartDelegate != nil {
```
k0nserv, line 145: Afaik you dont need to check this, just doing 

`barChartDelegate?.didSelect(barChart: self, index: indexPath.item)` will do nothing if `barChartDelegate` is `nil`

Backpack/BarChart/Classes/BarChartCollectionViewCell.swift
```diff
@@ -0,0 +1,60 @@
...
+    public override init(frame: CGRect) {
+        super.init(frame: frame)
+        setupViews()
+    }
+
+    fileprivate func setupViews() {
+        contentView.addSubview(barChartBar)
+
+        NSLayoutConstraint.activate([
+        barChartBar.topAnchor.constraint(equalTo: contentView.topAnchor),
```
k0nserv, line 53: Spacing

Backpack/BarChart/Classes/BarChartCollectionView.swift
georgegillams: nope

PR #660, opened 2020-04-24
[BPK-3812] Animate marker bar

georgegillams:

+ [ ] `UNRELEASED.md`
+ [ ] `README.md`
+ [ ] Tests

### DIFF
--- a/Backpack/BarChart/Classes/BarChartBar.swift
+++ b/Backpack/BarChart/Classes/BarChartBar.swift
@@ -72,7 +72,7 @@ public final class BPKBarChartBar: UIControl {
 
     /// The absolute position of the top of the bar view
     public var barTopPosition: CGFloat {
-        return self.barView.frame.origin.y
+        return barView.frame.origin.y
     }
 
     fileprivate var barViewHeightAnchor: NSLayoutConstraint?
--- a/Backpack/BarChart/Classes/BarChartCollectionView.swift
+++ b/Backpack/BarChart/Classes/BarChartCollectionView.swift
@@ -83,14 +83,18 @@ public final class BPKBarChartCollectionView: UICollectionView {
     lazy fileprivate var selectedMarker: UIView = {
         let view = UIView()
         view.backgroundColor = BPKColor.monteverde
+        view.layer.opacity = 0
         view.isHidden = true
         view.translatesAutoresizingMaskIntoConstraints = false
         return view
     }()
     
     fileprivate func updateSelectedMarkerPosition() {
+        var selectedBarTopPosition = selectedMarkerBottomConstraint.constant
+        var selectedBarOpacity: Float = 1.0
         if selectedIndexPath == nil {
             selectedMarker.isHidden = true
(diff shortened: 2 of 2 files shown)


Backpack/BarChart/Classes/BarChartBar.swift
```diff
@@ -72,7 +72,7 @@ public final class BPKBarChartBar: UIControl {
 
     /// The absolute position of the top of the bar view
     public var barTopPosition: CGFloat {
-        return self.barView.frame.origin.y
+        return barView.frame.origin.y
```
georgegillams, line 75: removed `self.` as it's more Swifty to not use it unless necessary

Backpack/BarChart/Classes/BarChartCollectionView.swift
```diff
@@ -99,8 +102,18 @@ public final class BPKBarChartCollectionView: UICollectionView {
...
+
+            if selectedCell.barChartBar.fillValue == nil {
+                selectedBarOpacity = 0
+            } else {
+                selectedBarTopPosition = selectedCell.barChartBar.barTopPosition
+                selectedBarOpacity = 1
+            }
+        }
+        UIView.animate(withDuration: BPKDuration.animationDurationSm) {
+            self.selectedMarkerBottomConstraint.constant = selectedBarTopPosition
```
georgegillams, line 114: `self.` is necessary here as the scope cannot be inferred inside an animation code block

imchristian:
@georgegillams I'll suggest an improvement for the animation. 
If the bar is not visible maybe because we tapped a bar without price or is the first time, the current behaviour is making appear the bar from the top and then goes to the given position, we should avoid that and do something more subtle like fade-in in the final position maybe for that initial state?

georgegillams:
> @georgegillams I'll suggest an improvement for the animation.
> If the bar is not visible maybe because we tapped a bar without price or is the first time, the current behaviour is making appear the bar from the top and then goes to the given position, we should avoid that and do something more subtle like fade-in in the final position maybe for that initial state?

That's a great suggestion. Have fixed and updated the gif

Backpack/BarChart/Classes/BarChartCollectionView.swift
```diff
@@ -83,14 +83,18 @@ public final class BPKBarChartCollectionView: UICollectionView {
...
         view.backgroundColor = BPKColor.monteverde
+        view.layer.opacity = 0
         view.isHidden = true
         view.translatesAutoresizingMaskIntoConstraints = false
         return view
     }()
     
     fileprivate func updateSelectedMarkerPosition() {
+        var selectedBarTopPosition: CGFloat = selectedMarkerBottomConstraint.constant
+        var selectedBarOpacity: Float = 1.0
```
imchristian, line 94: the type is not inferred in this case? 🤔

georgegillams: It infers type Double instead of float.
That said the CGFloat type above is inferred, so I've removed that

PR #688, opened 2020-06-08
[No Jira] Add dev environment details

shaundon:
I couldn't remember how to find it so I added some docs. Feel free to suggest edits if not quite right.


k0nserv:
It's already mentioned a few lines below though

shaundon:
omg how did I not see that, sorry! I'll close it. I'm an idiot

PR #696, opened 2020-07-01
[No Jira] Remove some warnings

shaundon:
All the warnings are annoying me so I removed some of the low-hanging fruit. There are still more, ideally I'd like to get down to zero, but this is a start.


Example/Backpack/Images.xcassets/AppIcon.appiconset/Contents.json
```diff
@@ -1,62 +1,116 @@
...
-      "size" : "60x60",
-      "idiom" : "iphone",
       "filename" : "icon60@3x.png",
-      "scale" : "3x"
+      "idiom" : "iphone",
+      "scale" : "3x",
+      "size" : "60x60"
+    },
+    {
+      "filename" : "icon1024-1.png",
```
georgegillams, line 52: This doesn't look right

georgegillams: Looks like the icons have been duplicated somehow 🤔

shaundon: This is just because I exported them from Sketch with the same filename. I don't think the filename matters though, they show up in Xcode just fine. This is just an internal detail.

georgegillams: But why are there more icons?

shaundon: One of the warnings was about us not providing iPadOS icons when our app supports iPad, so I added the missing ones to make the warnings go away.

georgegillams: Do they need to be separate files to the ones that already exist? I think you can just assign the icon for both iOS and iPadOS

shaundon: I don't think you can, it requires different sizes for each platform.

shaundon: Also bear in mind that this is a non-shipped app so it doesn't really matter

georgegillams: But a lot of the icons you've added are the same size 🤔

shaundon: Ah yeah sorry - it's confusing. Hopefully this screenshot will clear it up:

Screen Shot 2020-07-02 at 11 11 35

_Some of_ the icons are indeed the same size, for example 'iPhone notification 2x' and 'iPad Notification 2x' are both 40x40px. 

As far as I'm aware there's no way to alias them, you need to provide separate icons for each. I assume this is because of the way the App Store intelligently downloads only the bundle needed for each app - so you don't have to download the iPhone icons when you install on iPad, and so on.

The filenames behind the scenes are similar because of the way I exported from Sketch, but this isn't a problem.

georgegillams: Do we even support any iPad devices that use 1x?
Maybe this is just apple being dumb in the face 🤷 
Would be nice to have consistent file naming regardless

shaundon: > Do we even support any iPad devices that use 1x?

No, but Xcode requires it so it's a moot point.

It's good that you're being thorough here but this thread is total yak shaving - I can't think of something more minor than this. I'm gonna merge it and you can change the filenames later if you wanna do it 😛

georgegillams: Yeah I wanna
https://github.com/Skyscanner/backpack-ios/pull/698

PR #713, opened 2020-07-21
[BPK-4073] Update icons

georgegillams:

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [ ] Tests


UNRELEASED.md
```diff
@@ -1,8 +1,14 @@
 # Unreleased
+
 > Place your changes below this line.
 
+**Breaking:**
```
georgegillams, line 5: For most use cases, the changes I've made don't require consumers to change their code.

I'm making this breaking, however, for the following reasons:
 - Not all icons are available in both large and small, whereas before they were.
 - If someone is specifying the icon string manually from the font mapping instead of using one of our auto-generated defined strings then it will break. The solution to this is to use one of the predefined values!

Example/Backpack/View Controllers/IconsViewController.swift
```diff
@@ -77,8 +83,20 @@ extension IconsViewController {
                 fatalError("No cell registered for reuse with identifier \(IconsViewController.cellIdentifier)")
         }
 
-        let icon = IconsViewController.icons[indexPath.row]
-        cell.icon = icon
+        let iconSet = IconsViewController.icons[indexPath.section]
+
+        // Whether we set `size` or `icon` first, we will potentially be trying to use an icon that doesn't exist.
+        // By setting the icon to `accessibility`, then setting the size, and then setting the iconName, we should
+        // never be trying to use an icon at an unavailable size.
```
georgegillams, line 90: This is a bit of a hacky workaround. Fine for an example app, but maybe we should just reduce the severity of the error if the icon name and size combination doesn't exist, otherwise consumers might have to start doing similar things in their code

Backpack/Icon/Classes/BPKIcon.m
```diff
@@ -125,6 +123,19 @@ + (UIImage *)iconNamed:(NSString *)name color:(UIColor *)color size:(BPKIconSize
...
 }
 
++(NSString *)codepointForIcon:(NSString *)name size:(BPKIconSize)size {
+    NSString *lookupKey = name;
+    if(size == BPKIconSizeSmall) {
+        lookupKey = [NSString stringWithFormat:@"%@%@", name, @"-sm"];
+    }
+
+    if (self.iconMapping) {
+        NSAssert([self.iconMapping objectForKey:lookupKey], @"Unknown icon `%@` does not correspond to a known icon. Please check that the icon exists at the size you are trying to use https://backpack.github.io/components/icon?platform=design", lookupKey);
```
k0nserv, line 133: This might result in confusing errors because the `-sm` suffix is an implementation detail that the caller might not be aware of. It might be better to change the message to metnion the basename and the size requested

Backpack/Icon/Classes/BPKIcon.m
```diff
@@ -125,6 +123,19 @@ + (UIImage *)iconNamed:(NSString *)name color:(UIColor *)color size:(BPKIconSize
...
+    }
+
+    return [self.iconMapping objectForKey:lookupKey];
```
k0nserv, line 136: Unrelated: I noticed we aren't handling the `nil` case explicitly for the above assert. Can we verify that it does something sensible(like returning an empty image)

Example/Backpack/View Controllers/IconsViewController.swift
k0nserv: We could provide a method that can veirfy if a given icon is available at a given size?

Backpack/Icon/Classes/BPKIcon.m
georgegillams: Yeah this is the same as the logic we had before - will double check though 👍

georgegillams:
> I'm not stoked about the fact that we are going from an API that was very difficult to misuse to one where a fair amount of combinations are invalid.

Yeah me too. Was discussing this with Cristian just now. I think this is probably the fastest approach for now, with the smallest blast radius, but in the future we should move to an API like `icon:BPKiconNameBeerLg`.

I guess the question is whether we want a two-part rollout using this to get us there faster, or a single, massive breaking change

k0nserv:
> Yeah me too. Was discussing this with Cristian just now. I think this is probably the fastest approach for now, with the smallest blast radius, but in the future we should move to an API like icon:BPKiconNameBeerLg.

> I guess the question is whether we want a two-part rollout using this to get us there faster, or a single, massive breaking change

Doing it in two steps would end up being two breaking changes so if we think using names that include the size is ther better option making a single breaking change feels like the better idea.

Example/Backpack/View Controllers/IconsViewController.swift
georgegillams: That won't actually help though, because by definition you still have to change one property before the other

k0nserv: Hmm yeah I guess this speaks even more to the idea that we should encode the size in the enums

PR #737, opened 2020-09-03
[BPK-4167] Remove non UIKit aligned tokens

georgegillams:

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


UNRELEASED.md
```diff
@@ -2,6 +2,20 @@
 
 > Place your changes below this line.
 
+**BREAKING:**
+
+- Backpack/BorderSize:
+
+  - Has been removed
```
shaundon, line 9: Can we make this a bit more consumer friendly? It's not cool to just say it's been removed, I think we can do more.

Something like:

> The border size tokens have been removed. To migrate, simply replace them like for like with the border width tokens that were added in $version.

I know you've done the work to replace them in the apps codebase anyway, but I think it's still good practice to make this clear. There could be people in the future who try and use a border size token from memory, can't find it, get confused and come here to see where it went.

shaundon (changes requested):
Code looks good, just a comment about the documentation. Once fixed feel free to merge

shaundon (approved):

PR #739, opened 2020-09-07
[BPK-4180] Allow custom UITabBar for BPKTabBarController

imchristian:
https://gojira.skyscanner.net/browse/BPK-4180

We need to keep using [this subclass of UITabBar](https://github.skyscannertools.net/apps-tribe/skyscanner-app/blob/79afba3651f2c672e29fd1a3c180dafc2dc30c42/ios/Modules/Skyscanner.UICore/Headers/SkyUICore/SKYTabBar.h#L4) for the Native Navigation Roadmap, in order to do so, we need to add the capability to inject a custom UITabBar into our BPKTabBarController component.

+ [ ] `UNRELEASED.md`
+ [ ] `README.md`
+ [ ] Tests

### DIFF
--- a/Backpack/TabBarController/Classes/BPKTabBarController.h
+++ b/Backpack/TabBarController/Classes/BPKTabBarController.h
@@ -26,6 +26,13 @@ NS_ASSUME_NONNULL_BEGIN
  */
 IB_DESIGNABLE @interface BPKTabBarController : UITabBarController
 
+/**
+*  Creates a new BPKTabBarController instance with a custom UITabBar
+*
+*  @param tabBar Custom UITabBar instance to be used as the default tabBar
+*/
+- (instancetype)initWithTabBar:(UITabBar *)tabBar;
+
 @end
 
 NS_ASSUME_NONNULL_END
--- a/Backpack/TabBarController/Classes/BPKTabBarController.m
+++ b/Backpack/TabBarController/Classes/BPKTabBarController.m
@@ -29,8 +29,29 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+@interface BPKTabBarController ()
+@property (nonatomic, strong) UITabBar *customTabBar;
+@end
+
 @implementation BPKTabBarController
+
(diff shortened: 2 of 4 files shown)


Backpack/TabBarController/Classes/BPKTabBarController.h
```diff
@@ -26,6 +26,13 @@ NS_ASSUME_NONNULL_BEGIN
  */
 IB_DESIGNABLE @interface BPKTabBarController : UITabBarController
 
+/**
+*  Creates a new BPKTabBarController instance with a custom UITabBar
+*
+*  @param tabBar Custom UITabBar subclass to be used as the default tabBar
```
georgegillams, line 32: Technically this is an instance of UITabBar, right?

```suggestion
*  @param tabBar Custom UITabBar instance to be used as the default tabBar
```

Backpack/TabBarController/Classes/BPKTabBarController.m
```diff
@@ -29,8 +29,29 @@
...
 NS_ASSUME_NONNULL_BEGIN
 
+@interface BPKTabBarController ()
+@property (nonatomic, strong) UITabBar *customTabBar;
+@end
+
 @implementation BPKTabBarController
 
+- (instancetype)initWithTabBar:(UITabBar *)tabBar {
+    if (self = [super init]) {
```
georgegillams, line 39: How does this tab bar actually get used? Don't we need to call something like `[super initWithTabBar: tabBar]`?

imchristian: I'm not sure if I understood the question correctly but basically in order to set a new tabBar the only thing that we need to do is [override the seter](https://stackoverflow.com/a/47560280).
I created a new initialiser just to receive the custom UITabBar instance through there instead of setting a property after init.

georgegillams: Ah that makes perfect sense 👍

georgegillams (approved):

PR #752, opened 2020-09-28
[BPK-3692] Remove NSDate methods in BPKCalendar

k0nserv:

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/Backpack/Calendar/Classes/BPKCalendar.m
+++ b/Backpack/Calendar/Classes/BPKCalendar.m
@@ -39,7 +39,37 @@
 #import "BPKCalendarStickyHeader.h"
 #import "BPKCalendarYearPill.h"
 
+
 NS_ASSUME_NONNULL_BEGIN
+
+// This is kept around only to support the React Native calendar implementation
+@protocol BPKCalendarDelegateExtendedForRN <BPKCalendarDelegate>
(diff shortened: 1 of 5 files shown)


k0nserv:
The idea here is to keep the `DONT_USE` methods for the RN calendar implementation. 

I'm also considering changing some of the methods in the delegate to make it better in Swift since we are makign a breaking change anyway

Backpack/Calendar/Classes/BPKCalendar.m
```diff
@@ -39,8 +39,38 @@
 #import "BPKCalendarStickyHeader.h"
 #import "BPKCalendarYearPill.h"
 
+
 NS_ASSUME_NONNULL_BEGIN
 
+// This is kept around only to support the React Native calendar implementation
```
georgegillams, line 45: If we're removing the deprecated method, why not remove this for RN too and we can adopt the breaking change in `backpack-react-native`?

k0nserv: Actually the current RN implementation actually relies on this API with no other option to migrate to imo

georgegillams (approved):
LGTM

PR #757, opened 2020-10-06
[BPK-3963] Upgrade Swift 5

imchristian:

+ [ ] `UNRELEASED.md`
+ [ ] `README.md`
+ [ ] Tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,9 +15,9 @@ cache:
 matrix:
   include:
     - osx_image: xcode12
-      env: BUILD_SDK=iphonesimulator14.0 DESTINATION="OS=14.0,platform=iOS Simulator,name=iPhone 8" FULL_TESTS="true" MAIN_STAGE="true"
+      env: SWIFT=5.0 BUILD_SDK=iphonesimulator14.0 DESTINATION="OS=14.0,platform=iOS Simulator,name=iPhone 8" FULL_TESTS="true" MAIN_STAGE="true"
     - osx_image: xcode12
-      env: BUILD_SDK=iphonesimulator14.0 DESTINATION="OS=13.5,platform=iOS Simulator,name=iPhone 8" FULL_TESTS="false" MAIN_STAGE="false"
+      env: SWIFT=4.2 BUILD_SDK=iphonesimulator14.0 DESTINATION="OS=13.5,platform=iOS Simulator,name=iPhone 8" FULL_TESTS="false" MAIN_STAGE="false"
 before_install:
   - sudo sntp -sS time.apple.com
   - brew install clang-format
--- a/Backpack.podspec
+++ b/Backpack.podspec
@@ -43,5 +43,5 @@ Pod::Spec.new do |s|
   s.dependency 'MBProgressHUD', '~> 0.9.1'
   s.frameworks = 'UIKit', 'Foundation', 'CoreText'
   s.requires_arc = true
-  s.swift_version = '4.0'
+  s.swift_versions = ['5.0', '4.2', '4.0']
 end
--- a/Backpack/BarChart/Classes/BarChartCollectionView.swift
+++ b/Backpack/BarChart/Classes/BarChartCollectionView.swift
@@ -59,9 +59,15 @@ internal final class BPKBarChartCollectionView: UICollectionView {
         register(BPKBarChartCollectionViewCell.self,
                  forCellWithReuseIdentifier: BPKBarChartCollectionView.cellIdentifier)
 
+        #if swift(>=4.2)
+        register(BPKBarChartCollectionViewHeader.self,
+                 forSupplementaryViewOfKind: UICollectionView.elementKindSectionHeader,
+                 withReuseIdentifier: BPKBarChartCollectionView.headerIdentifier)
+        #else
         register(BPKBarChartCollectionViewHeader.self,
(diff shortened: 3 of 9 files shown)


Example/Podfile
```diff
@@ -29,6 +29,15 @@ post_install do |installer|
     config.build_settings.delete('CODE_SIGNING_REQUIRED')
   end
 
+  installer.pods_project.targets.each do |target|
+    target.build_configurations.each do |config|
+        if target.name == "Backpack" || target.name == "FloatingPanel"
+            config.build_settings['SWIFT_VERSION'] = '5.0'
+        end
+        config.build_settings['IPHONEOS_DEPLOYMENT_TARGET'] = '12.0'
```
imchristian, line 37: This eliminates the current warnings 

`The iOS Simulator deployment target 'IPHONEOS_DEPLOYMENT_TARGET' is set to 8.0, but the range of supported deployment target versions is 9.0 to 14.0.99.`

Example/Podfile
```diff
@@ -29,6 +29,15 @@ post_install do |installer|
     config.build_settings.delete('CODE_SIGNING_REQUIRED')
   end
 
+  installer.pods_project.targets.each do |target|
+    target.build_configurations.each do |config|
+        if target.name == "Backpack" || target.name == "FloatingPanel"
+            config.build_settings['SWIFT_VERSION'] = '5.0'
```
imchristian, line 35: Without this the pods reset to the previous swift versions every time we do `pod install` and it turns into a compile failure. [Some info about this](https://jeroenscode.com/migrating-to-swift-5/).

k0nserv: There's a way to indicate swift version via the `.podspec`. I feel like we should be using that and check if FloatingPanel hasn't added a way to do this too?

Example/Podfile
k0nserv: I think the warnings are good to have? We can chase the projects that need to update this

k0nserv (commented):
Can we have a CI job that runs on 4.2? This is what I have done for SkyFloatingLabelTextField, see https://github.com/Skyscanner/SkyFloatingLabelTextField/blob/master/.travis.yml

Backpack/BottomSheet/Classes/BackpackFloatingPanelController.swift
```diff
@@ -69,7 +69,7 @@ private extension BPKFloatingPanelController {
         #if swift(>=4.2)
         addChild(bottomSection)
         #else
-        addChildViewController(bottomSection)
+        addChild(bottomSection)
```
georgegillams, line 72: Seems like this should already be ignored by Swift 5?
Maybe we can just remove this `if...else` if we are dropping support for `<= 4.2`?

Backpack.podspec
```diff
@@ -43,5 +43,5 @@ Pod::Spec.new do |s|
   s.dependency 'MBProgressHUD', '~> 0.9.1'
   s.frameworks = 'UIKit', 'Foundation', 'CoreText'
   s.requires_arc = true
-  s.swift_version = '4.0'
+  s.swift_version = '5.0'
```
k0nserv, line 46: We can and should provide an array here `swift_versions = ['5.0', '4.2', '4.0']`

.travis.yml
```diff
@@ -15,9 +15,9 @@ cache:
 matrix:
   include:
     - osx_image: xcode12
-      env: BUILD_SDK=iphonesimulator14.0 DESTINATION="OS=14.0,platform=iOS Simulator,name=iPhone 8" FULL_TESTS="true" MAIN_STAGE="true"
+      env: SWIFT=5.0 BUILD_SDK=iphonesimulator14.0 DESTINATION="OS=14.0,platform=iOS Simulator,name=iPhone 8" FULL_TESTS="true" MAIN_STAGE="true"
     - osx_image: xcode12
-      env: BUILD_SDK=iphonesimulator14.0 DESTINATION="OS=13.5,platform=iOS Simulator,name=iPhone 8" FULL_TESTS="false" MAIN_STAGE="false"
+      env: SWIFT=4.2 BUILD_SDK=iphonesimulator14.0 DESTINATION="OS=13.5,platform=iOS Simulator,name=iPhone 8" FULL_TESTS="false" MAIN_STAGE="false"
```
k0nserv, line 20: This isn't enough, if you look at https://github.com/Skyscanner/SkyFloatingLabelTextField/blob/master/Rakefile you'll see I am also passing down the swift verison

Backpack.podspec
imchristian: ohh I see, thanks for info 👍

.travis.yml
imchristian: I guess that's the missing part to pass CI then

Backpack.podspec
k0nserv: I think this should be **versions** with an `s` per the [docs](https://guides.cocoapods.org/syntax/podspec.html#swift_versions)

imchristian: My change is `s.swift_versions` [here](https://github.com/Skyscanner/backpack-ios/pull/757/files#diff-ec850618833e8b72d1203858fa50fde5R46)

k0nserv: Ah, I didn't notice that the diff was outdated. My bad

k0nserv (approved):

PR #759, opened 2020-10-08
[BPK-4188] Maintain calendar selection when selection type is changed

georgegillams:

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/Backpack/Calendar/Classes/BPKCalendar.m
+++ b/Backpack/Calendar/Classes/BPKCalendar.m
@@ -91,6 +91,7 @@ NS_ASSUME_NONNULL_BEGIN
 @property(nonatomic, strong, nonnull) BPKCalendarAppearance *appearance;
 @property(nonatomic, strong, nonnull) UIView *bottomBorder;
 @property(nonatomic, strong, nonnull) NSCalendar *gregorian;
+@property(nonatomic, strong, nullable) NSArray<BPKSimpleDate *> *initiallySelectedDates;
 
 @property BOOL sameDayRange;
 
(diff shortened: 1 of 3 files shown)


Backpack/Calendar/Classes/BPKCalendar.m
```diff
@@ -91,6 +91,7 @@ @interface BPKCalendar () <FSCalendarDelegate, FSCalendarDelegateAppearance, FSC
 @property(nonatomic, strong, nonnull) BPKCalendarAppearance *appearance;
 @property(nonatomic, strong, nonnull) UIView *bottomBorder;
 @property(nonatomic, strong, nonnull) NSCalendar *gregorian;
+@property(nonatomic, strong, nullable) NSArray<BPKSimpleDate *> *initiallySelectedDates;
```
k0nserv, line 94: Do we need this? Can't we just use `[super selectedDates]` and retain a single source of truth?

georgegillams: No, because if 2 `selectedDates` are set before the selection type is changed from single to range, then only one of them will be retained

k0nserv: If we get the dates before the `self.calendarView.allowsMultipleSelection` that shouldn't be the case though?

georgegillams: Exactly - that's what we're doing here. Is there a better place to take a copy of the selected items?

k0nserv: I'm saying we just use a local copy inside `setSelectionType` instead of adding a new property. It's only ever used in that method and if we take the copy before the line that changes the value of `allowsMultipleSelection` we can avoid the property altogether

georgegillams: But we need to persist the value between multiple calls of `setSelectionType`

RN will call `setSelectedDates` before `setSelectionType: BPKSelectionTypeMultiple`

georgegillams:
Closing as this is redundant now

PR #767, opened 2020-10-14
[BPK-4188] Fix initial date select bug

k0nserv:
When rendering a calendar using the range select mode with pre-selected
dates any selected cells in the viewport would render their shape layer
incorrectly due to calculating negative values for `selectionRect`. This
in turn happened because `bounds` was set to
`self.selectionLayer.bounds` before its frame had been set the first
time.

+ [ ] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


shaundon (approved):
Looks good! I didn't run it locally but I can see from your screenshots that it works. CI failed, but it looks like it just timed out, so I've restarted it.

PR #772, opened 2020-10-19
BPK 3625 migrate ci to github actions

k0nserv:

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


.github/workflows/ci.yml
```diff
@@ -0,0 +1,136 @@
...
+      - name: Build and Test
+        env:
+          FULL_TEST: TRUE
+          BUILD_SDK: iphonesimulator14.0
+          DESTINATION: "OS=14.0,platform=iOS Simulator,name=iPhone 8"
+        run: bundle exec rake ci
+        working-directory: Example
+
+      - name: Deploy
+        uses: peaceiris/actions-gh-pages@v3
```
olliecurtis, line 113: Just something to keep in mind that has come from internally - when using 3rd party actions it is worth checking and reviewing what its doing and nothing out of the usual :)

olliecurtis (approved):

.github/workflows/ci.yml
k0nserv: Yeah good shout, I have reviewed it. There doesn't seem to be an official action for github pages from GitHub unfortuantely

.github/workflows/ci.yml
```diff
@@ -0,0 +1,136 @@
...
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+
+# For details on the macOS enviornment see https://github.com/actions/virtual-environments/blob/main/images/macos/macos-10.15-Readme.md
+#
+# We try to align the versions of tools with what is already available on the image to reduce the time it takes to run our CI.
+# At the time of writing these are:
```
shaundon, line 12: May be more informative to say the current date

```suggestion
# At the time of writing (October 2020) these are:
```

.nvmrc
```diff
@@ -1 +1 @@
-lts/erbium
+12.18.4
```
shaundon, line 1: How come this changed?

k0nserv: To ensure we use the same version as the macOS image

.github/workflows/ci.yml
k0nserv: I like  it

.github/workflows/ci.yml
```diff
@@ -0,0 +1,136 @@
...
+      - name: Run NPM build
+        run: npm run build
+
+      - name: Run danger
+        run: bundle exec danger
+        if: ${{ matrix.is_main }}
+
+      - name: Check pristine
+        run: ./scripts/check-pristine-state package-lock.json
+
```
georgegillams, line 90: extraneous linebreak

.nvmrc
shaundon: There's the potential to cause confusion here, I could see somebody changing it back in future to match the other repos. Could you add a comment to say it's to match the one on CI?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,136 @@
...
+      - name: Run Swiftlint
+        run: Pods/SwiftLint/swiftlint lint --strict --config .swiftlint.yml
+        working-directory: Example
+
+      - name: Build and verify docs
+        if: ${{ steps.extract_git_tag.outputs.GIT_TAG == '' }}
+        run: ./scripts/build-docs-ci && rm -rf docs
+
+      - name: Build docs
+        if: ${{ steps.extract_git_tag.outputs.GIT_TAG != '' }}
```
georgegillams, line 101: So we only test docs if there's no tag? Should we document why we do that?

georgegillams (approved):
This is great. Should we delete the Travis file in the same PR?

Couple of other minor comments from me

georgegillams (changes requested):
Sorry - meant to RC

.github/workflows/ci.yml
k0nserv: This replicates what we did on Travis namely we do one of:

1. Build the docs and then remove them(to verify it works)
2. Build the docs, but don't remove them(later deployed to backpack.github.io/ios)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,136 @@
...
+    if: github.ref == 'refs/heads/master'
+    runs-on: ubuntu-20.04
+
+    steps:
+      - name: Trigger docs Travis CI
+        env:
+          TRAVIS_API_KEY: ${{ secrets.TRAVIS_API_KEY }}
+        run: |
+          curl -s -X POST -H 'Content-Type: application/json' -H 'Accept: application/json' -H 'Travis-API-Version: 3' -H 'Authorization: token $TRAVIS_API_KEY' -d '{"request": {"branch":"master"}}' https://api.travis-ci.org/repo/Skyscanner%2Fbackpack-docs/requests
+
```
georgegillams, line 133: Should we do this after everything else so that it doesn't happen if something is broken?

.nvmrc
k0nserv: I tried but `.nvmrc` does not support comments

k0nserv: A danger rule maybe?

.github/workflows/ci.yml
k0nserv: I considered this but I think the only thing the docs site needs is the screemshots which aren't tested anyway 🤔

georgegillams: That is very true

.nvmrc
shaundon: Ah, yeah I didn't realise we couldn't put comments in there! Danger rule would be good, could you make an enghealth card for us to do it sometime?

k0nserv: Can do

shaundon (approved):

PR #777, opened 2020-10-22
[NO-JIRA] Always call super layoutSubviews

georgegillams:


shaundon:
Looks good but what's the purpose of this? i.e. why's it necessary?

georgegillams:
> Looks good but what's the purpose of this? i.e. why's it necessary?

When you override a method you should always call the super method otherwise you're replacing the implementation instead of building on top of it.

These ones must have just slipped through review.

shaundon:
Ok cool, thanks for explaining

shaundon (approved):

PR #778, opened 2020-10-23
[BPK-3625] Escape Curl string differently to fix a bug with deploys

shaundon:


.github/workflows/ci.yml
```diff
@@ -124,7 +124,7 @@ jobs:
         env:
           TRAVIS_API_KEY: ${{ secrets.TRAVIS_API_KEY }}
         run: |
-          curl -v -s -X POST -H 'Content-Type: application/json' -H 'Accept: application/json' -H 'Travis-API-Version: 3' -H 'Authorization: token $TRAVIS_API_KEY' -d '{"request": {"branch":"master"}}' https://api.travis-ci.org/repo/Skyscanner%2Fbackpack-docs/requests
+          curl -s -X POST -H "Content-Type: application/json" -H "Accept: application/json" -H "Travis-API-Version: 3" -H "Authorization: token $TRAVIS_API_KEY" -d '{"request": {"branch":"master"}}' https://api.travis-ci.org/repo/Skyscanner%2Fbackpack-docs/requests
```
georgegillams, line 127: Can we also add in `-f` so that the exit code will be non-zero if the call fails. By default curl always exits 0, so if we break this in the future we might not realise!

```suggestion
          curl -f -s -X POST -H "Content-Type: application/json" -H "Accept: application/json" -H "Travis-API-Version: 3" -H "Authorization: token $TRAVIS_API_KEY" -d '{"request": {"branch":"master"}}' https://api.travis-ci.org/repo/Skyscanner%2Fbackpack-docs/requests
```

PR #799, opened 2020-11-10
[BPK-4111] Add pill rating configuration

georgegillams:

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

Simulator Screen Shot - iPhone 8 - 2020-11-10 at 16 54 10

Simulator Screen Shot - iPhone 8 - 2020-11-10 at 16 54 12


shaundon (approved):
LGTM, I ran it locally and read the code, but you might want to wait for Hugo in case he has any more refined critiques. Up to you tbh

Backpack/Rating/Classes/BPKRatingTextWrapper.m
```diff
@@ -100,48 +99,72 @@ - (void)setSize:(BPKRatingSize)size {
...
-    }
-}
-
 - (void)setLayout:(BPKRatingLayout)layout {
     BPKAssertMainThread();
     if (_layout != layout) {
         _layout = layout;
 
+        [self updateFontSizes];
         [self updateConstraints];
```
k0nserv, line 108: This shouldn't be called directly, instead we should call `setNeedsUpdateConstraints`

georgegillams: I don't think I ever actually meant to override/call the UIKit method - I think I just chose a name that clashed.
Have renamed so that we can avoid invoking it ourselves.

k0nserv: I actually think this is a good situation to be using Apple's method though 🤔

georgegillams: Oh haha - I'll use that then 👍

k0nserv (approved):
Noice

PR #810, opened 2020-11-23
Add customisation capability to BPKBottomSheet

paulomendes:
## Add Persistent and Modal Presentation Style

This PR has the objective to introduce a new style of presentation to BottomSheet. Now we are capable to present it modally or persistently.

# Persistent Bottom Sheet

It can be initialized by using

```Swift
let bottomSheet = BPKBottomSheet(contentViewController: viewController, 
 scrollViewToTrack: viewController.tableView, 
 bottomSectionViewController: nil, 
 presentationStyle: .persistent)
```

To use the `.persistent` presentation style we need to ensure to use the method `addPanel` instead of `present`.

This will allow the following behavior. 

ezgif-6-7f75e33ed95b

# What is New

- New parameter in the initialization called `presentationStyle` which is an enum
```Swift
public enum PresentationStyle {
        case modal
        case persistent
    }
```
- New layout definition: `ModalBottomSheetLayout` and `PersistentBottomSheetLayout`
- New `addPanel(toParent parent: UIViewController)` method, that will add a BottomSheet and make it persistent and allow you to interact with the parent view controller
- New `removePanel(animated: Bool)` in case you added a BottomSheet persistently but needs to hide it for some loading interaction or something else.
- New `public func move(to position: BPKFloatingPanelPosition)` that will allow you to programmatically change the position of the BottomSheet.

+ [X] Check this if you have read and followed the [contributing guidelines](https://github.com/Skyscanner/backpack-ios/blob/master/CONTRIBUTING.md)

+ [ ] `UNRELEASED.md`
+ [ ] `README.md`
+ [ ] Tests

### DIFF
--- a/Backpack/BottomSheet/Classes/BottomSheet.swift
+++ b/Backpack/BottomSheet/Classes/BottomSheet.swift
@@ -22,13 +22,11 @@ import FloatingPanel
 @objcMembers
 @objc(BPKBottomSheet)
 public final class BPKBottomSheet: NSObject {
-
-    private enum Constants {
-        static let bottomSheetHeightInHalfPosition: CGFloat = 386.0
-        static let backdropAlpha: CGFloat = 0.3
-        static let grabberHandleWidth: CGFloat = 60.0
+    public enum PresentationStyle {
+        case modal
+        case persistent
     }
-
+    
     /// View controller that will be presented when calling
     /// `present(in: _, animated: _, completion: _)`.
     /// It can also be presented using UIKit's native presentation API.
@@ -58,8 +56,11 @@ public final class BPKBottomSheet: NSObject {
         }
     }
 
+    private let presentationStyle: PresentationStyle
+
     private lazy var floatingPanelController: BPKFloatingPanelController = {
-        let panel = BPKFloatingPanelController(delegate: self)
+        var panel = BPKFloatingPanelController(delegate: self)
+
         panel.surfaceView.backgroundColor = BPKColor.backgroundTertiaryColor
         panel.surfaceView.cornerRadius = BPKCornerRadiusLg
         panel.surfaceView.grabberTopPadding = BPKSpacingMd
@@ -67,14 +68,19 @@ public final class BPKBottomSheet: NSObject {
         panel.surfaceView.grabberHandleWidth = Constants.grabberHandleWidth
         panel.surfaceView.grabberHandle.barColor = BPKColor.skyGrayTint06
 
-        panel.isRemovalInteractionEnabled = true
+        switch presentationStyle {
+        case .modal:
+            panel.isRemovalInteractionEnabled = true
+        case .persistent:
+            panel.isRemovalInteractionEnabled = false
+            panel.backdropView.backgroundColor = .clear
+        }
 
         // We do this to hold a strong reference to `BPKBottomSheet` and force it
         // to exist as long as `floatingPanelController` exists.
         // Reference will be cleaned up by `floatingPanelController` when
         // it's dismissed, to avoid a reference cycle.
         panel.bottomSheet = self
-
         return panel
     }()
 
@@ -93,13 +99,18 @@ public final class BPKBottomSheet: NSObject {
     ///     that should be accessible at all times. A top shadow is automatically added so that
     ///     it integrates better with the content of the bottom sheet.
     ///     Note: Safe Area should be taken into account in the bottom section's inner constraints.
+    ///   - presentationStyle: .modal if you nedd a modal interaction with the BottomSheet. .persistent if
+    ///     you need a persistent BottomSheet and being able to interact with what is behind the BottomSheet
+    
     public init(contentViewController: UIViewController,
                 scrollViewToTrack: UIScrollView,
-                bottomSectionViewController: UIViewController? = nil) {
+                bottomSectionViewController: UIViewController? = nil,
+                presentationStyle: PresentationStyle = .modal) {
+        self.presentationStyle = presentationStyle
         super.init()
 
         self.scrollView = scrollViewToTrack
-
+        
         floatingPanelController.contentViewController = contentViewController
         floatingPanelController.track(scrollView: scrollViewToTrack)
         floatingPanelController.bottomSectionViewController = bottomSectionViewController
@@ -111,6 +122,7 @@ public final class BPKBottomSheet: NSObject {
     /// should be used instead.
     /// - Parameter contentViewController: Content of the bottom sheet.
     public init(contentViewController: UIViewController) {
+        self.presentationStyle = .modal
         super.init()
         floatingPanelController.contentViewController = contentViewController
     }
@@ -124,6 +136,7 @@ public final class BPKBottomSheet: NSObject {
     ///   - completion: Completion closure called after presentation animation.
     @objc(presentInViewController:animated:completion:)
     public func present(in viewController: UIViewController, animated: Bool, completion: (() -> Void)? = nil) {
+        assert(presentationStyle == .modal, "present(in:animated:completion:) not compatible with non modal presentation style")
         viewController.present(viewControllerToPresent, animated: animated, completion: completion)
     }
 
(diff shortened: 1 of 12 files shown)


fedetrim:
Technically this looks good to me. What I'm worried about is that we might be giving too much flexibility to the API, providing a way to have several different styles/behaviours of the bottom sheet within the Skyscanner app. 
Back when we introduced this, the idea was to try to understand the needs of the bottom sheet within our app and only allow to use those. If any new style/behaviour would come up in the future, we would add it. 

This PR however takes a different approach, making this component more flexible, in order to allow consumers to implement their own needs as opposed to the component providing pre-defined styles and behaviours. In that case, why not use the 3rd party `FloatingPanel` directly instead of hiding it behind Backpack's `BottomSheet`?

I can approve this from a technical point of view but I think @k0nserv and @georgegillams should make the final call

k0nserv:
@fedetrim makes a good point here. We want an API that encapsulates the variations we support but still hides details of the underlying library. Extreme customisation is not a goal as much as ease of use and encouraging the right patterns is

Backpack/BottomSheet/Classes/PersistentBottomSheetLayout.swift
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    public var supportedPositions: Set<FloatingPanelPosition> {
+        return [.half, .full, .tip]
+    }
+
+    public func insetFor(position: FloatingPanelPosition) -> CGFloat? {
+        switch position {
+        case .half:
+            return 380
```
k0nserv, line 33: How does this work on e.g. iPhone Se first gen? It's only 586pts tall. Maybe this logic needs to be device aware. Similarily how does this all work on iPad?

Backpack/BottomSheet/Classes/BottomSheet.swift
```diff
@@ -148,50 +162,70 @@ public final class BPKBottomSheet: NSObject {
         bottomSheet.present(in: floatingPanelController, animated: animated, completion: completion)
     }
 
+    /// This method should be used if you inted to show the BPKBottomSheet and keep tracking the backdrop view
+    /// user interactions. If you need your BPKBottomSheet have a modal behaviour you should use present method
+    /// should be used instead.
+    /// - Parameter parent: Parent view where the BPKBottomSheet will be placed.
+    public func addPanel(toParent parent: UIViewController) {
+        assert(presentationStyle == .persistent, "present(in:animated:completion:) not compatible with non modal presentation style")
```
k0nserv, line 170: I'm not stoked about these asserts, maybe we can utilise generic and optional conformance to have the type system solve this for us intead of using runtime errors with asserts? 

i.e. something like:

```swift
protocol BPKBottomSheetPresentationStyle {}
protocol BPKBottomSheetPresentationStyleModal: BPKBottomSheetPresentationStyle {}
protocol BPKBottomSheetPresentationStylePersistent: BPKBottomSheetPresentationStyle {}

 public final class BPKBottomSheet<Style: BPKBottomSheetPresentationStyle> : NSObject { 
....
}

extension BPKBottomSheet where Style: BPKBottomSheetPresentationStylePersistent {
     public func addPanel(toParent parent: UIViewController) { /* ... */ }
}

public struct _BPKBottomSheetPersistentStyle: BPKBottomSheetPresentationStyle, BPKBottomSheetPresentationStylePersistent {}

typealias BPKBottomSheetPersistent = BPKBottomSheet<_BPKBottomSheetPersistentStyle>
```

k0nserv (commented):
There are doc comments missing for a few public symbols, can we add them?

Backpack/BottomSheet/Classes/PersistentBottomSheetLayout.swift
paulomendes: My reasoning for this fixed height was based on what was previously done for the `Modal` presentation style. We had this constant `bottomSheetHeightInHalfPosition` set to `386.0`

Maybe we could let either FloatingPanel decide which will be the "half" height position or create a feature request where BottomSheet is aware of his content and set an appropriate half position. Let me know which option you consider is the best approach.

k0nserv: I guess if you have just ported over existing behaviour I can't very well hold it over you anyway 👍🏼 I'll see if we can fix this later

fedetrim: Shouldn't this be `Constants.bottomSheetHeightInHalfPosition` then? Which is 386 and not 380

paulomendes:
> There are doc comments missing for a few public symbols, can we add them?

The public methods that you refer to, are those in the `ModalBottomSheetLayout` and `PersistentBottomSheetLayout` ?

Backpack/BottomSheet/Classes/PersistentBottomSheetLayout.swift
paulomendes: Yes. I think I did this just for a particular adjustment on the content in the feature that I was working on. I'm going to be consistent here.

k0nserv:
Also the enum, in general we strive to have all public symbols documented with reference docs

Backpack/BottomSheet/Classes/BottomSheet.swift
paulomendes: I think if we follow this approach we would cause an API breaking change. When I opt-in for the `assert` approach, I took into consideration API breaking changes. 

If this is a hard request, I'll be happy to apply it. I don't have any other ideas right now about implement it in a different fashion where I would avoid asserting types. Your idea seems to work here and I would follow a similar path as you suggested.

k0nserv: > I think if we follow this approach we would cause an API breaking change. When I opt-in for the assert approach, I took into consideration API breaking changes.

Yeah that's a good instinct, but often it's okay to make a breaking change if it improves the API greatly(imo). Here we need to balance the worse feedback loop with asserts vs a the improved feedback loop when using an assert instead. 

I think we probably shouldn't do this and live with the asserts. One thing to note though is that in a production build the asserts will not do anything and we will still take the action. Should we ensure that in production builds we just do nothing?

paulomendes: We might get this behavior using `assertionFailure`-> 8a52851727adef325214ff87cb0bcb8a8ed64e1e

It will fail in Debug builds but will do nothing in prod builds.

By documentation:

```Swift
/// Indicates that an internal sanity check failed.
///
/// Use this function to stop the program, without impacting the performance of
/// shipping code, when control flow is not expected to reach the call---for
/// example, in the `default` case of a `switch` where you have knowledge that
/// one of the other cases must be satisfied. To protect code from invalid
/// usage in Release builds, see `preconditionFailure(_:file:line:)`.
///
/// * In playgrounds and -Onone builds (the default for Xcode's Debug
///   configuration), stop program execution in a debuggable state after
///   printing `message`.
///
/// * In -O builds, has no effect.
///
/// * In -Ounchecked builds, the optimizer may assume that this function is
///   never called. Failure to satisfy that assumption is a serious
///   programming error.
///
/// - Parameters:
///   - message: A string to print in a playground or `-Onone` build. The
///     default is an empty string.
///   - file: The file name to print with `message`. The default is the file
///     where `assertionFailure(_:file:line:)` is called.
///   - line: The line number to print along with `message`. The default is the
///     line number where `assertionFailure(_:file:line:)` is called.
```

k0nserv: Yeah the way you have rewritten it now works 👍🏼

k0nserv:
We should add an exampel in the app of this behaviour to

paulomendes:
> We should add an exampel in the app of this behaviour to, we can merge this and than add that later if you wish or you can do it in this PR?

Hey @k0nserv , take a look on it. I don't know if I did in the right way. Let me know if you are cool in merge this as well, otherwise I can revert this commit. -> 21c95db

PR #812, opened 2020-11-25
[BPK-4297] Add UITabBarItem+Backpack category

georgegillams:

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


Backpack/TabBarController/Classes/BPKTabBarController.m
```diff
@@ -16,21 +16,37 @@
...
 
-/*
- * Note that in some places we are having to resolve the correct UIColors from dynamic values ourselves to ensure
- * that we support overriding UIViewController interface styles. if we allow UIKit to resolve them for us,
- * it may do so wrongly.
- * This is because UIKit only sets the UIView's trait collection on a handful of lifecycle methods, and we
- * are using dynamic colours outside of these. If we do not manually resolve colours using self.traitCollection,
- * then UIKit will fall back to using [UITraitCollection currentTraitCollection] which does not reflect the override.
- */
-
```
georgegillams, line 27: This was copy-pasta'd from BPKButton

Backpack/TabBarController/Classes/BPKTabBarItem.m
```diff
@@ -0,0 +1,75 @@
...
+ * limitations under the License.
+ */
+
+#import "BPKTabBarItem.h"
+
+#import <Backpack/Color.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@implementation BPKTabBarItem
```
georgegillams, line 25: This class currently only sets a default badge colour. I'll add the dot logic in a follow-up PR.

Backpack/TabBarController/Classes/BPKTabBarController.m
```diff
@@ -16,21 +16,37 @@
...
+- (instancetype _Nullable)initWithCoder:(NSCoder *)coder
+{
+    self = [super initWithCoder:coder];
+    if (self) {
+        [self setup];
+    }
+    return self;
+}
+
+-(void)setup {
```
k0nserv, line 45: This should be done in `viewDidLoad`. I suspect that using `self.tabBar` in the init like this will cause the view to load prematurely.

olliecurtis (approved):
LGTM! One small question is regarding the dot without any content, can the size be controlled as it looks a bit on the large size? 

Is that an Apple standard thing or ours? Just used to seeing them a bit smaller

k0nserv (commented):
Question: Should we use `UIAppearance` instead?

Backpack/TabBarController/Classes/BPKTabBarController.m
georgegillams: Won't that override the colour if a consumer sets it first though?
```
let tabBarController = BPKTabBarController()
tabBarController.tabBar.tintColor = ...
// viewDidLoad happens
```

georgegillams:
> LGTM! One small question is regarding the dot without any content, can the size be controlled as it looks a bit on the large size?
> 
> Is that an Apple standard thing or ours? Just used to seeing them a bit smaller

It's the apple standard. In a follow up PR I'm going to try to make it look better, but wanted to create the basic components first 👍

Backpack/TabBarController/Classes/BPKTabBarController.m
georgegillams: As for using UIAppearance, shouldn't that be more of an app-level thing rather than a side-effect of the component? Might be hard to debug if our component is changing the UIAppearance?

k0nserv: With `UIAppearance` we can have a central method to call which sets up all our styles, so no effect without calling it.

> Won't that override the colour if a consumer sets it first though?
There's will be later I think, in any case uisng `UIAppearance` would solve that issue

georgegillams: Ah good shout!

k0nserv (approved):

PR #817, opened 2020-12-02
[BPK-4310] Refactor calendar colouring to use cell data

georgegillams:

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


Backpack/Calendar/Classes/BPKCalendar.h
```diff
@@ -124,7 +84,7 @@ NS_ASSUME_NONNULL_BEGIN
  * @note This method takes precedence over `calendar:fillColorFordate:` and `calendar:titleColorForDate:` however
  * if the receiver returns the custom cell type the calendar will call those methods in response.
  */
-- (BPKCalendarDateCellStyle)calendar:(BPKCalendar *)calendar cellStyleForDate:(BPKSimpleDate *)date;
+- (id<BPKCalendarCellData> _Nullable)calendar:(BPKCalendar *)calendar cellDataForDate:(BPKSimpleDate *)date;
```
k0nserv, line 87: The docs for this method are no longer accurate

Backpack/Calendar/Classes/BPKCalendar.m
```diff
@@ -439,15 +445,13 @@ - (nullable UIColor *)calendar:(FSCalendar *)calendar
...
-        if ([self.delegate respondsToSelector:@selector(calendar:cellStyleForDate:)]) {
+        if ([self.delegate respondsToSelector:@selector(calendar:cellDataForDate:)]) {
             BPKSimpleDate *simpleDate = [[BPKSimpleDate alloc] initWithDate:date forCalendar:self.gregorian];
 
-            BPKCalendarDateCellStyle style = [self.delegate calendar:self cellStyleForDate:simpleDate];
+            id<BPKCalendarCellData> cellData = [self.delegate calendar:self cellDataForDate:simpleDate];
 
-            // For the custom style we fallback to calling
-            // `calendar:fillColorForDate:` below
-            if (style != BPKCalendarDateCellStyleCustom) {
```
k0nserv, line 449: Doesn't RN rely on this still?

Backpack/Calendar/Classes/BPKCalendarCellData.h
```diff
@@ -0,0 +1,29 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+NS_ASSUME_NONNULL_BEGIN
+
+@protocol BPKCalendarCellData <NSObject>
+
+@required
+@property(strong) UIColor *backgroundColor;
```
k0nserv, line 24: `readonly` and `nonatomic` no?

Backpack/Calendar/Classes/BPKCalendarConfiguration.h
```diff
@@ -0,0 +1,31 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface BPKCalendarConfiguration : NSObject
```
k0nserv, line 21: This class and its properties/methods are not documented

Backpack/Calendar/Classes/BPKCalendarTrafficLightCellData.h
```diff
@@ -0,0 +1,36 @@
...
+#import "BPKCalendarCellData.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface BPKCalendarTrafficLightCellData : NSObject<BPKCalendarCellData>
+
+@property(class, readonly) BPKCalendarTrafficLightCellData *normal;
+@property(class, readonly) BPKCalendarTrafficLightCellData *positive;
+@property(class, readonly) BPKCalendarTrafficLightCellData *negative;
+@property(class, readonly) BPKCalendarTrafficLightCellData *noData NS_SWIFT_NAME(noData);
```
k0nserv, line 28: Why is `NS_SWIFT_NAME` needed if the name is the same?

Backpack/Calendar/Classes/BPKCalendarTrafficLightCellData.h
```diff
@@ -0,0 +1,36 @@
...
+@property(class, readonly) BPKCalendarTrafficLightCellData *neutral;
+
+@property(strong) UIColor *backgroundColor;
+@property(strong) UIColor *foregroundColor;
```
k0nserv, line 32: `readonly` and `nonatomic`

Backpack/Calendar/Classes/BPKCalendarCellData.h
```diff
@@ -0,0 +1,29 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+NS_ASSUME_NONNULL_BEGIN
+
+@protocol BPKCalendarCellData <NSObject>
```
k0nserv, line 21: Should all cell data provide `foregroundColor` and `backgroundColor`? Isn't that specific to the traffic light version?

I expect this to have a method like `- (void)configureCell:(BPKCalendarDateCell *)cell`

Backpack/Calendar/Classes/BPKCalendarTrafficLightCellData.m
```diff
@@ -0,0 +1,96 @@
...
+#import <Backpack/Color.h>
+
+@implementation BPKCalendarTrafficLightCellData
+
++ (BPKCalendarTrafficLightCellData *)normal {
+    static dispatch_once_t onceToken;
+    static BPKCalendarTrafficLightCellData  *cellData;
+
+    dispatch_once(&onceToken, ^{
+        BPKAssertMainThread();
```
k0nserv, line 31: I don't think we want this here. In general the code here should be thread safe and doesn't have to be on the main thread, AFAIK `UIColor` are okay to use from non-main thread

Backpack/Calendar/README.md
```diff
@@ -73,7 +73,6 @@ extension MyClass: BPKCalendarDelegate {
 
 ### Priced calendar
```
k0nserv, line 74: Maybe we should rename this

Backpack/Calendar/Classes/BPKCalendar.m
georgegillams: Yeah but I guess we can migrate it to instead return one of the instances of `BPKCalendarTrafficLightCellData`?

Backpack/Calendar/Classes/BPKCalendarTrafficLightCellData.h
georgegillams: Because Swift infers the name `no` from `noData`.

Backpack/Calendar/Classes/BPKCalendarCellData.h
georgegillams: Ah yeah - the issue with this is that we don't actually set those things on the cell atm. FSCalendar has a delegate that asks for the background and foreground colours. I wasn't sure how to make that work with the new mechanism 🤔

Backpack/Calendar/Classes/BPKCalendarTrafficLightCellData.m
georgegillams: Tru dat!

Backpack/Calendar/Classes/BPKCalendar.m
k0nserv: We should verify that this will indeed be sufficient

Backpack/Calendar/Classes/BPKCalendarTrafficLightCellData.h
k0nserv: Fair

Backpack/Calendar/Classes/BPKCalendar.m
georgegillams: I can't see any reason why not.
It also uses the `DONT_USE_calendar` delegate methods but those are still intact

Backpack/Calendar/Classes/BPKCalendarCellData.h
georgegillams: Update: have refactored this. `BPKCalendarCellData` is now empty so has been removed.

Backpack/Calendar/Classes/BPKCalendar.h
```diff
@@ -116,15 +74,14 @@ NS_ASSUME_NONNULL_BEGIN
...
  *
- * @param calendar The backpack calendar requesting the style.
- * @param date The date to provide the style for.
- * @return The cell style to use for this date.
- * @note This method takes precedence over `calendar:fillColorFordate:` and `calendar:titleColorForDate:` however
- * if the receiver returns the custom cell type the calendar will call those methods in response.
+ * @param calendar The backpack calendar requesting the data.
+ * @param date The date to provide the data for.
+ * @return The cell data to use for this date.
+ * @note This method takes precedence over `calendar:fillColorFordate:` and `calendar:titleColorForDate:`. If you return nil in this method, the values from the other methods will be used if provided.
```
k0nserv, line 82: These methods are gone, no?

Backpack/Calendar/Classes/BPKCalendar.m
```diff
@@ -199,6 +204,15 @@ - (void)setup {
     [self addSubview:self.yearPill];
 }
 
+- (void)setConfiguration:(BPKCalendarConfiguration *)configuration {
+    if (_configuration != configuration) {
+        _configuration = configuration;
+        [self.calendarView registerClass:configuration.cellClass forCellReuseIdentifier:CellReuseId];
```
k0nserv, line 210: I wonder about the implications of this when registerig a new class with a previously used identifier

Backpack/Calendar/Classes/BPKCalendarCell.m
```diff
@@ -212,4 +212,8 @@ - (void)setSelectionType:(SelectionType)selectionType {
     }
 }
 
+-(void)configureWithData:(id)data {
+    return;
```
k0nserv, line 216: Since this is an abstract base class we should have an assert here or throw an exception

Backpack/Calendar/Classes/BPKCalendarColor.h
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * `BPKCalendarColor` contains a number of pre-defined colours commonly used when presenting a calendar.
+ */
+@interface BPKCalendarColor : NSObject
+
+/**
+ * The normal calendar background color.
+ */
+@property(class, readonly) UIColor *normalColor;
```
k0nserv, line 30: Should we create a `BPKColorPair` class to avoid having two properties per class?

Backpack/Calendar/Classes/BPKCalendarTrafficLightCellData.m
```diff
@@ -0,0 +1,93 @@
...
+#import "BPKCalendarColor.h"
+
+@implementation BPKCalendarTrafficLightCellData
+
++ (BPKCalendarTrafficLightCellData *)normal {
+    static dispatch_once_t onceToken;
+    static BPKCalendarTrafficLightCellData  *cellData;
+
+    dispatch_once(&onceToken, ^{
+        cellData = [[BPKCalendarTrafficLightCellData alloc] initWithBackgroundColor:BPKCalendarColor.normalColor foregroundColor:BPKCalendarColor.normalTitleColor];
```
k0nserv, line 33: nit: These lines are slightly long

Backpack/Calendar/Classes/BPKCalendar.h
georgegillams: Nope - they're prefixed with DONT_USE to discourage using them

Backpack/Calendar/Classes/BPKCalendar.m
georgegillams: Yeah i was also thinking maybe the configuration should be set at initialisation. 
`initWithConfig:myConfiguration...`. Might make things easier

Backpack/Calendar/Classes/BPKCalendarColor.h
georgegillams: Maybe, but the hard thing is the background color in one use case is the forground color in another 🤔

Backpack/Calendar/Classes/BPKCalendar.h
k0nserv: Then the name we use here is wrong? We should remove the note imo

Backpack/Calendar/Classes/BPKCalendar.m
k0nserv: Sounds good

Backpack/Calendar/Classes/BPKCalendar.h
georgegillams: Yeah I guess as it's deprecated that would make sense

Backpack/Calendar/Classes/BPKCalendarColor.m
```diff
@@ -0,0 +1,136 @@
...
+
+#import "BPKCalendarColor.h"
+
+#import <Backpack/Common.h>
+#import <Backpack/Color.h>
+
+@implementation BPKCalendarColor
+
++ (UIColor *)normalColor {
+    static dispatch_once_t onceToken;
```
k0nserv, line 27: It's not worth it to use dispatch once for this, the overhead is too high

Backpack/Calendar/Classes/BPKCalendarColor.h
k0nserv: Yeah you're right. I am thinking of the case when we have a pair were the `foregroundColor` is suitable to put on top of a view with the specific backgroundColor

georgegillams: tbh I think it's probably yagni. We're gonna be the main consumers of pairs of colours anyway 🤷

k0nserv (approved):

PR #821, opened 2020-12-08
[BPK-4310] Add price configuration

georgegillams:

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


Backpack/Calendar/Classes/BPKCalendarPriceLabelCell.h
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+#import "BPKCalendarCell.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@class BPKLabel;
+
+/**
+ * `BPKCalendarPriceLabelCell` is a subclass of `BPKCalendarCell`. It allows cusotmisation of the cell fill colour and
```
k0nserv, line 26: > cusotmisation

k0nserv: > cell fill colour

Does it?

Backpack/Calendar/Classes/BPKCalendarPriceLabelCell.m
```diff
@@ -0,0 +1,80 @@
...
+- (instancetype)initWithFrame:(CGRect)frame {
+    self = [super initWithFrame:frame];
+    if (self) {
+        [self setup];
+    }
+    return self;
+}
+
+- (void)setup {
+    _priceLabel = [[BPKLabel alloc] initWithFontStyle:BPKFontStyleTextSm];
```
k0nserv, line 54: Nitpic: I prefer to explicitly delcare the read/write property in the `.m` file and use `self.priceLabel`. You could also use a lazy property

Backpack/Calendar/Classes/BPKCalendarPriceLabelCell.m
```diff
@@ -0,0 +1,80 @@
...
+    self.priceLabel.numberOfLines = 0;
```
k0nserv, line 55: Shouldnt' this be two?

Backpack/Calendar/Classes/BPKCalendarPriceLabelCellData.h
```diff
@@ -0,0 +1,49 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+NS_ASSUME_NONNULL_BEGIN
+
+@class BPKCalendarPriceLabelStyle;
+
+/**
+ * `BPKCalendarPriceLabelCellData` contains the properties required to customise a calendar cell under a traffic-light
```
k0nserv, line 24: This is out of date(copy and paste error)

Backpack/Calendar/Classes/BPKCalendarPriceLabelCellData.m
```diff
@@ -0,0 +1,36 @@
...
+
+#import "BPKCalendarPriceLabelStyle.h"
+
+@implementation BPKCalendarPriceLabelCellData
+
+- (instancetype)initWithPrice:(NSString *)price labelStyle:(BPKCalendarPriceLabelStyle *)labelStyle {
+    self = [super init];
+
+    if (self) {
+        _price = price;
```
k0nserv, line 29: We should copy price here `_price = [price copy]`

Backpack/Calendar/Classes/BPKCalendarPriceLabelStyle.h
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * The foreground colour to use on the cell's price label.
+ */
+@property(nonatomic, readonly) UIColor *color;
+
+/**
+ * Whether the price should be emphasized or not.
+ */
+@property(nonatomic, readonly) BOOL bold;
```
k0nserv, line 44: `getter=isBold`?

k0nserv (commented):
Simulator Screen Shot - iPhone 5s - 2020-12-14 at 13 33 57

It's still a bit broken in the most extreme case(92884400₫ is 4000 USD)

Backpack/Calendar/Classes/BPKCalendarPriceLabelCell.m
```diff
@@ -0,0 +1,87 @@
...
+}
+
+- (void)prepareForReuse {
+    [super prepareForReuse];
+
+    self.priceLabel.text = nil;
+}
+
+- (void)configureWithData:(id)data {
+    BPKCalendarPriceLabelCellData *priceLabelCellData = (BPKCalendarPriceLabelCellData *)data;
```
k0nserv, line 81: Probably worth adding an assert here to make sure the data is of the expected type

Backpack/Calendar/Classes/BPKCalendarPriceLabelCell.h
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * `BPKCalendarPriceLabelCell` is a subclass of `BPKCalendarCell`. It allows a custom label containing auxillary price information to be shown below the cell.
+ */
+@interface BPKCalendarPriceLabelCell : BPKCalendarCell
+
+/**
+ * The label displayed at the bottom of the cell.
+ */
+@property(nonatomic, strong, nonnull, readonly) BPKLabel *priceLabel;
```
k0nserv, line 33: `nonnull` is redundant because of `NS_ASSUME_NONNULL_BEGIN` and `NS_ASSUME_NONNULL_END`

Backpack/Calendar/Classes/BPKCalendarPriceLabelCell.m
```diff
@@ -0,0 +1,87 @@
...
+
+- (void)setup {
+    self.priceLabel = [[BPKLabel alloc] initWithFontStyle:BPKFontStyleTextXs];
+    self.priceLabel.numberOfLines = 2;
+    self.priceLabel.lineBreakMode = NSLineBreakByTruncatingTail;
+    [self.contentView addSubview:self.priceLabel];
+    self.priceLabel.translatesAutoresizingMaskIntoConstraints = NO;
+    self.priceLabel.textAlignment = NSTextAlignmentCenter;
+    [NSLayoutConstraint activateConstraints:@[
+        [self.priceLabel.topAnchor constraintEqualToAnchor:self.titleLabel.bottomAnchor constant:BPKSpacingSm / 2.0],
```
k0nserv, line 67: Is `BPKSpacingSm/2.0` even legal? 🤔 👮🏼

Backpack/Calendar/Classes/BPKCalendarPriceLabelStyle.h
```diff
@@ -0,0 +1,63 @@
...
+ * @param color The text color to apply.
+ * @param bold Whether the font should be bold or not.
+ * @return `BPKCalendarPriceLabelStyle` instance.
+ */
+- (instancetype)initWithColor:(UIColor *)color bold:(BOOL)bold;
+
+/**
+ * The foreground colour to use on the cell's price label.
+ */
+@property(nonatomic, readonly) UIColor *color;
```
k0nserv, line 39: Maybe we should call this `foregroundColor` or `textColor` to make it less ambiguous?

Example/SnapshotTests/BPKCalendarSnapshotTest.m
```diff
@@ -44,6 +45,7 @@ - (void)setUp {
     self.date1 = [NSDate dateWithTimeIntervalSince1970:2175785688];
     self.date2 = [NSDate dateWithTimeIntervalSince1970:2176044888];
     self.isColoringDates = NO;
+    self.isShowingPrices = NO;
```
k0nserv, line 48: This will break if we ever run these tests concurrenctly(is that even possible?)

Example/SnapshotTests/BPKCalendarSnapshotTest.m
```diff
@@ -205,26 +222,26 @@ - (id _Nullable)calendar:(BPKCalendar *)calendar cellDataForDate:(BPKSimpleDate
     NSDateComponents *components = [cal components:NSCalendarUnitDay fromDate:date1 toDate:date2 options:0];
 
     if (components.day == 2 || components.day == 8 || components.day == 12 || components.day == 20) {
-        return BPKCalendarTrafficLightCellData.positive;
+        return self.isShowingPrices ? [[BPKCalendarPriceLabelCellData alloc] initWithPrice:@"£14" labelStyle:BPKCalendarPriceLabelStyle.positive] : BPKCalendarTrafficLightCellData.positive;
     }
 
     if (components.day == 4 || components.day == 10 || components.day == 24) {
-        return BPKCalendarTrafficLightCellData.negative;
+        return self.isShowingPrices ? [[BPKCalendarPriceLabelCellData alloc] initWithPrice:@"11.884.400₫" labelStyle:BPKCalendarPriceLabelStyle.negative] : BPKCalendarTrafficLightCellData.negative;
```
k0nserv, line 229: In the resulting snapshot image this prices is truncated. I thought it shouldn't be 🤔

k0nserv (commented):
How does this work with VoiceOver?

Backpack/Calendar/Classes/BPKCalendarPriceLabelCell.m
georgegillams: The BPKCalendar already asserts it before it calls this method. Do you want a double assert? Or just move the assert to here?

IMO it makes more sense for the assert to be here as this method could be called (in theory) from anywhere.

Example/SnapshotTests/BPKCalendarSnapshotTest.m
georgegillams: It seemed to work before for `isColouringDates` so I assume it's fine.
I'm pretty sure the tests never run concurrently.

Example/SnapshotTests/BPKCalendarSnapshotTest.m
georgegillams: Because the snapshot view is smaller than a real iPhone.

Backpack/Calendar/Classes/BPKCalendarPriceLabelCell.m
k0nserv: We can double assert 👍🏼 It will get compiled out in production builds so the performance implications are moot anyway. We sould safeguard so we don't crash in prod here though

Example/SnapshotTests/BPKCalendarSnapshotTest.m
k0nserv: Can we use a more realistic frame for this test then?

georgegillams: IMO it's good to have it truncating because it means truncating is working.
We could make the screenshot bigger, and add another longer truncating case, but I can't see what it achieves?

k0nserv: The snapshot tests should catch us if we accidentally break something. We've designed this version of the calendar to work correctly with the edge case of tens of million dongs i.e. "99.999.999₫" shouldn't truncate. As it stands if we break this behaviour the test wouldn't catch this regression and we wouldn't notice that we broken a valid case.

Example/SnapshotTests/BPKCalendarSnapshotTest.m
k0nserv: It's probably the reverse i.e. that both would break but we don't run concurrently.

Example/SnapshotTests/BPKCalendarSnapshotTest.m
georgegillams: Ok cool I'll update it then 👍

Backpack/Calendar/Classes/BPKCalendarPriceLabelCell.m
georgegillams: BPKSpacingSm is just a float:
`CGFloat const BPKSpacingSm = 4;`

So don't see why it wouldn't be?

georgegillams: This is what the design wants, and the spacings here's are a bit custom because of the underlying library etc anyway, so I think it's fine?

Example/SnapshotTests/BPKCalendarSnapshotTest.m
```diff
@@ -201,26 +218,26 @@ - (id _Nullable)calendar:(BPKCalendar *)calendar cellDataForDate:(BPKSimpleDate
     NSDateComponents *components = [cal components:NSCalendarUnitDay fromDate:date1 toDate:date2 options:0];
 
     if (components.day == 2 || components.day == 8 || components.day == 12 || components.day == 20) {
-        return BPKCalendarTrafficLightCellData.positive;
+        return self.isShowingPrices ? [[BPKCalendarPriceLabelCellData alloc] initWithPrice:@"458.100₫" labelStyle:BPKCalendarPriceLabelStyle.positive] : BPKCalendarTrafficLightCellData.positive;
     }
 
     if (components.day == 4 || components.day == 10 || components.day == 24) {
-        return BPKCalendarTrafficLightCellData.negative;
+        return self.isShowingPrices ? [[BPKCalendarPriceLabelCellData alloc] initWithPrice:@"This should truncate" labelStyle:BPKCalendarPriceLabelStyle.negative] : BPKCalendarTrafficLightCellData.negative;
```
k0nserv, line 225: Neat 👍🏼

k0nserv (approved):
Awesome :shipit:

PR #823, opened 2020-12-08
[BPK-4310] Ensure cell rendering is consistent

georgegillams:

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


Backpack/Calendar/Classes/BPKCalendar.m
```diff
@@ -170,6 +170,9 @@ - (void)setup {
     self.gregorian = [[NSCalendar alloc] initWithCalendarIdentifier:NSCalendarIdentifierGregorian];
 
     self.calendarView = [[FSCalendar alloc] initWithFrame:CGRectZero];
+    if (self.configuration.rowHeight > 0) {
+        self.calendarView.rowHeight = self.configuration.rowHeight;
+    }
```
georgegillams, line 175: We need to make this configurable so that the price label config can specify a taller cell size.

Backpack/Calendar/Classes/BPKCalendarCell.m
```diff
@@ -56,27 +54,29 @@ - (instancetype)initWithFrame:(CGRect)frame {
...
 
-    CGFloat padding = (CGRectGetWidth(self.bounds) - CGRectGetWidth(self.shapeLayer.bounds)) / 2.0 - 0.5;
-    if (padding > CGRectGetWidth(self.shapeLayer.bounds) / 8.0) {
-        padding = CGRectGetWidth(self.shapeLayer.bounds) / 8.0;
-    }
+    CGFloat TITLE_HEIGHT = 31;
+    CGFloat CIRCLE_HEIGHT = 31.6;
+    CGFloat sameDayOffset = 3.75;
+
+    CGFloat paddingX = (CGRectGetWidth(self.bounds) - CIRCLE_HEIGHT) / 2.0;
```
georgegillams, line 61: I've reworked this as before it made assumptions about the background being the full height of the cell, and the label being placed centrally.
In the new price config, we'll need the circle and title to stay still while we add additional content below.

Backpack/Calendar/Classes/BPKCalendarCell.m
```diff
@@ -56,27 +54,29 @@ - (instancetype)initWithFrame:(CGRect)frame {
 - (void)layoutSubviews {
     [super layoutSubviews];
 
-    CGFloat padding = (CGRectGetWidth(self.bounds) - CGRectGetWidth(self.shapeLayer.bounds)) / 2.0 - 0.5;
-    if (padding > CGRectGetWidth(self.shapeLayer.bounds) / 8.0) {
-        padding = CGRectGetWidth(self.shapeLayer.bounds) / 8.0;
-    }
+    CGFloat titleHeight = 31;
```
k0nserv, line 57: Can we move these magic numbers to file level constants instead of variables in the method?

k0nserv (commented):
It strikes me that this layout will break completely when we adopt dynamic text. Maybe we should look at preemptively opting out of dynamic text here to prevent future pain?

georgegillams:
> It strikes me that this layout will break completely when we adopt dynamic text. Maybe we should look at preemptively opting out of dynamic text here to prevent future pain?

I guess the calendar as a whole will break as cells will be too narrow etc.

AFAIK all BPKText is opted out of dynamic? I think when we move to support it we'll have to review every component?

Backpack/Calendar/Classes/BPKCalendarCell.m
```diff
@@ -56,27 +54,29 @@ - (instancetype)initWithFrame:(CGRect)frame {
...
 
     self.shapeLayer.hidden = NO;
     self.selectionLayer.hidden = NO;
-    self.selectionLayer.frame =
-        CGRectMake(0.0, CGRectGetMinY(self.shapeLayer.frame), CGRectGetWidth(self.bounds), CGRectGetHeight(self.shapeLayer.frame));
+    self.selectionLayer.frame = CGRectMake(0.0, 0, circleHeight, circleHeight);
+
+    self.layer.borderColor = [BPKColor.panjin CGColor];
+    self.titleLabel.frame = CGRectMake(0, 0, self.titleLabel.frame.size.width, titleHeight);
+    self.titleLabel.layer.borderColor = [BPKColor.panjin CGColor];
```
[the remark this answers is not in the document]
georgegillams: Probably not. It was being done before, but I'm sure we can refactor it not to 👍

Backpack/Calendar/Classes/BPKCalendarCell.m
```diff
@@ -56,27 +54,29 @@ - (instancetype)initWithFrame:(CGRect)frame {
...
+    CGFloat paddingX = (CGRectGetWidth(self.bounds) - circleHeight) / 2.0;
 
     self.shapeLayer.hidden = NO;
     self.selectionLayer.hidden = NO;
-    self.selectionLayer.frame =
-        CGRectMake(0.0, CGRectGetMinY(self.shapeLayer.frame), CGRectGetWidth(self.bounds), CGRectGetHeight(self.shapeLayer.frame));
+    self.selectionLayer.frame = CGRectMake(0.0, 0, circleHeight, circleHeight);
+
+    self.layer.borderColor = [BPKColor.panjin CGColor];
+    self.titleLabel.frame = CGRectMake(0, 0, self.titleLabel.frame.size.width, titleHeight);
```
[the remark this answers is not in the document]
georgegillams: FSCalendar set it to the full size of the cell, so we're overriding that behaviour and making it stick to the correct size.

k0nserv:
> AFAIK all BPKText is opted out of dynamic? I think when we move to support it we'll have to review every component?
Yes, but would it save  us time to opt the calendar out ahead of time, or do we expect we'll support it here?

Backpack/Calendar/Classes/BPKCalendarConfiguration.h
```diff
@@ -34,6 +34,21 @@ NS_ASSUME_NONNULL_BEGIN
...
+
+/**
+ * Creates a `BPKCalendarConfiguration` with the specific classes.
+ *
+ * @param cellClass Class to use for individual cells.
+ * @param cellDataClass Class used to represent additional cell-data.
+ * @param rowHeight The height that calendar rows should be rendered,.
+ * @return `BPKCalendarConfiguration` instance.
+ */
+- (instancetype)initWithCellClass:(Class)cellClass cellDataClass:(Class)cellDataClass rowHeight:(CGFloat)rowHeight;
```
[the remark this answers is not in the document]
georgegillams: I think autocomplete should do that for us?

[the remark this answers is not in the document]
k0nserv: Sorry I was probably unclear. What I am getting at is how to set `rowHeight` to the default value and what impact that has on the calendar. Something like 

> When rowHeight isn't specified the calendar uses a default row height for each cell.

maybe as part of `- (instancetype)initWithCellClass:(Class)cellClass cellDataClass:(Class)cellDataClass;` or attached to the `@property`

Backpack/Calendar/Classes/BPKCalendarCell.m
```diff
@@ -25,17 +25,30 @@
 #import <Backpack/Font.h>
 #import <Backpack/Spacing.h>
 
-#import "BPKCalendar.h"
-
 @interface BPKCalendarCell ()
 
 @property(weak, nonatomic) CAShapeLayer *selectionLayer;
 @property(weak, nonatomic) CAShapeLayer *samedayLayer;
+@property(class, readonly) CGFloat titleHeight;
```
k0nserv, line 32: Probably better to use C constants at the file level instead. This is somewhat wasteful as it will involve generating and calling class method via message dispatch.

So instead of `@property(class, readonly) CGFloat titleHeight;` + the method implementation use `const CGFloating BPKCalendarCellTitleHeight = 31.0;` at the top level.

[`BPKNavigationBar.m`](https://github.com/Skyscanner/backpack-ios/blob/master/Backpack/NavigationBar/Classes/BPKNavigationBar.m#L30) has an example.

Backpack/Calendar/Classes/BPKCalendarCell.m
```diff
@@ -164,9 +171,16 @@ - (void)layoutSubviews {
 - (void)configureAppearance {
     [super configureAppearance];
 
+    UIColor *rangeColor = self.appearance.selectionColor;
     UIColor *selectedColor = self.preferredTitleSelectionColor ?: self.appearance.titleColors[@(FSCalendarCellStateSelected)];
     UIColor *color = self.preferredTitleDefaultColor ?: [self colorForCurrentStateInDictionary:self.appearance.titleColors];
 
+    self.selectionLayer.fillColor = rangeColor.CGColor;
```
k0nserv, line 178: Does this work correctly if the user switches appearance(light vs dark) while the calendar is on screen? `CGColor` isn't appearance aware like `UIColor`

Backpack/Calendar/Classes/BPKCalendarCell.m
georgegillams: Thanks - that's much better!

Backpack/Calendar/Classes/BPKCalendarCell.m
georgegillams: Yeah it does 👍

k0nserv (approved):

PR #830, opened 2020-12-16
[BPK-4310] Update calendar snapshot test size

georgegillams:


Example/SnapshotTests/BPKCalendarSnapshotTest.m
```diff
@@ -60,11 +61,11 @@ - (void)configureParentView:(UIView *)parentView forCalendar:(BPKCalendar *)cale
                                                                        options:0
                                                                        metrics:nil
                                                                          views:@{@"calendar": calendar}]];
-    [parentView addConstraints:[NSLayoutConstraint constraintsWithVisualFormat:@"[parentView(320)]"
```
k0nserv, line 63: Does this mean the current calendar design will not work on the previous generation of iPhone SE?

georgegillams: You mean for price labels?
I guess if the iPhone SE has a screen width of 320 then no 🤷

k0nserv: Okay that's probably worth highlighting too

k0nserv (approved):
LGTM

Example/SnapshotTests/BPKCalendarSnapshotTest.m
georgegillams: Do we still support iOS 12?

georgegillams: iOS 13 supports the SE 😮

k0nserv: Some iPod is also 320pt wide I believe

PR #833, opened 2020-12-17
[BPK-4321] Upgrade floating-panel dependency

k0nserv:
<details>
<summary>After(2.1.0)</summary>

https://user-images.githubusercontent.com/1333960/102518085-c0d75700-4088-11eb-9128-c9aa544bcd4b.mov

</details>

<details>
<summary>Before(1.6.6)</summary>

https://user-images.githubusercontent.com/1333960/102517113-8d47fd00-4087-11eb-986a-55c2eafa0e5f.mov

</details>

There's some wonkiness with the intrinsic content size version still. I'm unsure if it's even possible to prevent the scrolling entirely and there's a bit of weird text trunacting going on with the static case

+ [ ] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/.spelling
+++ b/.spelling
@@ -190,6 +190,7 @@ div
 divs
 env
 fieldset
+FloatingPanel
 fontWeight
 globals
 gray
--- a/Backpack.podspec
+++ b/Backpack.podspec
@@ -16,7 +16,7 @@
 
 Pod::Spec.new do |s|
   s.name             = 'Backpack'
-  s.version          = "39.1.1"
+  s.version          = "39.2.0-alpha.1"
   s.summary          = "Skyscanner's Design System Backpack for iOS"
 
   s.description      = <<-DESC
--- a/Backpack.podspec
+++ b/Backpack.podspec
@@ -37,7 +37,7 @@ Pod::Spec.new do |s|
   s.ios.resource_bundle = {
     'Icon' => 'Backpack/Icon/Assets/*'
   }
-  s.dependency 'FloatingPanel', '1.6.6'
+  s.dependency 'FloatingPanel', '2.1'
   s.dependency 'FSCalendar', '~> 2.8'
   s.dependency 'TTTAttributedLabel', '~> 1.13.4'
   s.dependency 'MBProgressHUD', '~> 0.9.1'
--- a/Backpack/BottomSheet/Classes/BackpackFloatingPanelController.swift
+++ b/Backpack/BottomSheet/Classes/BackpackFloatingPanelController.swift
@@ -31,6 +31,10 @@ final class BPKFloatingPanelController: FloatingPanelController {
         }
     }
 
+    override func viewDidLoad() {
+        backdropView.dismissalTapGestureRecognizer.isEnabled = true
+    }
+
     override func viewDidLayoutSubviews() {
         super.viewDidLayoutSubviews()
         updateScrollViewInsetsForBottomSection()
@@ -85,8 +89,9 @@ private extension BPKFloatingPanelController {
             bottomSection.view.trailingAnchor.constraint(equalTo: view.trailingAnchor),
             bottomContainerConstraint,
 
+            // Ensure the bottomSection is pushed off the screen when the bottom-sheet is less than half-shown
             outsideSafeAreaView.topAnchor.constraint(greaterThanOrEqualTo: content.view.topAnchor,
-                                                     constant: layout.insetFor(position: .half) ?? 0)
+                                                     constant: BPKBottomSheet.Constants.bottomSheetHeightInHalfPosition)
         ])
     }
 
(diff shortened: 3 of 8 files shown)


fedetrim:
We are using the "non-scrollable" style a lot in Trips, and looks like with the new changes, this style is having issues with the safe area layout guide and the intrinsic size of the content.

Do you think it's related to https://github.com/Skyscanner/backpack-ios/pull/833/files#diff-d16f28063a7492d7af08a8c3e66527bd269c890722eb29c968bee41aea7f38efR203-R207 ...?

Maybe we should use `safeArea` instead of `superview`/

k0nserv:
> We are using the "non-scrollable" style a lot in Trips, and looks like with the new changes, this style is having issues with the safe area layout guide and the intrinsic size of the content.
> 
> 
> 
> Do you think it's related to https://github.com/Skyscanner/backpack-ios/pull/833/files#diff-d16f28063a7492d7af08a8c3e66527bd269c890722eb29c968bee41aea7f38efR203-R207 ...?
> 
> 
> 
> Maybe we should use `safeArea` instead of `superview`?

I tried both `safeArea` and `superView`. I believe I settled on `superView` because it matched the current behaviour. I don't recall if `safeArea` exhibit the same layout issues, will test this tomorrow and get back to you

georgegillams:
As far as I can tell this is working well, apart from this strange behaviour:


Is this the same issue you're referring to @fedetrim?

fedetrim:
@georgegillams No, I'm referring to second 18 of the first video fo the description, when you tap on "Bottom Sheet with a non-scrollable content". Notice how the sheet doesn't have the right size to fit the content, and seems to be ignoring safe area as well. Also the size changes a few times in a strange way as you scroll

k0nserv:
I don't think there's a way to lock the scroll in place with this new API anymore, at least I haven't been able to spot anything like it

fedetrim:
@k0nserv 
I think we don't need to lock the scroll in place though as long as we can make it work without any UI glitches

k0nserv:
Update it to use `.safeArea` instead, this doesn't have the visual issues of `.superView` but does mean the height will differ from the previous version.

k0nserv:
https://github.com/Skyscanner/backpack-ios/pull/833/commits/01422987548d67736d5951799ac462ff4916dc75 fixes the issues with dragging when using the intrinsic size mode(by disabling it entirely). 

I think  now this should be good to merge

georgegillams:
> [0142298](https://github.com/Skyscanner/backpack-ios/commit/01422987548d67736d5951799ac462ff4916dc75) fixes the issues with dragging when using the intrinsic size mode(by disabling it entirely).
> 
> I think now this should be good to merge.
> 
> This does mean you cannot swipe to dimiss non scrolalble panels anymore, I'm inclined to say we should keep the grabbable behaviour 🤔

I feel like swiping to dismiss is going to be the most expected behaviour from users 🤔

k0nserv:
I agree 👍🏼 Let me remove that last part hten

georgegillams:
This is still a little funky 🤔

k0nserv:
How should that work in your opinion?

fedetrim:
> This is still a little funky 🤔
> 

Yes, there's definitely an issue when calculating the size of the content. I think for scrolling it doesn't consider safe area but when animation ends, safe area is suddenly added and the size jumps to the correct value making it look like a UI glitch.
TBH I don't think this is an acceptable UX/UI, we should test integrating this branch into the main Skyscanner app and see how it looks on real features

k0nserv:
Yeah, I'm going to checkout the examples in the main repo but I think that's intended behaviour

k0nserv:
I've found and fixed the issues. The constraints for the views inside the bottom sheet should not reference the safearea, they should reference the superview.

georgegillams (approved):
It's beautiful!

fedetrim:
@k0nserv 
>The constraints for the views inside the bottom sheet should not reference the safearea, they should reference the superview.

I think that's a workaround not ideal because if we wanted to reuse a view controller for presenting it in two different styles (maybe as a native modal, and also as a bottom sheet), then native modal would need constraints referencing safearea, but bottom sheet would need superview. So it's inconsistent.

However it might be an acceptable tradeoff in order to be able to update this framework to the latest version. Before merging and releasing this, it would be nice to have a branch in the main Skyscanner app repo using this branch of Backpack, so that we can make sure that all bottom sheets look and behave correctly and update the code if needed. Then we can release this, and then update the app repo branch to use the Backpack release version

k0nserv:
Yeah I agree we need to see what the impact is on the live app, however this seems like a better outcome than the janky behaviour we'd otherwise get

k0nserv:
I've done more investigating with the samples from FloatingPanel, as we've noticed there are two options:

+ Live with the janky behaviour on devices with a bottom safe area(it doesn't happen on i.e. an iPhone 8)
+ Fixed it by constraining to `superView.bottom` instead of `safeArea.bottom`. 

In either case this is a documentation issue not something we can fix in Backpack. It depends on how you wire up your views.

k0nserv:
Closing for now

PR #845, opened 2020-12-30
[BPK-4344] Upgrade MBProgressHUD to 1.2.0

georgegillams:
Changelog: https://github.com/matej/MBProgressHUD/blob/master/CHANGELOG.mdown

For now I'm keeping the same style as before, but in the future maybe adopting their new background blurred style would be better?

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


Example/Backpack/View Controllers/ToastViewController.swift
```diff
@@ -19,7 +19,6 @@
 
 import UIKit
 import Backpack
-import MBProgressHUD
```
georgegillams, line 22: No idea why this needed to import it too, as importing BPKToast should be enough

k0nserv (approved):
LGTM, but we need to get some input from design I think

PR #849, opened 2020-12-31
[BPK-4298] Swiftify BPKRootListTableViewController

georgegillams:


Example/Backpack/View Controllers/BPKRootListTableViewController.m
```diff
@@ -1,57 +0,0 @@
...
-        self.settingsButton.target = self;
-        self.settingsButton.action = @selector(didTapSettingsButton);
-    }
-}
-
-- (void)didTapSettingsButton {
-    [ExampleApp showSettingsView];
-}
-
-- (void)prepareForSegue:(UIStoryboardSegue *)segue sender:(id)sender {
```
k0nserv, line 48: Don't we still need this?

georgegillams: I was thinking we could drop as we'll be migrating to using the new data, and it doesn't currently work anyway as it always says "Labels" in the app

k0nserv: huh, fair enough then

k0nserv (approved):

PR #855, opened 2021-01-05
[BPK-4347] Skip comment if tagged

georgegillams:
When we do a release, two CI jobs run. One for the push to master, and one for the push to the tag (because tags are just special branches).
The master run doesn't try to comment on the PR, but the tag one does. I've added an additional check to counter this.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -172,13 +172,13 @@ jobs:
   comment_on_pr:
     needs: CI
     name: Comment on PR
-    if: github.ref != 'refs/heads/master'
+    # Only run if there is no tag, and the commit is not on master
+    if: contains('refs/tags/', github.ref) && github.ref != 'refs/heads/master'
     runs-on: ubuntu-20.04
 
     steps:
       - name: Link to the pull request build
         uses: unsplash/comment-on-pr@master
-        if: github.ref != 'refs/heads/master'
         with:
           msg: 'Visit https://backpack.github.io/ios-prs/${{ env.PR_NUMBER }} to see the docs for this PR.'
         env:


olliecurtis (approved):
Looks good to me - One question just thinking on the small chance, what happens if its on a PR and there is for some reason a tag available? 

The comment on PR wouldn't work, so are we safe to assume PRs would never have tags or should we just have the if be for just `master`?

k0nserv:
Shouldn't we also avoid this for greenkeeeper PRs?

georgegillams:
> Shouldn't we also avoid this for greenkeeeper PRs?(both generating the docs and the comment)

We could, but what's the benefit?

k0nserv:
> We could, but what's the benefit?
Faster CI, less load on the build agents?

georgegillams:
> Looks good to me - One question just thinking on the small chance, what happens if its on a PR and there is for some reason a tag available?
> 
> The comment on PR wouldn't work, so are we safe to assume PRs would never have tags or should we just have the if be for just `master`?

I think we can assume people won't tag their PRs. And if they do they're only missing out on a link to docs 🤷‍♂️

georgegillams:
> > We could, but what's the benefit?
> > Faster CI, less load on the build agents?

Seems reasonable. I'll make a ticket in the backlog

olliecurtis:
> Shouldn't we also avoid this for greenkeeeper PRs?(both generating the docs and the comment)

We could but Greenkeeper is gone (long live Greenkeeper) so no benefit 🙊

k0nserv:
> We could but Greenkeeper is gone (long live Greenkeeper) so no benefit 🙊

Dependanbot then ;)

georgegillams:
> > Shouldn't we also avoid this for greenkeeeper PRs?(both generating the docs and the comment)
> 
> We could but Greenkeeper is gone (long live Greenkeeper) so no benefit 🙊

I think he just means Greenkeeper or equivalent, ie `dependabot`

PR #860, opened 2021-01-12
[BPK-4342] Parallelise tests

georgegillams:
Turns out the way XCode handles concurrent tests is to just boot multiple clones of the simulator, so there are no concurrency issues in our code 🙌

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/Example/SnapshotTests/BPKSnapshotTest.h
+++ b/Example/SnapshotTests/BPKSnapshotTest.h
@@ -22,7 +22,7 @@
         int correctMinorVersion = 0;                                                                                                                 \
         NSString *deviceName = [[UIDevice currentDevice] name];                                                                                      \
         NSOperatingSystemVersion deviceOSVersion = [[NSProcessInfo processInfo] operatingSystemVersion];                                             \
-        BOOL validDevice = [deviceName isEqual:correctDeviceName];                                                                                   \
+        BOOL validDevice = [deviceName hasSuffix:correctDeviceName];                                                                                   \
         if (deviceOSVersion.majorVersion != correctMajorVersion) {                                                                                   \
             validDevice = NO;                                                                                                                        \
         }                                                                                                                                            \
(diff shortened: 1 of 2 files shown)


Example/SnapshotTests/BPKSnapshotTest.h
```diff
@@ -22,7 +22,7 @@
         int correctMinorVersion = 0;                                                                                                                 \
         NSString *deviceName = [[UIDevice currentDevice] name];                                                                                      \
         NSOperatingSystemVersion deviceOSVersion = [[NSProcessInfo processInfo] operatingSystemVersion];                                             \
-        BOOL validDevice = [deviceName isEqual:correctDeviceName];                                                                                   \
```
k0nserv, line 25: Is this change unrelated then?

k0nserv (commented):
Neat, does it actually make the build faster though?

Example/SnapshotTests/BPKSnapshotTest.h
georgegillams: Nah - because when running concurrently the device name might be `iPhone 8` or `Clone 1 of iPhone 8` or `Clone 2 of iPhone 8` etc

georgegillams:
> Neat, does it actually make the build faster though?

It's quite a bit faster if you already have the clone simulators open. Average 56s vs 80s.
However, creating the extra simulators and starting them takes some time. So for now the benefit is limited (especially in CI where they won't already be running).
Even so, running concurrently will certainly keep things fast as we go forward and inevitably add more tests.

k0nserv:
>  Even so, running concurrently will help to keep things fast as we go forward and inevitably add more tests.
Should we just keep this learning for then and avoid the added complexity until then?

georgegillams:
>  Should we just keep this learning for then and avoid the added complexity until then?

Given that the added complexity is basically 0, I'd say just ship it?

PR #863, opened 2021-01-12
[BPK-4298] Remove Bar charts sub menu

georgegillams:
As it only had one sub-item (default) I've removed that layer of navigation completely


k0nserv (approved):
Neat

PR #877, opened 2021-02-02
[BPK-4355] Add map annotations

georgegillams:

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


Backpack/FlareView/Classes/BPKFlarePath.m
```diff
@@ -92,6 +92,8 @@ + (UIBezierPath *)flareViewPathForSize:(CGSize)size
         [self appendFlareToPath:path size:size flareHeight:flareHeight flarePosition:flarePosition];
     }
 
+    [path addLineToPoint:firstCornerCurveStart];
```
georgegillams, line 95: This is needed to actually close the path. Because we only used it for masking before it worked even though it was left open

Backpack/Map/Classes/BPKMapAnnotationView.m
```diff
@@ -0,0 +1,162 @@
...
+    if(shouldShowCallout && self.calloutView == nil) {
+        [self createCallout];
+    }
+    if(!shouldShowCallout && self.calloutView != nil) {
+        [self destroyCallout];
+    }
+
+    self.calloutView.selected = self.selected;
+    self.calloutView.enabled = self.enabled;
+    self.calloutView.hasBeenSelected = self.hasBeenSelected ;
```
georgegillams, line 150: I guess the callout view could just have a reference to the `BPKAnnotationView` so that we don't need to duplicate these properties? WDYT?

Backpack/FlareView/Classes/BPKFlarePath.m
k0nserv: Sounds like we should do close the path in the other use case too?

Backpack/Map/Classes/BPKMapAnnotation.m
```diff
@@ -0,0 +1,60 @@
...
+    self = [super init];
+
+    if (self) {
+        [self setup];
+    }
+
+    return self;
+}
+
+- (instancetype)initWithCoordinate:(CLLocationCoordinate2D)coordinate {
```
k0nserv, line 35: Why aren't these inits documented in `BPKMapAnnotation.h`?

Backpack/Map/Classes/BPKMapAnnotationView.m
```diff
@@ -0,0 +1,162 @@
...
+#import <Backpack/Color.h>
+#import <Backpack/Spacing.h>
+
+#import "BPKMapAnnotationViewCalloutView.h"
+#import "BPKMapAnnotation.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface BPKMapAnnotationView()
+@property(nonatomic, nullable) BPKMapAnnotation *bpk_annotation;
```
k0nserv, line 30: This name is a bit surprising, snake case isn't a convention we use outside of category added method names, but I supposes this is similar. 

This should be `readonly` no?

Backpack/Map/Classes/BPKMapAnnotationView.m
```diff
@@ -0,0 +1,162 @@
...
+-(BPKMapAnnotation *_Nullable)bpk_annotation {
+    if(self.annotation == nil) {
+        return nil;
+    }
+
+    if(![self.annotation isKindOfClass:BPKMapAnnotation.class]) {
+        return nil;
+    }
+
+    BPKMapAnnotation *annotation = (BPKMapAnnotation *)self.annotation;
```
k0nserv, line 67: nitpic: This line does nothing that we don't get with

```objective-c
return (BPKMapAnnotation *)self.annotation;
```

Backpack/Map/Classes/BPKMapAnnotationView.m
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    self.calloutView = [[BPKMapAnnotationViewCalloutView alloc] initWithText:self.annotation.title];
+    self.calloutView.enabled = self.enabled;
+    self.calloutView.selected = self.selected;
+    [self addSubview:self.calloutView];
+    self.calloutView.translatesAutoresizingMaskIntoConstraints = NO;
+    [NSLayoutConstraint activateConstraints:@[
+        [self.calloutView.centerXAnchor constraintEqualToAnchor:self.dotView.centerXAnchor],
+        [self.calloutView.bottomAnchor constraintEqualToAnchor:self.dotView.topAnchor constant:2]
```
k0nserv, line 104: This could be `BPKSpacingSm / 2` which is a slightly improvement

Backpack/Map/Classes/BPKMapAnnotationView.m
```diff
@@ -0,0 +1,162 @@
...
+    ]];
+}
+
+-(void)destroyCallout {
+    [self.calloutView removeFromSuperview];
+    self.calloutView = nil;
+}
+
+-(void)updateImage {
+    self.image = nil;
```
k0nserv, line 114: 🤔

Backpack/Map/Classes/BPKMapAnnotationView.m
```diff
@@ -0,0 +1,162 @@
...
+}
+
+- (void)setAnnotation:(id<MKAnnotation> _Nullable)annotation {
+    [super setAnnotation:annotation];
+
+    [self updateCallout];
+}
+
+-(void)updateCallout {
+    self.enabled = self.bpk_annotation.enabled;
```
k0nserv, line 138: This will default to `NO` if `self.bpk_annotation` is `nil` which is correct, but the code is a bit magical since it relies on the default behaviour of sending a message to ` nil`

Backpack/Map/Classes/BPKMapAnnotationView.m
```diff
@@ -0,0 +1,162 @@
...
+
+    BOOL shouldShowCallout = self.selected || self.bpk_annotation.alwaysShowCallout;
+    if(shouldShowCallout && self.calloutView == nil) {
+        [self createCallout];
+    }
+    if(!shouldShowCallout && self.calloutView != nil) {
+        [self destroyCallout];
```
k0nserv, line 145: I feel like it's better to hide just hide the call out in this case? Destroying and re-creating involves a good chunk of view trashing 🤔

Backpack/Map/Classes/BPKMapAnnotationView.m
```diff
@@ -0,0 +1,162 @@
...
+
+    [self updateCallout];
+}
+
+-(void)updateCallout {
+    self.enabled = self.bpk_annotation.enabled;
+
+    BOOL shouldShowCallout = self.selected || self.bpk_annotation.alwaysShowCallout;
+    if(shouldShowCallout && self.calloutView == nil) {
+        [self createCallout];
```
k0nserv, line 142: Maybe we can just use a lazy readonly property for `calloutView`? The logic will be simpler and we'll still benefit from not creating the callout view until needed

Backpack/Map/Classes/BPKMapAnnotationView.m
k0nserv: Yes that way we have a single source of truth, just make sure you make it `weak` so we don't get a retain cycle

Backpack/Map/Classes/BPKMapAnnotation.m
georgegillams: Because they're just overriding the underlying initialisers.
Can document them if that's helpful?

Backpack/Map/Classes/BPKMapAnnotationView.m
georgegillams: The underlying class shows an image of a pin. We can set the image to be nil as we want to use a subview instead of an image.

We are also required to set the image when the InterfaceStyle changes, otherwise it will be reverted by the underlying component.

simulator_screenshot_9DD7B9DA-8628-4BBE-890A-53BBE090A2BD

Backpack/Map/Classes/BPKMapAnnotationView.m
georgegillams: `bpk_annotation` being `nil` is a very very edge case anyway, because it means someone initialised a `BPKMapAnnotationView` with a `MKMapAnnotation`.
Maybe `bpk_annotation` should `NSAssert the type so that devs know they've used it differently to how it's meant to be?

Backpack/FlareView/Classes/BPKFlareView.m
```diff
@@ -141,6 +141,11 @@ - (CGFloat)flareHeight {
 - (void)setCornerRadius:(CGFloat)cornerRadius {
     if (_cornerRadius != cornerRadius) {
         _cornerRadius = cornerRadius;
+
+        self.layer.cornerRadius = cornerRadius;
+        self.backgroundView.layer.cornerRadius = cornerRadius;
+        self.contentView.layer.cornerRadius = cornerRadius;
```
georgegillams, line 147: This is done to ensure the outline is rounded too.

Backpack/Map/Classes/BPKMapAnnotationView.h
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * `BPKMapAnnotationView` is a subclass of `MKPinAnnotationView` which contains the Skyscanner map annotation style.
+ */
+IB_DESIGNABLE @interface BPKMapAnnotationView : MKPinAnnotationView
+
+/**
+ * Whether the annotation view has previously been selected by the user.
+ */
+@property(nonatomic, assign) BOOL hasBeenSelected;
```
k0nserv, line 33: We are not setting an explicit default value for this, can we?

Backpack/Map/Classes/BPKMapAnnotationView.m
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    return _calloutView;
+}
+
+-(void)updateImage {
+    self.image = nil;
+}
+
+- (void)setSelected:(BOOL)selected {
```
k0nserv, line 110: Can this just call `[self setSelected:selected animated:NO]`?

Backpack/Map/Classes/BPKMapAnnotationView.m
```diff
@@ -0,0 +1,158 @@
...
+-(void)updateCallout {
+    self.enabled = self.bpk_annotation.enabled;
+
+    BOOL shouldShowCallout = self.selected || self.bpk_annotation.alwaysShowCallout;
+    if (shouldShowCallout) {
+        self.calloutView.hidden = false;
+    }
+
+    // Using `_calloutView` so that if it is not already created we don't access the lazy property
+    if (!shouldShowCallout && _calloutView != nil) {
```
k0nserv, line 139: Fab

Backpack/Map/Classes/BPKMapAnnotationViewCalloutView.m
```diff
@@ -0,0 +1,144 @@
...
+    self.label.fontStyle = BPKFontStyleTextSmEmphasized;
+    self.backgroundView.backgroundColor = BPKColor.skyBlue;
+    self.contentView.backgroundColor = BPKColor.clear;
+    self.label.textColor = BPKColor.white;
+    self.contentView.layer.borderColor = BPKColor.clear.CGColor;
+}
+
+- (void)setupAppearance {
+    self.contentView.layer.borderWidth = BPKBorderWidthLg;
+    self.layer.masksToBounds = NO;
```
k0nserv, line 122: This defaults to `NO` why are we setting it?

k0nserv (approved):

PR #888, opened 2021-02-09
[BPK-4397] Refactor calendar cell size constants

georgegillams:

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/Backpack/Calendar/Classes/BPKCalendarCellSpacing.h
+++ b/Backpack/Calendar/Classes/BPKCalendarCellSpacing.h
@@ -1,0 +1,32 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2018-2021 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


Backpack/Calendar/Classes/BPKCalendarCellSpacing.h
```diff
@@ -0,0 +1,32 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface BPKCalendarCellSpacing : NSObject
```
k0nserv, line 23: My one concern here is the extra cost of Objective-C method dispatch, but that's probably a premature optimisation. Can we add a commen there stating something to the effect of "If we encounter performance issues with the calendar we should switch these properties to C functions which the compiler could inline"?

k0nserv (approved):
Noice

PR #894, opened 2021-02-17
[BPK-4412] Extend calendar range at edges

georgegillams:

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


Backpack/Calendar/Classes/BPKCalendar.m
```diff
@@ -238,7 +245,16 @@ - (void)layoutSubviews {
     CGFloat calendarWidth = CGRectGetWidth(self.calendarView.frame);
     CGFloat weekdayViewHeight = 6 * BPKSpacingMd;
 
-    self.calendarView.frame = CGRectMake(BPKSpacingBase, weekdayViewHeight, width - 2 * BPKSpacingBase, height - weekdayViewHeight);
+    self.calendarView.translatesAutoresizingMaskIntoConstraints = NO;
+    [NSLayoutConstraint activateConstraints:@[
```
k0nserv, line 249: This is no good for two reasons:

1. We are in `layoutsubviews`, in which we shouldn't be adding constraints
2. We explicitly avoid using Auto Layout in this view because it breaks React Native when we do.

Backpack/Calendar/Classes/BPKCalendarCell.m
```diff
@@ -85,57 +85,65 @@ - (void)layoutSubviews {
 
     self.shapeLayer.frame = CGRectMake(paddingX, 0, BPKCalendarCellSpacing.cellCircleHeight, BPKCalendarCellSpacing.cellCircleHeight);
 
+    CGFloat underflow = 0;
+    CGFloat overflow = 0;
+    CGFloat overflowLength = [[UIScreen mainScreen] bounds].size.width;
```
k0nserv, line 90: This needlessly limits the view to being full screen. I think we need to think of another method for this

k0nserv: For example, we could keep a `weak` reference to the calendar itself and use that instead

Backpack/Calendar/Classes/BPKCalendarCell.m
```diff
@@ -85,57 +85,65 @@ - (void)layoutSubviews {
...
-        }
-        cornerRadii = CGSizeMake(height / 2.0, height / 2.0);
-        break;
-
-    case RowTypeBoth:
-        corners = UIRectCornerAllCorners;
-        cornerRadii = CGSizeMake(height / 2.0, height / 2.0);
+        case RowTypeStart:
+            if (self.selectionType != SelectionTypeLeadingBorder) {
+                underflow = overflowLength;
```
k0nserv, line 94: Isn't this too much? i.e. it should be something like ~16-40pts but this will be more like ~350pt?

Backpack/Calendar/Classes/BPKCalendarCell.m
```diff
@@ -85,57 +85,65 @@ - (void)layoutSubviews {
...
+            }
+            break;
+        case RowTypeEnd:
+            if (self.selectionType != SelectionTypeTrailingBorder) {
+                overflow = overflowLength;
```
k0nserv, line 99: Same question here about this being too much overflow

Backpack/Calendar/Classes/BPKCalendar.m
```diff
@@ -171,7 +172,12 @@ - (instancetype)initWithConfiguration:(BPKCalendarConfiguration *)configuration
 - (void)setup {
     self.gregorian = [[NSCalendar alloc] initWithCalendarIdentifier:NSCalendarIdentifierGregorian];
 
+    self.calendarWrapperView = [UIView new];
     self.calendarView = [[FSCalendar alloc] initWithFrame:CGRectZero];
+    self.calendarView.clipsToBounds = NO;
+    self.calendarView.subviews[0].clipsToBounds = NO;
```
k0nserv, line 178: We should annotate this with a comment like

```objective-c
// HACK: This relies on messing with the internal view hierarchy of `FSCalendar` and is liable to break.
```

Furhter we need to ensure snapshot tests fail if FSCalendar changes in a way that will break our assumptions about the view hiearchy.

Lastly, do we know that the subview orders are stable within FSCalendar? Is there a more stable way do write this logic i.e. by identifying the relevant views based on their class, tag, or something? 

Las

Backpack/Calendar/Classes/BPKCalendarCell.m
georgegillams: I don't think it does? It just ensures that the overflow will be long enough regardless of screen size?

k0nserv: It feels wrong 🤔 If this is the case why not use 9000?

georgegillams: That would work too?
Until apple release a device that's wider than 9000pt

Backpack/Calendar/Classes/BPKCalendar.m
georgegillams: I think the best way would be to iterate through subviews until you get to a specific class. There should be a max depth we go to in case we don't find the target view.
WDYT?

k0nserv: I was thinking something less extreme, maybe just do one level but it means we can handle if the index changes?

Backpack/Calendar/Classes/BPKCalendarCell.m
k0nserv: I feel like it should be the exact value required calculated, but maybe that's just dumb

Backpack/Calendar/Classes/BPKCalendarCell.m
georgegillams: The maximum it could need to be would be `6*cellWidth + bpkSpacingBase`

k0nserv: Right, my thoughts on this are in the above conversation

Backpack/Calendar/Classes/BPKCalendarCell.m
k0nserv: Can we add a comment to overflowLength that explains why that values is used?

Backpack/Calendar/Classes/BPKCalendar.m
```diff
@@ -168,10 +169,23 @@ - (instancetype)initWithConfiguration:(BPKCalendarConfiguration *)configuration
     return [self initWithFrame:CGRectZero];
 }
 
+-(void)applySubviewClipsToBoundsHack {
```
k0nserv, line 172: Top notch 🤩

k0nserv (approved):
Great 👍🏼

PR #895, opened 2021-02-18
Feature/map marker fixes

jookes:
Changed the base class of BPKMapMarkerAnnotationView, when using it in a project the default pin marker would show up occasionally. 

Reset hasBeenSelected flag when the AnnotationView is reused by the system. This prevents a different annotation from showing the 'has visited before' state when a totally new set of markers is put on the map. 

+ [ ] `UNRELEASED.md`
+ [ ] `README.md`
+ [ ] Tests

### DIFF
--- a/Backpack/Map/Classes/BPKMapAnnotationView.m
+++ b/Backpack/Map/Classes/BPKMapAnnotationView.m
@@ -159,5 +159,9 @@ NS_ASSUME_NONNULL_BEGIN
     }
 }
 
+- (void)prepareForReuse {
+    self.hasBeenSelected = false;
+}
+
 @end
 NS_ASSUME_NONNULL_END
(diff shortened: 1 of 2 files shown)


Backpack/Map/Classes/BPKMapAnnotationView.m
```diff
@@ -159,5 +159,9 @@ - (void)traitCollectionDidChange:(UITraitCollection * _Nullable)previousTraitCol
     }
 }
 
+- (void)prepareForReuse {
```
georgegillams, line 162: Shouldn't this call `[super prepareForReuse]`

jookes: It totally should. Just pushed :D

georgegillams (approved):
LGTM - great catches!

PR #899, opened 2021-02-22
[BPK-4417] Migrate chip to new navigation story format

shaundon:
I've added a new `Story` protocol and `ChipStory` which implements it. The protocol ensures that `presentableStory: Presentable` is implemented, and it also adds adherence to `CaseIterable` and a `storyStructure` property that returns `[Item]`.

This means that the code to add a story can be very simple and all the relevant things – like naming the stories – are abstracted out of `NavigationData`.

The only thing I couldn't figure out was how to make `Story` adhere to `String` so that we don't need to manually add it to each instance.

### DIFF
--- a/Example/Backpack/NavigationData.swift
+++ b/Example/Backpack/NavigationData.swift
@@ -16,11 +16,13 @@
  * limitations under the License.
  */
 
+// swiftlint:disable type_body_length
 class NavigationData: NSObject {
     static var mainStoryboard = loadStoryboard(name: "Main")
     static var buttonsStoryboard = loadStoryboard(name: "Buttons")
     static var calendarStoryboard = loadStoryboard(name: "Calendar")
     static var cardStoryboard = loadStoryboard(name: "Cards")
+    static var chipStoryboard = loadStoryboard(name: "Chips")
 
     static func setButtonStyle(style: BPKButtonStyle) -> (UIViewController) -> Void {
         return {storyboardVC in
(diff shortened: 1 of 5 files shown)


Example/Backpack/NavigationData.swift
```diff
@@ -236,14 +237,20 @@ class NavigationData: NSObject {
                 Item(name: "With divider without padding", value: .story(dividedCardStoryboard(story: .dividedVerticalNoPadding)))
                 Item(name: "With divider and corner style large", value: .story(dividedCardStoryboard(story: .dividedHorizontalCornerStyleLarge)))
             }
+            Group(name: "Chips") {
+              Item(name: "Default", value: .story(ChipStory.default.presentableStory))
+              Item(name: "With icons", value: .story(ChipStory.withIcons.presentableStory))
+              Item(name: "With background colour", value: .story(ChipStory.withBackgroundColor.presentableStory))
+              Item(name: "Filled", value: .story(ChipStory.filled.presentableStory))
+              Item(name: "Filled with background colour", value: .story(ChipStory.filledWithBackgroundColor.presentableStory))
+            }
```
shaundon, line 246: I wasn't 100% on how to do it, but we could also make this part of the `Story` protocol so that this logic could remain in `ChipStory.swift`. Then here we could just do:

```swift
ChipStory.storyGroup
```

..or similar.

Example/Backpack/Stories/ChipStory.swift
```diff
@@ -0,0 +1,49 @@
+//
```
georgegillams, line 1: extraneous line

georgegillams (commented):
Couple of extraneous lines at the top of the licenses

Example/Backpack/Stories/ChipStory.swift
shaundon: I'm so used to outsourcing my brain to eslint 😂

Example/Backpack/NavigationData.swift
```diff
@@ -21,6 +21,7 @@ class NavigationData: NSObject {
     static var buttonsStoryboard = loadStoryboard(name: "Buttons")
     static var calendarStoryboard = loadStoryboard(name: "Calendar")
     static var cardStoryboard = loadStoryboard(name: "Cards")
+    static var chipStoryboard = loadStoryboard(name: "Chips")
```
georgegillams, line 24: Do we still need this now?

shaundon: Kind of, but maybe we should move it. If you look at line 29 of `ChipStory` we reference this property, but maybe we should move it into `ChipStory` so it can be more self contained?

shaundon: Just tried that and worked fine, will update the PR.

k0nserv: Yeah and possibly even remove the storyboard? We can just move it to Main?

shaundon: Is there a benefit to having it in a separate storyboard? I personally found it quite useful to find `Chips.storyboard` in the sidebar and go there as opposed to hunting through the main one

georgegillams (approved):

Example/Backpack/NavigationData.swift
georgegillams: At the moment we have a mixture. At some point it would be nice to put them all in one, or make them all separate, but that's out of scope for this change IMO

PR #903, opened 2021-02-23
[No Jira] Refactor Story protocol to work better with single stories

shaundon:
PoC where I migrate a single story – colours – to this new format.

### DIFF
--- a/Example/Backpack/NavigationData.swift
+++ b/Example/Backpack/NavigationData.swift
@@ -23,7 +23,7 @@ class NavigationData: NSObject {
 
     static var appStructure: [Item] = makeApp {
         Group(name: "Tokens") {
-            Item(name: "Colors", value: .story(mainStoryboard("ColorsViewController")))
+            ColorsStory().example
             Group(name: "Gradients") {
                 Item(
                     name: "Primary",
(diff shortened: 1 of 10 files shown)


Example/Backpack/Story.swift
```diff
@@ -18,14 +18,13 @@
...
 }
 
-extension Story where Self.RawValue == String {
-  static var storyStructure: [Item] {
-    return self.allCases.map({ story in
-      return Item(name: story.rawValue, value: .story(story.presentableStory))
-    })
+extension Story {
+  var example: Item {
+    Item(name: self.title, value: .story(self.presentableStory))
```
shaundon, line 28: `Story` is now the base protocol for use with a single story item, e.g. the colours story.

Example/Backpack/StoryGroup.swift
```diff
@@ -0,0 +1,27 @@
...
+
+import Foundation
+
+protocol StoryGroup: CaseIterable, Story {}
+
+extension StoryGroup {
+  static var allExamples: [Item] {
+    self.allCases.map({ $0.example })
+  }
+}
```
shaundon, line 27: `StoryGroup` extends story and is intended to be used with enums for when there are multiple examples in a story.

Example/Backpack/NavigationData.swift
```diff
@@ -43,20 +43,20 @@ class NavigationData: NSObject {
             Item(name: "Bar charts", value: .story(loadStoryboard(name: "BarCharts", identifier: "BarChartsViewController")))
             Item(name: "Bottom sheet", value: .story(loadStoryboard(name: "BottomSheet", identifier: "BottomSheetViewController")))
             Group(name: "Buttons") {
-              ButtonStory.storyStructure
+              ButtonStory.allExamples
```
shaundon, line 46: I felt that 'examples' was better syntax here as otherwise it gets a bit confusing if we keep saying 'stories'. Like with web Storybook, each story can have one or more examples contained within.

georgegillams (approved):

PR #908, opened 2021-02-26
[BPK-3546] Refactor calendar selection config

georgegillams:
- This PR doesn't change any behaviour.
 - The API of the calendar now requites a `selectionConfiguration` instance instead of a `selectionType` enum.
 - A follow-up PR will make another breaking change to enforce providing the additional a11y strings required to make this more accessible. 

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/Backpack/Calendar/Classes/BPKCalendarSelectionConfiguration.h
+++ b/Backpack/Calendar/Classes/BPKCalendarSelectionConfiguration.h
@@ -1,0 +1,46 @@
+/*
+ * Backpack - Skyscanner's Design System
+ *
+ * Copyright 2018-2021 Skyscanner Ltd
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface BPKCalendarSelectionConfiguration : NSObject
+
+/**
+ * Creates a `BPKCalendarSelectionConfiguration` with the specific classes.
+ *
+ * @param rangeStyleSelection Whether the selection configuration is a range style.
+ * @param allowsMultipleSelection Whether the selection configuration allows multiple simultaneous selections.
+ * @return `BPKCalendarSelectionConfiguration` instance.
+ */
+- (instancetype)initWithRangeStyleSelection:(BOOL)rangeStyleSelection allowMultipleSelection:(BOOL)allowsMultipleSelection;
+
+/**
+ * Whether the selection configuration is a range style (ie start date and end date with 0 or more dates between).
+ */
+@property(readonly)BOOL isRangeStyleSelection;
+
+/**
+ * Whether the selection configuration allows multiple dates to be selected simultaneously.
+ */
+@property(readonly)BOOL allowsMultipleSelection;
+
+-(BOOL)shouldClearSelectedDates:(NSArray<NSDate *> *)selectedDates whenSelectingDate:(NSDate *)date;
+
+@end
+
+NS_ASSUME_NONNULL_END
--- a/Backpack/Calendar/Classes/BPKCalendarSelectionConfiguration.h
+++ b/Backpack/Calendar/Classes/BPKCalendarSelectionConfiguration.h
@@ -16,6 +16,26 @@
  * limitations under the License.
  */
 
+/**
+ * Enum values for specifying calendar selection style
+ */
+typedef NS_ENUM(NSUInteger, BPKCalendarSelectionStyle) {
+    /**
+     * Select a single date in the calendar.
+     */
+    BPKCalendarSelectionStyleSingle = 0,
+
+    /**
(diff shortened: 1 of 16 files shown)


Backpack/Calendar/Classes/BPKCalendarSelectionConfiguration.h
```diff
@@ -0,0 +1,46 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface BPKCalendarSelectionConfiguration : NSObject
```
k0nserv, line 21: This needs to be documented otherwise the whole class will not show up in docs

Backpack/Calendar/Classes/BPKCalendarSelectionConfiguration.h
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Whether the selection configuration is a range style (ie start date and end date with 0 or more dates between).
+ */
+@property(readonly)BOOL isRangeStyleSelection;
+
+/**
+ * Whether the selection configuration allows multiple dates to be selected simultaneously.
+ */
+@property(readonly)BOOL allowsMultipleSelection;
```
k0nserv, line 40: What happens if I set both this and `isRangeStyleSelection` to `YES`?

Backpack/Calendar/Classes/BPKCalendarSelectionConfigurationSingle.m
```diff
@@ -0,0 +1,31 @@
...
+#import "BPKCalendarSelectionConfigurationSingle.h"
+
+@implementation BPKCalendarSelectionConfigurationSingle
+
+- (instancetype)init {
+    return [super initWithRangeStyleSelection:NO allowMultipleSelection:NO];
+}
+
+- (BOOL)shouldClearSelectedDates:(NSArray<NSDate *> *)selectedDates whenSelectingDate:(NSDate *)date {
+    return NO;
```
k0nserv, line 28: Shouldn't this be `return YES`?

Example/Tests/BPKCalendarSelectionTests.m
```diff
@@ -0,0 +1,100 @@
...
+@interface BPKCalendarSelectionTests : XCTestCase
+
+@property NSDate *date1;
+@property NSDate *date2;
+@property NSDate *date3;
+
+@end
+
+NS_ASSUME_NONNULL_BEGIN
+@implementation BPKCalendarSelectionTests
```
k0nserv, line 32: It's nice that we can now unit test some of this logic 🙌

Backpack/Calendar/Classes/BPKCalendarSelectionConfiguration.h
```diff
@@ -0,0 +1,46 @@
...
+ * @param rangeStyleSelection Whether the selection configuration is a range style.
+ * @param allowsMultipleSelection Whether the selection configuration allows multiple simultaneous selections.
+ * @return `BPKCalendarSelectionConfiguration` instance.
+ */
+- (instancetype)initWithRangeStyleSelection:(BOOL)rangeStyleSelection allowMultipleSelection:(BOOL)allowsMultipleSelection;
+
+/**
+ * Whether the selection configuration is a range style (ie start date and end date with 0 or more dates between).
+ */
+@property(readonly)BOOL isRangeStyleSelection;
```
k0nserv, line 35: These getters need to be implemented otherwise the compiler will generate storage for these properties redundantly.

k0nserv (commented):
We also need to add doc comments to the different configurations for them to show up in docs

Backpack/Calendar/Classes/BPKCalendarSelectionConfiguration.h
georgegillams: That would be valid, as range selection is `rangeStyle` and allows multiple (up to 2) selections.
I guess the inconsistency would be setting `isRangeStyleSelection = YES` and `allowsMultipleSelection = NO`.

georgegillams: I'm not sure - most consumers shouldn't be worried about these properties anyway as they'll use one of the 3 preconfigured subclasses. And maybe in the future we will want to support something else, such as several ranges?

Backpack/Calendar/Classes/BPKCalendarSelectionConfiguration.h
georgegillams: What do you mean "redundantly"? We want these properties to exist?

Backpack/Calendar/Classes/BPKCalendarSelectionConfiguration.h
k0nserv: We could still do that with something explicit like with `BPKCalendarSelectionStyleMultiPleRanges` ?

Backpack/Calendar/Classes/BPKCalendarSelectionConfiguration.h
k0nserv: Yes, but we don't want the compiler to generate ivars for them? If we don't implement the getters we'll have `_isRangeSelection` which is never used.

georgegillams: We use those in the constructor 🤔

k0nserv: Hmm I suppose I am just confused then 😓

k0nserv (approved):
LGTM

PR #909, opened 2021-02-27
[BPK-3546] Make price labels available to AT

georgegillams:

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/Backpack/Calendar/Classes/BPKCalendar.m
+++ b/Backpack/Calendar/Classes/BPKCalendar.m
@@ -630,7 +630,7 @@ CGFloat const BPKCalendarDefaultCellHeight = 44;
 
         calendarCell.selectionType = selectionType;
         calendarCell.rowType = rowType;
-        calendarCell.accessibilityLabel = [self formattedDate:date];
+        calendarCell.accessibilityLabel = [calendarCell defaultAccessibilityLabelForDate:date locale:self.locale];
 
         if ([self isDateEnabled:date]) {
             calendarCell.isAccessibilityElement = YES;
@@ -690,13 +690,6 @@ CGFloat const BPKCalendarDefaultCellHeight = 44;
 
     // Gonna return true, because in the words of Sia, I'm still here...
     return true;
-}
-
-- (NSString *)formattedDate:(NSDate *)date {
-    NSDateFormatter *dateFormatter = [NSDateFormatter new];
-    dateFormatter.locale = self.locale;
-    dateFormatter.dateStyle = NSDateFormatterLongStyle;
-    return [dateFormatter stringFromDate:date];
 }
 
 - (BOOL)isDateInToday:(NSDate *)date {
--- a/Backpack/Calendar/Classes/BPKCalendar.m
+++ b/Backpack/Calendar/Classes/BPKCalendar.m
@@ -92,6 +92,7 @@ NS_ASSUME_NONNULL_BEGIN
 @property(nonatomic, strong, nonnull) BPKCalendarAppearance *appearance;
 @property(nonatomic, strong, nonnull) UIView *bottomBorder;
 @property(nonatomic, strong, nonnull) NSCalendar *gregorian;
+@property(nonatomic, strong, nonnull) NSDateFormatter *dateFormatter;
 
 @property BOOL sameDayRange;
 
@@ -178,6 +179,7 @@ CGFloat const BPKCalendarDefaultCellHeight = 44;
 
 - (void)setup {
     self.gregorian = [[NSCalendar alloc] initWithCalendarIdentifier:NSCalendarIdentifierGregorian];
+    self.dateFormatter = [NSDateFormatter new];
 
     // We use a wrapper view here to prevent the calendar cells being rendered over the week days
     self.calendarWrapperView = [UIView new];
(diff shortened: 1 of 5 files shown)


Backpack/Calendar/Classes/BPKCalendarPriceLabelCell.m
```diff
@@ -80,9 +82,26 @@ - (void)prepareForReuse {
...
+}
+
+-(void)updateAccessibilityLabel {
+    NSString *accessibilityLabel = self.consumerAccessibilityLabel;
+
+    if (self.priceLabelCellData.price != nil) {
+        accessibilityLabel = [NSString stringWithFormat:@"%@ %@", self.consumerAccessibilityLabel, self.priceLabelCellData.price];
+    }
+
+    [super setAccessibilityLabel:accessibilityLabel];
```
georgegillams, line 104: I guess we could override the `accessibilityLabel` getter instead? Maybe that would be cleaner?

georgegillams: Actually I'm pretty sure it's an improvement so I've just gone ahead and done it 👍

Backpack/Calendar/Classes/BPKCalendarPriceLabelCell.m
```diff
@@ -80,9 +82,24 @@ - (void)prepareForReuse {
...
 
+- (void)setAccessibilityLabel:(NSString *)accessibilityLabel {
+    self.consumerAccessibilityLabel = accessibilityLabel;
+}
+
+- (NSString *)accessibilityLabel {
+    NSString *accessibilityLabel = self.consumerAccessibilityLabel;
+
+    if (self.priceLabelCellData.price != nil) {
+        accessibilityLabel = [NSString stringWithFormat:@"%@ %@", self.consumerAccessibilityLabel, self.priceLabelCellData.price];
```
k0nserv, line 99: I'm not sure about this. AFAIK naive concatenation of strings doesn't work well in all languages and the ordering/structure of sentence is a localisation concern.

georgegillams: I guess we need to provide a delegate method so that consumers can provide a string, but maybe this is a good default?

k0nserv: Yeah that's probably a good idea. I suppose this might be a good first step yes

k0nserv: Although, what if the consumers have already localised the `accessibilityLabel` with the price?

georgegillams: I guess we need to configure this to only append the price if the consumer isn't setting a custom value?

Backpack/Calendar/Classes/BPKCalendarCell.m
```diff
@@ -217,4 +217,15 @@ - (void)configureWithData:(id)data {
     NSAssert(false, @"BPKCalendarCell configureWithData: should be overridden. The base class method should never be called.");
 }
 
+- (NSString *)formattedDate:(NSDate *)date locale:(NSLocale *)locale {
+    NSDateFormatter *dateFormatter = [NSDateFormatter new];
```
k0nserv, line 221: Formatters are, supposedly, expensive to create. I haven't looked at this in a long time, but worth doing some research before we decided to create them in a hot path like this.

georgegillams: This is actually no different to how the code worked before, but I'm sure there's some optimisation that can be done 👍

georgegillams: I've optimised this by making the calendar provide it's own formatter

Backpack/Calendar/Classes/BPKCalendarCell.m
```diff
@@ -217,4 +217,14 @@ - (void)configureWithData:(id)data {
     NSAssert(false, @"BPKCalendarCell configureWithData: should be overridden. The base class method should never be called.");
 }
 
+- (NSString *)formattedDate:(NSDate *)date locale:(NSLocale *)locale formatter:(NSDateFormatter *)formatter {
+    formatter.locale = locale;
+    formatter.dateStyle = NSDateFormatterLongStyle;
```
k0nserv, line 222: Shouldn't this also be the responsibility of the calendar not the cell?

georgegillams: of course

Backpack/Calendar/Classes/BPKCalendar.m
```diff
@@ -178,6 +179,9 @@ -(void)applySubviewClipsToBoundsHack {
 
 - (void)setup {
     self.gregorian = [[NSCalendar alloc] initWithCalendarIdentifier:NSCalendarIdentifierGregorian];
+    self.dateFormatter = [NSDateFormatter new];
+    self.dateFormatter.locale = locale;
```
k0nserv, line 183: This doesn't compile. We need to use `self.locale`. We also need to update this property in `setLocale:`

k0nserv (approved):

PR #919, opened 2021-03-17
Bpk 4426 add calendar accessibility strings

georgegillams:

+ [ ] `UNRELEASED.md`
+ [ ] `README.md`
+ [ ] Tests

### DIFF
--- a/Backpack/Calendar/Classes/BPKCalendarSelectionConfiguration.h
+++ b/Backpack/Calendar/Classes/BPKCalendarSelectionConfiguration.h
@@ -68,6 +68,33 @@ NS_ASSUME_NONNULL_BEGIN
 
 -(BOOL)shouldClearSelectedDates:(NSArray<NSDate *> *)selectedDates whenSelectingDate:(NSDate *)date;
 
+/**
+ * Provides the accessibility hint that should be applied to a cell.
+ *
+ * @param date The date of the cell.
+ * @param selectedDates The dates that are already selected on the calendar.
+ * @returns The appropriate accessibility hint (or nil if no hint is required).
+ */
+- (NSString *_Nullable)accessibilityHintForDate:(NSDate *)date selectedDates:(NSArray<NSDate *> *)selectedDates;
+
+/**
+ * Provides the accessibility label that should be applied to a cell.
+ *
+ * @param date The date of the cell.
+ * @param selectedDates The dates that are already selected on the calendar.
+ * @param baseLabel The value of the cell, which should be included in the new accessibility label.
+ * @returns The appropriate accessibility label.
+ */
+- (NSString *_Nullable)accessibilityLabelForDate:(NSDate *)date selectedDates:(NSArray<NSDate *> *)selectedDates baseLabel:(NSString *)baseLabel;
+
+/**
+ * Provides the accessibility instruction that should be provided following selection of a date..
+ *
+ * @param selectedDates The dates that have been selected on the calendar.
+ * @returns The appropriate accessibility instruction to be announced (or nil if no instruction is required).
+ */
+- (NSString *_Nullable)accessibilityInstructionHavingSelectedDates:(NSArray<NSDate *> *)selectedDates;
+
 @end
 
 NS_ASSUME_NONNULL_END
--- a/Backpack/Calendar/Classes/BPKCalendarSelectionConfigurationMultiple.h
+++ b/Backpack/Calendar/Classes/BPKCalendarSelectionConfigurationMultiple.h
@@ -20,13 +20,32 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+/**
+ * `BPKCalendarSelectionConfigurationMultiple` is a subclass of `BPKCalendarSelectionConfiguration` designed to enable a user to select any number of individual dates.
+ */
 @interface BPKCalendarSelectionConfigurationMultiple : BPKCalendarSelectionConfiguration
 
-@property(readonly)NSString *selectionHint;
-@property(readonly)NSString *deselectionHint;
+/**
+ * The hint provided to assistive technologies informing a user how to select a date.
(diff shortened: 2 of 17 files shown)


Example/Backpack/ViewControllers/CalendarViewController.swift
```diff
@@ -37,13 +37,33 @@ class CalendarViewController: UIViewController, BPKCalendarDelegate {
...
+        secondSelectionHint: "Double tap to select return date",
+        firstSelectionState: "Selected as departure date",
+        secondSelectionState: "Selected as return date",
+        betweenSelectionState: "Between departure and return date",
+        firstAndSecondSelectionState: "Selected as both departure and return date",
+        returnDatePrompt: "Now please select a return date"
+    )
+
     override func viewDidLoad() {
+        // I guess it shouldn't be possible to create a calendar without a selection config 🤔
```
georgegillams, line 58: Should we add a constructor argument to `BPKCalendar` for an instance of `BPKCalendarSelectionConguration`?

Backpack/Calendar/Classes/BPKCalendarSelectionConfiguration.h
```diff
@@ -68,6 +68,33 @@ NS_ASSUME_NONNULL_BEGIN
...
+ */
+- (NSString *_Nullable)accessibilityLabelForDate:(NSDate *)date selectedDates:(NSArray<NSDate *> *)selectedDates baseLabel:(NSString *)baseLabel;
+
+/**
+ * Provides the accessibility instruction that should be provided following selection of a date..
+ *
+ * @param selectedDates The dates that have been selected on the calendar.
+ * @returns The appropriate accessibility instruction to be announced (or nil if no instruction is required).
+ */
+- (NSString *_Nullable)accessibilityInstructionHavingSelectedDates:(NSArray<NSDate *> *)selectedDates;
```
k0nserv, line 96: Should these method use `NSOrderedSet` instead of `NSarray`? I think `NSDate` implements both `hash` and `isEqual:`

Backpack/Calendar/Classes/BPKCalendarSelectionConfigurationRange.h
```diff
@@ -22,6 +22,18 @@ NS_ASSUME_NONNULL_BEGIN
 
 @interface BPKCalendarSelectionConfigurationRange : BPKCalendarSelectionConfiguration
 
+@property(readonly)NSString *firstSelectionHint;
```
k0nserv, line 25: These should all be `copy` since they are `NSString`. Also `nonatomic`

Backpack/Calendar/Classes/BPKCalendarSelectionConfigurationRange.m
```diff
@@ -20,8 +20,21 @@
...
 @implementation BPKCalendarSelectionConfigurationRange
 
-- (instancetype)init {
-    return [super initWithSelectionStyle:BPKCalendarSelectionStyleRange];
+
+- (instancetype)initWithFirstSelectionHint:(NSString *)firstSelectionHint secondSelectionHint:(NSString *)secondSelectionHint firstSelectionState:(NSString *)firstSelectionState secondSelectionState:(NSString *)secondSelectionState betweenSelectionState:(NSString *)betweenSelectionState firstAndSecondSelectionState:(NSString *)firstAndSecondSelectionState returnDatePrompt:(NSString *)returnDatePrompt {
+    self = [super initWithSelectionStyle:BPKCalendarSelectionStyleRange];
+
+    if (self) {
+        _firstSelectionHint = firstSelectionHint;
```
k0nserv, line 28: These should copy the provided string

Backpack/Calendar/Classes/BPKCalendarSelectionConfigurationRange.h
```diff
@@ -22,6 +22,18 @@ NS_ASSUME_NONNULL_BEGIN
 
 @interface BPKCalendarSelectionConfigurationRange : BPKCalendarSelectionConfiguration
```
k0nserv, line 23: None of the selection configurations are documented which means they'll not show up on the docs site

Example/Backpack/ViewControllers/CalendarViewController.swift
k0nserv: Yes that seems reasonable, we can no longer create an instance that acts as a default. It will complicated the RN code(it will have to hardcode an english string I think).

k0nserv (commented):
We should be abel to add tests for logic for the various selection configurations and their methods. 

Not everything is documented, we need this to ensure the symbols are available on the docs site. 

Should we write a migration guide, like I did for RN?

Backpack/Calendar/Classes/BPKCalendarSelectionConfigurationMultiple.h
```diff
@@ -20,8 +20,34 @@
...
+/**
+ * `BPKCalendarSelectionConfigurationMultiple` is a subclass of `BPKCalendarSelectionConfiguration` designed to enable a user to select any number of individual dates.
+ */
 @interface BPKCalendarSelectionConfigurationMultiple : BPKCalendarSelectionConfiguration
 
+/**
+ * The hint provided to assistive technologies informing a user how to select a date.
+ * This hint will be provided on any cell that is not selected.
+ */
+@property(readonly, nonatomic, copy)NSString *selectionHint;
```
k0nserv, line 32: nitpic: We usually include a space after `)`

k0nserv (commented):
Looking great. My only nitpic would be to wrap some of the very long lines, e.g. when constructing range selections, with explicit newlines.

k0nserv (commented):
Looks good, what about the migration guide?

k0nserv (approved):
LGTM

PR #929, opened 2021-03-30
[BPK-4405] Map annotation shadows

georgegillams:
Previously the callout view was a subclass of FlareView. As FlareViews clip content to the shape of the flare, a shadow cannot be applied to them.
This PR refactors the callout view to have a FlareView subview. This means that the callout view itself can render the shadow, whilst only the flare-view subview clips its content.

image

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


Backpack/FlareView/Classes/FlareView.h
```diff
@@ -19,5 +19,6 @@
 #define __BACKPACK_MASK_LAYER__
 
 #import "BPKFlareView.h"
+#import "BPKFlarePath.h"
```
georgegillams, line 22: Exposing this so that the path can be consumed by other files.

Backpack/Map/Classes/BPKMapAnnotationViewCalloutView.m
```diff
@@ -99,43 +96,46 @@ - (void)setContentColor:(UIColor *)contentColor {
...
+        backgroundColor = BPKColor.skyBlueTint03;
+        contentBackgroundColor = BPKColor.clear;
+        contentColor = BPKColor.skyBlue;
+        borderColor = BPKColor.clear;
+    } else {
+        fontStyle = BPKFontStyleTextSmEmphasized;
+        backgroundColor = BPKColor.skyBlue;
+        contentBackgroundColor = BPKColor.clear;
+        contentColor = BPKColor.white;
+        borderColor = BPKColor.clear;
```
georgegillams, line 128: Refactored to store the values in variables and then apply them all once at the end of the method.

This is more readable as it avoids us having to return in each case.

Backpack/Map/Classes/BPKMapAnnotationViewCalloutView.h
```diff
@@ -15,19 +15,18 @@
...
 
-#import <Backpack/FlareView.h>
-
 NS_ASSUME_NONNULL_BEGIN
 
 @class BPKMapAnnotationView;
 
 /**
- * `BPKMapAnnotationViewCalloutView` is a subclass of `BPKFlareView` which contains the Skyscanner map callout style.
+ * `BPKMapAnnotationViewCalloutView` dispalys a subclass of `BPKFlareView`. It defines the Skyscanner map callout style.
```
k0nserv, line 27:
```suggestion
 * `BPKMapAnnotationViewCalloutView` displays a subclass of `BPKFlareView`. It defines the Skyscanner map callout style.
```

k0nserv (approved):
Awesome job 🙌

Did you try the performance story on device with a navie shadow implementation for comparison? It'd be interesting to prove that it actually is required and improves performance

k0nserv:
Actually we need the shadow path regardless right? UIKit cannot generate the correct shadow without it so there is no "naive" approach to be had here. 

Sounds like we are good to go wiht this approach

PR #941, opened 2021-04-12
[KOA-4442] Fix blinking on reloadData

georgegillams:
Calling `reoadData` on the FSCalendar causes flashing/blinking as found by other consumers:
https://github.com/WenchaoD/FSCalendar/issues/1012
https://github.com/WenchaoD/FSCalendar/issues/1047

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/Backpack/Calendar/Classes/BPKCalendar.m
+++ b/Backpack/Calendar/Classes/BPKCalendar.m
@@ -375,7 +375,7 @@ CGFloat const BPKCalendarDefaultCellHeight = 44;
 }
 
 - (void)reloadData {
-    [self.calendarView reloadData];
+    [self invalidateVisibleCells];
 }
 
 - (void)refreshDateAppearance {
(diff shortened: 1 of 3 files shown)


Backpack/Calendar/Classes/BPKCalendar.m
```diff
@@ -375,7 +375,7 @@ - (void)didMoveToSuperview {
 }
 
 - (void)reloadData {
-    [self.calendarView reloadData];
+    [self invalidateVisibleCells];
```
georgegillams, line 378: I'm in two minds about this.

On one hand, it seems wrong to not call `reloadData` on the calendar because that's what the exposed method is called. Consumers might be surprised by this.

On the other hand, this achieves the same thing as `reloadData` so consumers would already know how to use it and wouldn't expect the method to be called something else.

k0nserv: If it achieves the same thing, why does it fix the issue?

k0nserv (approved):
I'm not stoked about this, it feels like `reloadData` will be doing exactly what our new method is doing but if it fixes the issue 🤷🏼

georgegillams:
> I'm not stoked about this, it feels like `reloadData` will be doing exactly what our new method is doing but if it fixes the issue 🤷🏼

Let's find out if it does

PR #969, opened 2021-05-12
[NO-JIRA] Add version to example app settings

georgegillams:
Simulator Screen Shot - iPhone 8 - 2021-05-12 at 16 23 28


Example/Backpack/ViewControllers/SettingsViewController.swift
```diff
@@ -91,10 +95,14 @@ class SettingsViewController: UITableViewController {
     // MARK: Table View
 
     override func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
+        if indexPath.section != 2 {
+            return
+        }
+
```
georgegillams, line 101: Previously clicking any section would trigger changing theme 🙈

olliecurtis (approved):

PR #977, opened 2021-06-03
[KOA-4578]: Adopting new Backpack foundations packages

olliecurtis:
This PR adopts the new foundations packages and all the required changes to use the correct packages, where necessary.

+ [x] `UNRELEASED.md`
+ [ ] `README.md`
+ [ ] Tests


k0nserv:
This looks good to me. You need to re-generate the icon snapshot tests to make CI pass. I can either do this for you or walk you through how to do it?

UNRELEASED.md
```diff
@@ -2,6 +2,12 @@
 
 > Place your changes below this line.
 
+**Changed:**
```
k0nserv, line 5: This is non-breaking to consumers, right?

`Changed` usually indicates a breaking change. We need to make sure we don't release this as breaking as a result

olliecurtis: no this change is not breaking. My apologies, I thought changed is minor and `Breaking` is breaking?

k0nserv: We've historically used "Changed" to mean breaking i.e. something has changed and the user will have to modify how they are using Backpack in response. At least that's how I remember things

k0nserv (approved):

PR #980, opened 2021-06-08
[NO JIRA]: Fix contributing guide for iOS versions

olliecurtis:
The purpose of this PR is to update the CONTRIBUTING.md to align the versions for testing with the CI version currently being used.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -45,7 +45,7 @@ Once the fonts are downloaded, add them to the Xcode project under _Example for 
 
 ## Testing
 
-Tests can be run as usual from Xcode(Product -> Test or cmd+U). Snapshot tests should be run on an iPhone 8 running iOS 13.5 to match what is used on CI.
+Tests can be run as usual from Xcode(Product -> Test or cmd+U). Snapshot tests should be run on an iPhone 8 running iOS 14.4 to match what is used on CI.
 
 ## Taking screenshots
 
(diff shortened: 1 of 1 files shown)


k0nserv (approved):
Ace 👍🏼

CONTRIBUTING.md
```diff
@@ -45,7 +45,7 @@ Once the fonts are downloaded, add them to the Xcode project under _Example for
 
 ## Testing
 
-Tests can be run as usual from Xcode(Product -> Test or cmd+U). Snapshot tests should be run on an iPhone 8 running iOS 13.5 to match what is used on CI.
+Tests can be run as usual from Xcode(Product -> Test or cmd+U). Snapshot tests should be run on an iPhone 8 running iOS 14.4 to match what is used on CI.
```
georgegillams, line 48: To avoid this in the future, what if we say
```suggestion
Tests can be run as usual from Xcode(Product -> Test or cmd+U). Snapshot tests should be run on the [device specified for CI](https://github.com/Skyscanner/backpack-ios/blob/3b23a51b85a9dc778599c4eab21a8b8b44903cb8/.github/workflows/ci.yml#L132).
```

georgegillams: Or maybe we should just put a comment somewhere to say if you're changing it, we need to update CONTRIBUTING.md too?

k0nserv: I think the link might be okay but it should link to the file on main rather than a specific commit

olliecurtis: Yep indeed! I did notice that one!

k0nserv (approved):

PR #983, opened 2021-06-15
[KOA-4588] Add XL icon tests and examples

georgegillams:
+ [x] `UNRELEASED.md`
+ NA `README.md`
+ [x] Tests


Backpack/Icon/Classes/BPKXlIconView.m
```diff
@@ -25,6 +25,17 @@ - (instancetype)initWithIconName:(nullable BPKXlIconName)iconName {
     return [super initWithIconName:iconName size:BPKIconSizeXLarge];
 }
 
+- (instancetype)initWithFrame:(CGRect)frame {
+    self = [super initWithFrame:frame];
+    if (self) {
```
georgegillams, line 30: This was missing from the XLIconView, but was present in the other 2 classes. As a result the icons were being rendered too small in the example app.

olliecurtis (approved):

Example/Backpack/Views/CollectionView Cells/IconsPreviewCollectionViewCell.swift
```diff
@@ -79,9 +90,7 @@ class IconsPreviewCollectionViewCell<T: IconDetails>: UICollectionViewCell {
 
         NSLayoutConstraint.activate([
             imageView.centerXAnchor.constraint(equalTo: contentView.centerXAnchor),
-
-            imageView.topAnchor.constraint(greaterThanOrEqualTo: contentView.topAnchor, constant: BPKSpacingSm),
-            contentView.bottomAnchor.constraint(greaterThanOrEqualTo: imageView.bottomAnchor, constant: BPKSpacingSm)
```
georgegillams, line 84: These were just centering on the Y axis, so I made it more readable.

Backpack/Icon/Classes/BPKXlIconView.m
shaundon: Maybe it should be a protocol to enforce that it exists? Is that even a thing in Obj-C or is it just Swift?

UNRELEASED.md
```diff
@@ -3,8 +3,12 @@
 > Place your changes below this line.
 
 **Fixed:**
+
 - Migrated to new Backpack Foundations packages, `@skyscanner/bpk-ios` and `@skyscanner/bpk-svgs`.
 
+- Backpack/Icon
+ - Minor changes to `BPKIcon` which should not affect how the are used or work.
```
shaundon, line 10:
```suggestion
 - Minor changes to `BPKIcon` which should not affect how they are used or work.
```

georgegillams:
@shaundon 
> Maybe it should be a protocol to enforce that it exists? Is that even a thing in Obj-C or is it just Swift?

I'm not sure if that'd help as `initWithFrame` is defined by the superclass. The thing we need to ensure is that we've overridden it and set the `size` property correctly.

I think the tests I've added to verify it works as expected are sufficient. Might be more we can do though?

shaundon (approved):

PR #987, opened 2021-06-22
[KOA-4519] Add fonts pod

georgegillams:
| VPN connection | Pod already installed | Font env | Behaviour |
| --------- | --------- | -------- | -------- |
| 🤷 | Pod not installed | 0 | Font's won't be fetched. System fonts will be used instead |
| Not on VPN | Pod not installed | 1 | Will fail to get fonts. `pod install` will error |
| On VPN | Pod not installed | 1 | Fetching Skyscanner fonts will succeed. Skyscanner fonts will be used |
| 🤷 | Pod already installed with fonts | 🤷 | No change |
| 🤷 | Pod already installed without fonts | 🤷 | No change |

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


.gitignore
```diff
@@ -5,6 +5,7 @@
 build/
 bpk-fonts-tmp/
 Example/Backpack/Resources/SkyscannerRelative*
+Fonts/Assets/SkyscannerRelative*
```
georgegillams, line 8: Ignoring the new path that fonts are stored in. I'll leave the old path for now, incase anyone still has them there

Fonts/Fonts.podspec
```diff
@@ -0,0 +1,35 @@
...
+  #
+
+  spec.name         = "Fonts"
+  spec.version      = "0.0.1"
+  spec.summary      = "A short description of Fonts."
+  spec.description      = <<-DESC
+  A local pod that can optionaly contain Skyscanner fonts to include in the bundle.
+                       DESC
+  spec.homepage         = 'https://github.com/Skyscanner/backpack-ios'
+  spec.license          = { type: 'Apache-2.0', file: 'LICENSE' }
```
georgegillams, line 25: Need to clean up this file a bit, but it works

Fonts/Scripts/download-relative-fonts
```diff
@@ -9,8 +9,10 @@ git init
 git remote add origin git@github.skyscannertools.net:backpack/bpk-fonts.git
 git config core.sparsecheckout true
 echo "iOS" >> .git/info/sparse-checkout
-git pull --depth=1 origin main
+# We should be smarter about this, as it currently takes too long to fail.
+# Maybe we should only pull fonts if not already present, as most internal devs will have them from a previous install
+git pull --depth=1 origin main || echo "\033[0;31mFailed to pull internal Skyscanner fonts. If you want to develop with the fonts, check that you are on the VPN\033[0m"
```
georgegillams, line 14: For now I've just updated the existing font pulling script.
Need to make this a bit smarter and probably rewrite in ruby

Fonts/Scripts/download-relative-fonts.rb
```diff
@@ -0,0 +1,25 @@
...
+# This script pulls Skyscanner fonts only if `USE_RELATIVE=1`.
+# It adds them to the `Fonts` pod to make them available to the Example app build.
+
+require 'fileutils'
+
+if ENV["USE_RELATIVE"] == "1"
+  puts "Attempting to pull Skyscanner fonts"
+  FileUtils.mkdir_p 'bpk-fonts-tmp'
+  Dir.chdir("bpk-fonts-tmp") do
+    system "git init"
```
k0nserv, line 12: Can we surpress the output of these git commands in favour of only outputting our own messaging. Maybe we can hook into `CocoaPods` verbose mode and add output if it's enabled?

Fonts/Scripts/download-relative-fonts.rb
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env ruby
+
+# This script pulls Skyscanner fonts only if `USE_RELATIVE=1`.
+# It adds them to the `Fonts` pod to make them available to the Example app build.
+
+require 'fileutils'
+
+if ENV["USE_RELATIVE"] == "1"
```
k0nserv, line 8: Can we prefix this variable? `BPK_USE_RELATIVE` perhaps?

Fonts/Scripts/download-relative-fonts.rb
georgegillams: Yeah we can. All `stdout` is hidden unless you specify `--verbose` anyway

k0nserv (approved):

PR #991, opened 2021-06-30
[KOA-4294] Move BPKHorizontalNavigation to Swift

k0nserv:
This PR moves `BPKHorizontalNavigation` from Objective-C to Swift. 

As part of this the goal is to achieve an erconomic API for workign with the different available icon sizes in a type safe manner.

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigationItemDefault.swift
```diff
@@ -0,0 +1,208 @@
...
+            let iconHeight = icon.size.height
+            let width = buttonIntrinsicContentSize.width + 3 * Size.horizontalSpacing
+            let height = iconHeight + Size.verticalSpacing / 2.0
+
+            return CGSize(width: width, height: height)
+        } else {
+            let width = buttonIntrinsicContentSize.width + 2 * Size.horizontalSpacing
+            let height = buttonIntrinsicContentSize.height
+
+            return CGSize(width: width, height: height)
```
frugoman, line 135: I think the else is not needed and can be written like this
```swift 
if let icon = icon {
   let iconHeight = icon.size.height
   let width = buttonIntrinsicContentSize.width + 3 * Size.horizontalSpacing
   let height = iconHeight + Size.verticalSpacing / 2.0
 
   return CGSize(width: width, height: height)
}
let width = buttonIntrinsicContentSize.width + 2 * Size.horizontalSpacing
let height = buttonIntrinsicContentSize.height
 
return CGSize(width: width, height: height)
```
or even using a `guard let icon = icon else { return without icon logic }`

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigationItemDefault.swift
```diff
@@ -0,0 +1,208 @@
...
+            self.notificationDot.frame = CGRect(x: location.x, y: location.y, width: BPKSpacingMd, height: BPKSpacingMd)
+        }
+    }
+
+    private func notificationLocation(for labelFrame: CGRect) -> CGPoint {
+        if BPKRTLSupport.viewIsRTL(self) {
+            return CGPoint(x: labelFrame.minX - BPKSpacingMd, y: labelFrame.minY - BPKSpacingSm)
+        } else {
+            return CGPoint(x: labelFrame.maxX, y: labelFrame.minY - BPKSpacingSm)
+        }
```
frugoman, line 153: WDYT about this?
```swift
let rtlCoords: (x: CGFloat, y: CGFloat) = BPKRTLSupport.viewIsRTL(self) ? (BPKSpacingMd, BPXSpacingSm) : (0, 0)
return CGPoint(x: labelFrame.minX - rtlCoords.x, labelFrame.minY - rtlCoords.y)
```

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigationItemDefault.swift
```diff
@@ -0,0 +1,208 @@
...
+            titleEdgeInsets = UIEdgeInsets(
+                top: Size.verticalSpacing, left: Size.horizontalSpacing,
+                bottom: Size.verticalSpacing, right: Size.horizontalSpacing
+            )
+            return
+        }
+
+        let horizontalInset = Size.horizontalSpacing / 2.0
+
+        if BPKRTLSupport.viewIsRTL(self) {
```
frugoman, line 190: as a side comment:
if RTL is used as a configuration point, we can probably think of a configuration protocol that can have 2 or more implementations. This should make the views a bit more agnostic of the user's configuration and have that be configuration of the component. something like this:
```swift
protocol HorizontalNavigationItemDefaultViewConfiguration {
   func imageEdgeInsets(forSize: Size) -> UIEdgeInsets
   func titleEdgeInsets(forSize: Size) -> UIEdgeInsets
   func notificationLocation(fromCoords coords: (x: CGFloat, y: CGFloat)) -> CGPoint
}

class RTL_HNIDefaultViewConfiguration: HorizontalNavigationItemDefaultViewConfiguration {
   func imageEdgeInsets(forSize size: Size) -> UIEdgeInsets {
       return UIEdgeInsets(top: 0, left: size.horizontalSpacing / 2, bottom: 0, right: 0)
   }

   func titleEdgeInsets(forSize size: Size) -> UIEdgeInsets {
      UIEdgeInsets(top: 0, left: 0, bottom: 0, right: size.horizontalSpacing / 2)
   }
   func notificationLocation(fromCoords coords: (x: CGFloat, y: CGFloat)) -> CGPoint {
      return CGPoint(x: coords.x - BPKSpacingMd, y: coords.y - BPKSpacingSm)
   }
}

class NON_RTL_HNIDefaultViewConfiguration: HorizontalNavigationItemDefaultViewConfiguration {
   func imageEdgeInsets(forSize size: Size) -> UIEdgeInsets {
       return UIEdgeInsets(top: 0, left: 0, bottom: 0, right: size.horizontalSpacing / 2)
   }

   func titleEdgeInsets(forSize size: Size) -> UIEdgeInsets {
      UIEdgeInsets(top: 0, left: size.horizontalSpacing / 2, bottom: 0, right: 0)
   }
   func notificationLocation(fromCoords coords: (x: CGFloat, y: CGFloat)) -> CGPoint {
      return CGPoint(x: coords.x, y: coords.y)
   }
}
```

then it would be called like this

```swift
viewConfiguration.notificationlocation(fromCoords: (labelFrame.minx, labelFrame.minY))

imagEdgeinsets = viewConfiguration.imageInsets(forSize: size)
titleEdgeinsets = viewConfiguration.titleInsets(forSize: size)
```
WDYT?

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigationItemWithBadge.swift
```diff
@@ -0,0 +1,128 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Foundation
+
+internal class BPKHorizontalNavigationItemWithBadge<Size: BPKHorizontalNavigationSize>: UIControl, BPKHorizontalNavigationItem {
```
frugoman, line 21: `internal` is not needed

Backpack/HorizontalNavigation/Classes/HorizontalNavigation.h
```diff
@@ -18,12 +18,12 @@
...
-#import "BPKHorizontalNavigationOptionType.h"
-#import "BPKHorizontalNavigationOptionWithBadge.h"
-#import "BPKHorizontalNavigationSize.h"
+//#import "BPKHorizontalNavigation.h"
+//#import "BPKHorizontalNavigationAppearance.h"
+//#import "BPKHorizontalNavigationDelegate.h"
+//#import "BPKHorizontalNavigationOption.h"
+//#import "BPKHorizontalNavigationOptionType.h"
+//#import "BPKHorizontalNavigationOptionWithBadge.h"
+//#import "BPKHorizontalNavigationSize.h"
```
frugoman, line 27: I think this should be deleted instead of commented

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigation.swift
```diff
@@ -0,0 +1,520 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Foundation
+import UIKit
+
+
+public protocol BPKHorizontalNavigationSize {
```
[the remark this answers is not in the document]
k0nserv: It seems like this have been proposed but not included in Swift https://forums.swift.org/t/constrain-protocol-usage-outside-the-module/8832/96

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigation.swift
```diff
@@ -0,0 +1,520 @@
...
+    case alternate = 1
+}
+
+public protocol BPKHorizontalNavigationItem: UIControl {
+    var selectedColor: UIColor? { get set }
+
+    var appearance: BPKHorizontalNavigationAppearance { get set }
+}
+
+public protocol BPKHorizontalNavigationOption {
```
frugoman, line 61: This seems to be an ItemProvider instead of an Option instance.
Maybe a renderer.
wdyt about this approach?

```swift
public protocol ViewItem: UIControl { }

public protocol ViewRenderer {
   func render() -> ViewItem
}

// Examples
protocol SelectableViewItem: ViewItem {
   var selectedColor: UIColor { get }
}

class CellViewItem: SelectableViewItem {
    var selectedColor: UIColor = .red
}

public class SelectableViewRenderer: ViewRenderer {
   func renderer() -> ViewItem {
      return CellViewItem()
   }
}

class HorizontalNavigationItemWithBackground: ViewItem {}

class HorizontalNavigationOptionWithBackgroundRenderer<Size: BPKHNSize>: ViewRenderer {
   func render() -> ViewItem {
      return HorizontalNavigationItemWithBackground()
   }
}

class Container: ViewItem {}

class ListComposerRenderer: ViewRenderer {
   let renderers: [ViewRenderer] = [...]
   func render() -> ViewItem {
      let container = Container()
      renderers
            .map { $0.render() }
            .forEach(container.addSubview)
     return container
   }
}

// Usage
class HorizontalNavigation {
   init(renderer: ViewRenderer) {
      let renderedControl = renderer.render()
   }
}
```

frugoman: we can discuss further about this later

frugoman (commented):
In swift there's no need for the prefix BPK, since files are encapsulated in the modules, so no collision would happen.

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigation.swift
```diff
@@ -0,0 +1,520 @@
...
+import UIKit
+
+
+public protocol BPKHorizontalNavigationSize {
+    static var fontStyle: BPKFontStyle { get }
+
+    static var verticalSpacing: CGFloat { get }
+}
+
+public enum BPKHorizontalNavigationSizeDefault: BPKHorizontalNavigationSize {
```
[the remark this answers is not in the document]
k0nserv: The reason for enum is exactly because we only use this at the type level though. We don't want to create instance of it which an `enum` without cases nicely supports

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigation.swift
```diff
@@ -0,0 +1,520 @@
...
+
+    public static var fontStyle: BPKFontStyle { .textSmEmphasized }
+}
+
+extension BPKHorizontalNavigationSize {
+    static var horizontalSpacing: CGFloat { BPKSpacingBase }
+}
+
+public protocol BPKHorizontalNavigationDelegate: AnyObject {
+    func horizontalNavigation<Size>(_ horizontalNavigation: BPKHorizontalNavigation<Size>, didSelectItemAtIndex index: NSInteger, withTag tag: NSInteger)
```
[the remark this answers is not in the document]
k0nserv: Yup, I've already changed all usage of `NSInteger` to `Int` locally. I guess it's Objective-c force of habit :D

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigation.swift
```diff
@@ -0,0 +1,520 @@
...
+
+import Foundation
+import UIKit
+
+
+public protocol BPKHorizontalNavigationSize {
+    static var fontStyle: BPKFontStyle { get }
+
+    static var verticalSpacing: CGFloat { get }
+}
```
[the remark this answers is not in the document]
k0nserv: There's no point in allocating an instance here though. Seems cleaner to use static since we are after the type level difference and a couple of static properties

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigation.swift
```diff
@@ -0,0 +1,520 @@
...
+
+public protocol BPKHorizontalNavigationOption {
+    associatedtype Size: BPKHorizontalNavigationSize
+
+    /**
+     * A tag representing the option, can be used to distinguish options from
+     * eachother.
+     */
+    var tag: NSInteger { get }
```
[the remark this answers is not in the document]
k0nserv: This is replicating the existing behaviour which we presumably added for a reason. The user has no access to the underlying `UIControl` that gets created here so the `tag` is helpful to disambiguate

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigation.swift
k0nserv: I think this providers slightly more flexibility than desirable 🤔 

The thinking here is that `Option` represents the combination of model and the knowledge about how to create a view to render that model. I guess the best example of this is `UITabBarItem` which is kind of like our `Option` here and then there's some other view that we don't have access to that is actually rendered. The `Option` is exposed to the user whereas the view is hidden from them, you can provide custom options by defining your own option and item.

We want all potential items that we render to support both `selectedColor` and `appearance` which is why that's part of the `BPKHorizontalNavigationItem` protocol atm.

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigationItemDefault.swift
```diff
@@ -0,0 +1,208 @@
...
+        let size: CGSize
+    }
+
+    override var isSelected: Bool {
+        didSet {
+            updateStyle()
+        }
+    }
+
+    dynamic var selectedColor: UIColor? {
```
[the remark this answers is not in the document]
frugoman: ahh makes sense then!

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigationItemDefault.swift
```diff
@@ -0,0 +1,208 @@
...
+            return selectedColor ?? .bpk_primary
+        } else {
+            switch (appearance) {
+            case .normal: return .bpk_textPrimary
+            case .alternate: return .bpk_skyGrayTint07
+            }
+        }
+    }
+
+    init(name: String) {
```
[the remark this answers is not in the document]
frugoman: I think the safest place to initialise is the initialiser, since you can have convenience initialisers in the future that define different values. I guess it's a personal preference though, no real objective argue agains it (maybe only showing the possible values on autocompletion, when you actually call this initialiser)

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigationItemDefault.swift
frugoman: maybe haha

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigationItemDefault.swift
frugoman: oh I see!
Maybe those configuration structs can be internal then, so that all the configuration is centralised in a single point (`private protocol Config` `struct RTLConfig` or `struct LTRConfig`)

k0nserv: Yes maybe, but currently it's just a manner of mirroring so it feels like more work than it's worth

k0nserv:
> In swift there's no need for the prefix BPK, since files are encapsulated in the modules, so no collision would happen.
We've gone back and forth on this and the current decision is to keep the prefixes around. Especially due to the issues we had with not using clang modules in the main app.

Backpack.podspec
```diff
@@ -34,6 +34,7 @@ Pod::Spec.new do |s|
   s.ios.deployment_target = '12.0'
   s.source_files = 'Backpack/Backpack.h', 'Backpack/Common.h', 'Backpack/*/Classes/**/*.{h,m,swift}'
   s.public_header_files = 'Backpack/Backpack.h', 'Backpack/*/Classes/**/*.h'
+  s.private_header_files = 'Backpack/*/Classes/**/Private/*.h'
```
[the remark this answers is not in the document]
frugoman: Not really sure what this does or if we need it or not!

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigationItemDefault.swift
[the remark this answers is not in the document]
k0nserv: I think you are right but then you have to duplicate the logic and values in two places which can cause bugs. `BPKHorizontalNavigationItemDefault` will never be exposed publicly as it's just an internal API. With the defaults duplicated there's a chance we introduce bugs by changing something only in a single place

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigation.swift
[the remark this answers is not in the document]
frugoman: right, but my point is, why limit the clients of the library to only make it this way. I mean, why do we want to make it a type level enum? Maybe I'm missing context tho

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigation.swift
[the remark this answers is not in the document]
frugoman: Guess same as above, why do we want it to be type level ?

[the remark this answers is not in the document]
k0nserv: Because the goal is to achieve a type safe API that doesn't allow misuse, in particular it should be impossible to use a small icon with the default size of the horizontal navigation and vice versa.

This is the case now because of the `where` constraint in this definition

```swift
        public static func textAndIcon<
            S,
            Icon: BPKHorizontalNavigationOptionIcon
        >(name: String, icon: Icon, tag: Int = 0, showNotificationDot: Bool = false) -> AnyOption<S>
        where Icon.Size == S {
```

This prevents you from writing the following code:

```swift
let hoirzontalNav = BPKHorizontalNavigation<BPKHorizontalNavigationSizeDefault>(options: [
  .textAndIcon("Hotels", icon: BPKIconNameSmall.hotels, tag: 0)
], selectedItemIndex: 0)
```

In the previous Objective-c version we had a `size` property but that doesn't work as are migrating the icon API to one where not all icons are available at all sizes.

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigation.swift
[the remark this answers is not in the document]
k0nserv: See this [comment](https://github.com/Skyscanner/backpack-ios/pull/991#discussion_r662168661). The gist is that we want to provide the ability to use two sizes but limit the type of icons that can be used with each size to:

* `BPKHorizontalNavigationSizeDefault` -> `BPKLargeIconName`
* `BPKHorizontalNavigationSizeSmall` -> `BPKSmallIconName`

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigation.swift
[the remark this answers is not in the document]
frugoman: Hm, Ok, I see your point!
I think this is a use case that makes sense for the main app, but since this is an open source lib, maybe there are some cases where clients would like a bit more control over what to draw, even if it's now what our designers state, like using a big image with a small text.
Either way, I got why it's been done this way now!

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigationItemDefault.swift
frugoman: giving a bit more thought on it (left for debate), I think that this is useful for Skyscanner app, but this being open source, I think this would be forcing users to use RTL components, even if they don't want to.

k0nserv: Supporting RTL is part of our offering and open source is only secondary. Consumers can still disable RTL support by forcing LTR if they wish

georgegillams (changes requested):

k0nserv:
What nomenclature should we use for the string within the option: **title** or **name**? Historically we've used **name**, but `UIKit` seems to mostly favour **title** e.g. `UITabBarItem.title`.

WDYT @georgegillams ?

frugoman:
I know this might be work in progress, but can we avoid using force-pushing? it makes it easier to review new changes after reviews (we can always squash and merge later to avoid multiple commits showing on repo's history)

k0nserv:
>I know this might be work in progress, but can we avoid using force-pushing? it makes it easier to review new changes after reviews (we can always squash and merge later to avoid multiple commits showing on repo's history)
Sorry yeah can do if you are reviewing. The "viewed"  function in GitHub works pretty well though in my experience

k0nserv:
The current CI failures are weird. I don't see them locally with `be rake ci` and they seem entirely unrelated to this PR. I'm testing the issue in #999

Example/SnapshotTests/Backpack_SnapshotTests-Bridging-Header.h
```diff
@@ -15,15 +15,3 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
```
georgegillams, line 17: Is there any point in this file anymore?

Example/SnapshotTests/BPKSnapshotTest.swift
```diff
@@ -0,0 +1,80 @@
...
+        if let identifier = identifier {
+            finalIdentifier = "\(finalIdentifier)_\(identifier)"
+        }
+
+        if #available(iOS 13.0, *) {
+            view.overrideUserInterfaceStyle = .light
+        }
+
+        FBSnapshotVerifyView(view, identifier: finalIdentifier)
+    }
```
georgegillams, line 63: We already wrote macros for this, but I guess they don't work from swift hence adding them here?

k0nserv: Yeah, macros are a C thing so they cannot be used from Swift

Example/SnapshotTests/Backpack_SnapshotTests-Bridging-Header.h
k0nserv: I think I added in this PR. Unsure if it's needed at all, I'll try removing it

k0nserv: Looks like we need it

Backpack/HorizontalNavigation/README.md
```diff
@@ -4,40 +4,67 @@
...
+      }
+  }
+
+  // ...omitted
+}
+```
+
 
-### Appearance attributes
-`(UIColor)contentColor`
```
georgegillams, line 43: I still think this should be documented, even though we've got the section about the bug

k0nserv: The section describes it though and it's not an appearance attribute anymore

georgegillams: Can't really argue with that 😂

georgegillams (approved):

PR #1014, opened 2021-08-04
[KOA-4654] Align switch color to figma

georgegillams:
+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


Backpack/Switch/Classes/BPKSwitch.m
```diff
@@ -33,8 +33,6 @@ - (instancetype)initWithCoder:(NSCoder *)coder {
     BPKAssertMainThread();
     self = [super initWithCoder:coder];
     if (self) {
-        _primaryColor = BPKColor.skyBlue;
```
georgegillams, line 36: This is a nullable property (for theming), and if it's not set we just use skyBlue. So there was no point in setting this.

olliecurtis (approved):

PR #1037, opened 2021-09-06
[KOA-4422] Improve Toast with VoiceOver

k0nserv:
+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/Backpack/Toast/Classes/BPKToast.m
+++ b/Backpack/Toast/Classes/BPKToast.m
@@ -48,7 +48,6 @@ NSString *const ToastAccessibilityIdentifier = @"toastView";
         self.hud = [MBProgressHUD showHUDAddedTo:view animated:animated];
         _accessibilityAnnouncement = [accessibilityAnnouncement copy];
         [self setupHUD];
-        [self postAccessibilityNotification];
     }
     return self;
 }
--- a/Backpack/Toast/Classes/BPKToast.m
+++ b/Backpack/Toast/Classes/BPKToast.m
@@ -31,22 +31,31 @@ NSString *const ToastAccessibilityIdentifier = @"toastView";
 @implementation BPKToast
 
 #pragma mark - Initializers
-+ (instancetype)showToastAddedTo:(UIView *)view animated:(BOOL)animated {
-    BPKToast *toast = [[BPKToast alloc] initWithView:view animated:animated];
++ (instancetype)showToastAddedTo:(UIView *)view
+                        animated:(BOOL)animated
+       accessibilityAnnouncement:(nonnull NSString *)accessibilityAnnouncement {
+    BPKToast *toast = [[BPKToast alloc] initWithView:view
+                                            animated:animated
+                           accessibilityAnnouncement:accessibilityAnnouncement];
     return toast;
 }
 
-- (instancetype)initWithView:(UIView *)view animated:(BOOL)animated {
+- (instancetype)initWithView:(UIView *)view
+                    animated:(BOOL)animated
+   accessibilityAnnouncement:(nonnull NSString *)accessibilityAnnouncement {
     if (self = [super initWithFrame:view.bounds]) {
         self.hud = [MBProgressHUD showHUDAddedTo:view animated:animated];
+        _accessibilityAnnouncement = [accessibilityAnnouncement copy];
         [self setupHUD];
+        [self postAccessibilityNotification];
     }
     return self;
 }
 
-- (instancetype)initWithView:(UIView *)view {
+- (instancetype)initWithView:(UIView *)view accessibilityAnnouncement:(nonnull NSString *)accessibilityAnnouncement{
     if (self = [super initWithFrame:view.bounds]) {
         self.hud = [[MBProgressHUD alloc] initWithView:self];
+        _accessibilityAnnouncement = [accessibilityAnnouncement copy];
         [self setupHUD];
         [self addSubview:self.hud];
     }
@@ -56,6 +65,7 @@ NSString *const ToastAccessibilityIdentifier = @"toastView";
 #pragma mark - Public methods
 - (void)show:(BOOL)animated {
     [self.hud showAnimated:animated];
+    [self postAccessibilityNotification];
 }
 
 - (void)hide:(BOOL)animated {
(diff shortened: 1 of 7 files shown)


Backpack/Toast/Classes/BPKToast.m
```diff
@@ -128,12 +131,19 @@ - (void)hudWasHidden:(MBProgressHUD *)hud {
...
 - (void)setupHUD {
     self.hud.bezelView.style = MBProgressHUDBackgroundStyleSolidColor;
     self.hud.bezelView.backgroundColor = [[BPKColor dynamicColorWithLightVariant:BPKColor.skyGray darkVariant:BPKColor.blackTint03] colorWithAlphaComponent:0.85];
     self.hud.contentColor = BPKColor.white;
     self.hud.delegate = self;
     self.hud.accessibilityIdentifier = ToastAccessibilityIdentifier;
+    self.hud.accessibilityViewIsModal = YES;
+}
+
+- (void)postShowAccessibilityNotification {
```
georgegillams, line 145: Maybe the name of this method could be clearer? Maybe `postAccessibilityNotification`?

Backpack/Toast/Classes/BPKToast.m
```diff
@@ -128,12 +131,19 @@ - (void)hudWasHidden:(MBProgressHUD *)hud {
...
+    UIAccessibilityPostNotification(UIAccessibilityLayoutChangedNotification, self.hud.label);
```
georgegillams, line 146: I'm not sure whether we should send focus to the toast, or just announce it leaving focus intact.
Sending focus to the toast could be a poor experience if it disappears automatically. eg what happens if someone is still hearing the content when it disappears? Where will focus go to after it does.

Backpack/Toast/Classes/BPKToast.m
k0nserv: Sure

Backpack/Toast/Classes/BPKToast.m
k0nserv: Yeah I was concerned about that too. Doing it this way would be a breaking change to introduce an accessibility label

georgegillams: We could easily adopt such a breaking change in the app by just combining the title and description. We can inform squads of this and ask them to fix if necessary?
It would still be better than what we currently have

k0nserv: Yeah fair

Backpack/Toast/Classes/BPKToast.m
```diff
@@ -128,12 +146,24 @@ - (void)hudWasHidden:(MBProgressHUD *)hud {
...
+
 - (void)setupHUD {
     self.hud.bezelView.style = MBProgressHUDBackgroundStyleSolidColor;
     self.hud.bezelView.backgroundColor = [[BPKColor dynamicColorWithLightVariant:BPKColor.skyGray darkVariant:BPKColor.blackTint03] colorWithAlphaComponent:0.85];
     self.hud.contentColor = BPKColor.white;
     self.hud.delegate = self;
     self.hud.accessibilityIdentifier = ToastAccessibilityIdentifier;
+    self.hud.isAccessibilityElement = NO;
+    self.hud.label.isAccessibilityElement = NO;
+    self.hud.detailsLabel.isAccessibilityElement = NO;
```
georgegillams, line 159: Do we need to hide the spinner too?

Backpack/Toast/Classes/BPKToast.m
```diff
@@ -31,22 +31,31 @@ @interface BPKToast () <MBProgressHUDDelegate>
...
 
-- (instancetype)initWithView:(UIView *)view animated:(BOOL)animated {
+- (instancetype)initWithView:(UIView *)view
+                    animated:(BOOL)animated
+   accessibilityAnnouncement:(nonnull NSString *)accessibilityAnnouncement {
     if (self = [super initWithFrame:view.bounds]) {
         self.hud = [MBProgressHUD showHUDAddedTo:view animated:animated];
+        _accessibilityAnnouncement = [accessibilityAnnouncement copy];
         [self setupHUD];
+        [self postAccessibilityNotification];
```
georgegillams, line 50: Should this announcement happen when the Toast is initialised? What if we initialise it but don't show it immediately?

k0nserv: This is an internal init which is only called form `showToast:` but you are probably right that it'd be better to move this call

Backpack/Toast/Classes/BPKToast.m
k0nserv: yes, but it's not publicly accessible from what I can tell

Backpack/Toast/Classes/BPKToast.m
georgegillams: Yeah just trying to future proof it

Backpack/Toast/Classes/BPKToast.m
georgegillams: Maybe we should recurse over the subviews then?

Backpack/Toast/Classes/BPKToast.m
```diff
@@ -31,22 +31,31 @@ @interface BPKToast () <MBProgressHUDDelegate>
...
+       accessibilityAnnouncement:(nonnull NSString *)accessibilityAnnouncement {
+    BPKToast *toast = [[BPKToast alloc] initWithView:view
+                                            animated:animated
+                           accessibilityAnnouncement:accessibilityAnnouncement];
+    [toast postAccessibilityNotification];
     return toast;
 }
 
-- (instancetype)initWithView:(UIView *)view animated:(BOOL)animated {
+- (instancetype)initWithView:(UIView *)view
```
georgegillams, line 44: On second thoughts, should this method be called `showWithView` because it creates and shows the hud?

k0nserv: Eventually we have to call an `init` though and that `init` has to call `showHUDAddedTo:animated:` anyway

georgegillams (approved):

PR #1077, opened 2021-11-05
[no jira] Update DANGER_GITHUB_API_TOKEN to GITHUB_TOKEN

zgeor:
Updated the danger CI token in the PR workflow.

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -4,6 +4,8 @@ Thanks for contributing to Backpack :pray:
 Please include a description of the changes you are introducing and some screenshots if appropriate.
 -->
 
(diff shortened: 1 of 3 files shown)


marianeum (approved):

frugoman (approved):

zgeor:
Thanks to @marianeum I realised the new contributor greeting will not work with the GITHUB_TOKEN. This is due to the fact that it doesnt have read:org permission in order to check the contributor membership to backpack. 

GITHUB_TOKENS are safer than personal tokens, as they are regenerated for each run. So I think this is an acceptable tradeoff.

marianeum (approved):

PR #1128, opened 2022-01-05
KOA-4768 migrate to SceneDelegate

jookes:
Migrate the AppDelegate from ObjectiveC to Swift and split into the newer SceneDelegate

+ [ ] Check this if you have read and followed the [contributing guidelines](https://github.com/Skyscanner/backpack-android/blob/main/CONTRIBUTING.md)

+ [ ] `UNRELEASED.md`
+ [ ] `README.md`
+ [ ] Tests


Example/Backpack/SceneDelegate.swift
```diff
@@ -0,0 +1,87 @@
...
+
+import UIKit
+import Backpack
+
+class SceneDelegate: UIResponder, UIWindowSceneDelegate {
+
+    var window: UIWindow?
+    var isUITestingEnabled = false
+
+    // swiftlint:disable line_length
```
zgeor, line 28: Any reason to disable the lint rule?

jookes: When I split it into multiple lines, it will complain about indentation :/.

zgeor: Can the indentation be fixed? Or are we using an incompatible ruleset?

jookes: With the current ruleset we either:
- hit `line_length`
- hit `indentation_width`
But when we solve indentation_width, we will hit `vertical_parameter_alignment`

@frugoman What's your opinion on this?
We could disable the `line_length` or `indentation_width`, both would be fair game. But we can't get around both of these at the same time :( 

Relevant discussion:
https://github.com/realm/SwiftLint/issues/3046

frugoman: In these cases I do commonly prefer this notation
```swift
func scene(
   _ scene: UIScene,
   willConnectTo session: UIScene,
   ....
) {
   //code here
}
```
not sure if there's a rule against it tho

frugoman (approved):
Looking good!

PR #1129, opened 2022-01-06
Koa/koa 3827 badge to swift

frugoman:
Migrating Badge to swift

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


Backpack/Badge/Classes/Badge.swift
```diff
@@ -0,0 +1,109 @@
...
+    }
+    
+    private func setup() {
+        addSubview(label)
+        label.translatesAutoresizingMaskIntoConstraints = false
+        
+        label.leadingAnchor.constraint(equalTo: leadingAnchor, constant: BPKSpacingMd).isActive = true
+        trailingAnchor.constraint(equalTo: label.trailingAnchor, constant: BPKSpacingMd).isActive = true
+        label.topAnchor.constraint(equalTo: topAnchor, constant: BPKSpacingSm).isActive = true
+        bottomAnchor.constraint(equalTo: label.bottomAnchor, constant: BPKSpacingSm).isActive = true
```
jookes, line 69: We could use NSLayoutConstraint.activate([ ]}. This should be 'more efficient' according to the docs
https://developer.apple.com/documentation/uikit/nslayoutconstraint/1526955-activate

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigationBadge.swift
```diff
@@ -0,0 +1,47 @@
...
+    }
+    
+    init(color: UIColor) {
+        self.color = color
+        super.init(frame: .zero)
+        type = .outline
+        updateStyle(with: color)
+    }
+    
+    required init?(coder: NSCoder) {
```
jookes, line 32: Can we add `@available(*, unavailable)` above this function? This way no one is able to call it by accident.
It does seem however that the BPKBadge class did implement this, so we might want to call `super`?

Backpack/Badge/Classes/Badge.swift
```diff
@@ -0,0 +1,109 @@
...
+        get { label.text }
+        set { label.text = newValue }
+    }
+    
+    public var type: BPKBadgeType = .success {
+        didSet { updateLookAndFeel() }
+    }
+    
+    public convenience init(type: BPKBadgeType = .success, message: String) {
+        self.init(frame: .zero)
```
jookes, line 35: The init with frame already calls setup() is it still needed to call `setup()` at the end of this init?

Example/Tests/BPKBedgeTest.swift
```diff
@@ -0,0 +1,65 @@
...
+    
+    private func createViewWithTypes() -> UIView {
+        let stackView = createStackView()
+        let parentView = UIView(frame: .zero)
+        parentView.backgroundColor = BPKColor.skyGrayTint06
+        parentView.translatesAutoresizingMaskIntoConstraints = false
+        parentView.addSubview(stackView)
+        
+        let badges = [BPKBadgeType.light, .inverse, .outline, .success, .warning, .destructive]
+            .map { BPKBadge(type: $0, message: "Backpack rocks") }
```
jookes, line 47: Do we want to add an `!` to keep in sync with the original test? (Backpack rocks!)

frugoman: this tests was actually like this! it's another test that has the bang!

Backpack/Badge/Classes/Badge.swift
frugoman: good catch!

Backpack/HorizontalNavigation/Classes/BPKHorizontalNavigationBadge.swift
frugoman: only issue with that is that a color must then be assumed while initializing.
having the color to be optional doesn't really make sense conceptually just to avoid this.
I'll set an accent principal color as a default

Backpack/Badge/Classes/Badge.swift
frugoman: I replaced calling setup with just calling updateLookAndFeel, since didSet are not called if inside an init and didn't want to add a `defer` statement

Backpack/Badge/Classes/Badge.swift
```diff
@@ -0,0 +1,111 @@
...
+        self.message = message
+        self.type = type
+        updateLookAndFeel()
+    }
+    
+    public override init(frame: CGRect) {
+        super.init(frame: frame)
+        setup()
+    }
```
jookes, line 44: In the original version, the setup() function would default to the `.success` type. We should be careful not to break this flow if implementers assumed this default behaviour

frugoman: This is now done in the initialisation of the type property! `.success` will be the default

Backpack/Badge/Classes/Badge.swift
```diff
@@ -0,0 +1,111 @@
...
+    
+    public var type: BPKBadgeType = .success {
+        didSet { updateLookAndFeel() }
+    }
+    
+    public convenience init(type: BPKBadgeType = .success, message: String) {
+        self.init(frame: .zero)
+        self.message = message
+        self.type = type
+        updateLookAndFeel()
```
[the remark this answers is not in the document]
frugoman: This is actually not the case, didSet (and willSet) methods are only called if the var is modified outside of inits!
So in this case, calling `updateLookAndFeel()` is actually needed!

PR #1131, opened 2022-01-07
KOA-3827 badge icon

frugoman:
Adding support for icons on Badges

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


Backpack/Badge/Classes/Badge.swift
```diff
@@ -0,0 +1,158 @@
...
+        imageView.translatesAutoresizingMaskIntoConstraints = false
+        return imageView
+    }()
+    private let containerStackView: UIStackView = {
+        let stack = UIStackView(frame: .zero)
+        stack.translatesAutoresizingMaskIntoConstraints = false
+        stack.axis = .horizontal
+        stack.spacing = BPKSpacingSm
+        return stack
+    }()
```
jookes, line 56: Should we use `private lazy var` for these?

Backpack/Badge/Classes/Badge.swift
```diff
@@ -0,0 +1,158 @@
...
+            layer.borderColor = BPKColor.white.cgColor
+            layer.borderWidth = BPKBorderWidthSm
+        }
+        
+        iconView.image = icon.orNil(forType: type)
+        placeElements()
+    }
+    
+    private func placeElements() {
+        containerStackView.arrangedSubviews.forEach(containerStackView.removeArrangedSubview)
```
jookes, line 91: We should also call view.removeFromSuperView() to make sure we don't cause any bugs.
https://stackoverflow.com/a/53260606/2611649 (Related StackOverflow)

Backpack/Badge/Classes/Badge.swift
```diff
@@ -0,0 +1,158 @@
...
+    }
+    
+    private func setup() {
+        addSubview(containerStackView)
+        
+        layer.cornerRadius = BPKCornerRadiusXs
+        layer.masksToBounds = true
+        
+        NSLayoutConstraint.activate([
+            heightAnchor.constraint(greaterThanOrEqualToConstant: 24),
```
jookes, line 109: Should we declare a constant for this magic number?

jookes (commented):
The icon is an awesome addition! Could we also test if the icon is positioned in the correct location for RTL layouts (Leading -> right, trailing -> left)?

Backpack/Badge/Classes/Badge.swift
frugoman: Good one!

Backpack/Badge/Classes/Badge.swift
frugoman: I think this makes sense when the views are used further down the line, after the view's been initialised. This are all subviews that are used to setup the badge, so I don't think we win anything making them lazy. Also, it would force us to make them optional and to `?.` every time we want to use them.

frugoman: there's also the fact that making them `var`, even if they're private, they're communicating that are variables that are ok to be reassigned, and in this case does not make much sense... but I can agree that this is more on the philosophical side of the discussion hahaha

Backpack/Badge/Classes/Badge.swift
frugoman: I should use an existing one in fact!

jookes (approved):

PR #1177, opened 2022-02-07
KOA-4835 mapping old styles to new ones behind the scenes

frugoman:
mapping old styles to new ones behind the scenes so that migration can happen seamlessly in the future.

+ [x] `UNRELEASED.md`


Backpack/Font/Classes/Generated/BPKFont.m
```diff
@@ -228,8 +230,40 @@ + (UIFont *)fontForStyle:(BPKFontStyle)style fontManager:(BPKFontManager *)fontM
     }
 }
 
-+ (NSNumber *)letterSpacingForStyle:(BPKFontStyle)style {
++ (BPKFontStyle)newStyleMappingForStyle:(BPKFontStyle)style {
     switch (style) {
+        case BPKFontStyleTextBase:
```
marianeum, line 235: should we also map the actually deprecated styles to their replacement (e.g. the heavy styles)?

marianeum (approved):

PR #1211, opened 2022-03-14
[KOA-3844] calendar UI tests

frugoman:
Adding UI Tests for Happy paths on Calendar

+ [x] `UNRELEASED.md`
+ [x] Tests

### DIFF
--- a/Example/Backpack NativeUITests/CalendarUITest.swift
+++ b/Example/Backpack NativeUITests/CalendarUITest.swift
@@ -84,5 +84,17 @@ class CalendarUITest: BackpackUITestCase {
         XCTAssertTrue(app.cells["March 16, 2020"].isSelected)
         XCTAssertTrue(app.cells["March 10, 2020"].isSelected)
         XCTAssertTrue(app.cells["March 13, 2020"].isSelected)
+        
+        collectionViewsQuery.cells["March 1, 2020"].staticTexts["1"].tap()
+        collectionViewsQuery.cells["March 3, 2020"].staticTexts["3"].tap()
+        collectionViewsQuery.cells["March 16, 2020"].staticTexts["16"].tap()
+        collectionViewsQuery.cells["March 10, 2020"].staticTexts["10"].tap()
+        collectionViewsQuery.cells["March 13, 2020"].staticTexts["13"].tap()
+        
+        XCTAssertFalse(app.cells["March 1, 2020"].isSelected)
+        XCTAssertFalse(app.cells["March 3, 2020"].isSelected)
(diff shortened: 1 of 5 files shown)


Example/Backpack NativeUITests/CalendarUITest.swift
```diff
@@ -0,0 +1,88 @@
...
+ * limitations under the License.
+ */
+
+import XCTest
+
+class CalendarUITest: BackpackUITestCase {
+    private func goToPreselectedDatesCalendar(_ app: XCUIApplication) {
+        let tablesQuery = app.tables
+        tablesQuery.staticTexts["Calendar"].tap()
+        tablesQuery.staticTexts["With preselected dates"].tap()
```
frugoman, line 25: Using this one since it's the story in which it's calendar is not always the current date, but a fixed date, so that tests always run in the same dates

Example/Backpack NativeUITests/CalendarUITest.swift
```diff
@@ -0,0 +1,88 @@
...
+        XCTAssertFalse(app.cells["March 13, 2020"].isSelected)
+        
+        let collectionViewsQuery = app.collectionViews
+        collectionViewsQuery.cells["March 1, 2020"].staticTexts["1"].tap()
+        collectionViewsQuery.cells["March 3, 2020"].staticTexts["3"].tap()
+        collectionViewsQuery.cells["March 16, 2020"].staticTexts["16"].tap()
+        collectionViewsQuery.cells["March 10, 2020"].staticTexts["10"].tap()
+        collectionViewsQuery.cells["March 13, 2020"].staticTexts["13"].tap()
+        
+        XCTAssertTrue(app.cells["March 1, 2020"].isSelected)
```
marianeum, line 82: should we also include other dates that are NOT selected? (ie not the days on either side of a date for example)

frugoman: Yes! But it would be hard to know if the tests is actually effective.
Since picking the unselected dates to check would be discretional, we may never see the issue (if one comes out)
An option would be rather brute-force checking of all other dates. WDYT?

frugoman: Yeah! I can add those checks!
Although I'm not really sure if it would add any value, since to be sure nothing else is selected, we should check all other dates. This would fit in a Unit Tests rather than a UI one I think.
WDYT?

marianeum: yeah i guess the alternative is that it's similar to the above test where some get selected & then unselected

marianeum (approved):

PR #1224, opened 2022-03-25
big buttons now center it's content to match multiple app usages

frugoman:
Big buttons now center it's content to match usages on the app.

+ [x] `UNRELEASED.md`
+ [x] Tests


Backpack/NavigationBar/Classes/BPKNavigationBar.m
```diff
@@ -276,6 +276,7 @@ - (BPKNavigationBarButton *)leftButton {
     if (!_leftButton) {
         _leftButton = [[BPKNavigationBarButton alloc] init];
         _leftButton.translatesAutoresizingMaskIntoConstraints = NO;
+        _leftButton.title = @"";
```
frugoman, line 279: Kind of a hack, to remove margins on the sides when navigation bar buttons are icon only

Backpack/Button/Classes/BPKButton.m
```diff
@@ -155,9 +157,16 @@ - (void)setup {
...
+    self.stackLeadingFallbackConstraint = [self.contentStack.leadingAnchor constraintEqualToAnchor:self.contentView.leadingAnchor
+                                                                                          constant:BPKSpacingBase];
+    self.stackLeadingFallbackConstraint.priority = UILayoutPriorityDefaultLow;
+    self.stackTrailingFallbackConstraint = [self.contentStack.trailingAnchor constraintEqualToAnchor:self.contentView.trailingAnchor
+                                                                                            constant:-BPKSpacingBase];
+    self.stackTrailingFallbackConstraint.priority = UILayoutPriorityDefaultLow;
+    self.stackLeadingConstraint = [self.contentStack.leadingAnchor constraintGreaterThanOrEqualToAnchor:self.contentView.leadingAnchor
+                                                                                               constant:BPKSpacingBase];
+    self.stackTrailingConstraint = [self.contentStack.trailingAnchor constraintLessThanOrEqualToAnchor:self.contentView.trailingAnchor
+                                                                                              constant:-BPKSpacingBase];
```
frugoman, line 169: Setting up a secondary constraint for the content so that if size grows more than the content itself, the content remains centered

Backpack/NavigationBar/Classes/BPKNavigationBar.m
marianeum: might that break other button usages where they somehow give it an empty string title? (particularly the hasTitle change)

frugoman: Actually no, this is just a default value for the title for the navigation bar buttons.
`hasTitle` will return `true` now, making the BPKButton think that it's not an icon only button. This makes the Link button to NOT have side margins, which is what's required for the navigation bar buttons.
This is a hack of course and a proper refactor on the Navigation Bar is needed, but it would require some ""big"" changes on the app

marianeum (approved):

PR #1244, opened 2022-04-16
Simplify Spacing and Radius token demo pages

gert-janvercauteren:
Introduce
- SpacingsView (SwifUI)
- RadiusView (SwiftUI)

Remove
- SpacingsViewController (Objc)
- RadiiViewController (Objc)
- 2 views on Main.storyboard

The spacingsView now has a visual indicator for how big the spacing is
The radiiView now shows all different available cornerRadius options, the original implementation only displayed 2

simulator_screenshot_2D145B22-5F6D-44AE-8F2B-6704A720DC96

simulator_screenshot_D112E731-2483-436C-895C-0D3ECBAA7877

+ [ ] `UNRELEASED.md`
+ [ ] `README.md`
+ [ ] Tests


Example/Backpack/ViewControllers/SwiftUIHostViewController.swift
```diff
@@ -0,0 +1,63 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import UIKit
+import SwiftUI
+
+class SwiftUIHostViewController<Content: View>: UIViewController {
+    
+    public var content: Content?
```
frugoman, line 25: can we make the contentView `not optional` and receive the `content` on the initialiser as `non optional`? by how the class work, it doesn't seem to be conceptually optional.
it will ensure there's content to be setup in the viewDidLoad, to avoid temporal coupling.
It can probably be a `let` too
also no need to store the content instance.

```swift
class ContentUIHostingController<Content: View>: UIViewController {
  private let contentView: UIHostingController<Content>

  init(_ content: Content) { 
    contentView = UIHostingController(rootView: content)
    super.init()
  }
  required init?(coder: NSCoder) {
    fatalError("init(coder:) has not been implemented")
  }
  
  override func viewDidLoad() {
    super.viewDidLoad()
    addChild(contentView)
    view.addSubview(contentView.view)
    setupConstraints()
  }
  private func setupConstraints() { ... }
}
// And use it like ...
ContentUIHostingController(Text("Hello!))
```

Example/Backpack/ViewControllers/SwiftUIHostViewController.swift
```diff
@@ -0,0 +1,63 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import UIKit
+import SwiftUI
+
+class SwiftUIHostViewController<Content: View>: UIViewController {
```
frugoman, line 23: opinionated suggestion:
could this be named something like `ContentUIHostingController`? to be a bit more explicit on what it works with rather than what it should be used in

frugoman, line 23: opinionated suggestion:
could this be named something like `ContentUIHostingController`? to be a bit more explicit on what it works with rather than what it should be used in

frugoman (changes requested):

Example/Backpack/ViewControllers/ContentUIHostingController.swift
```diff
@@ -0,0 +1,55 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import UIKit
+import SwiftUI
+
+class ContentUIHostingController<Content: View>: UIViewController {
+    
+    public var content: Content?
```
frugoman, line 25: this var can be removed, right?

frugoman (approved):

PR #1264, opened 2022-04-25
[KOA-5035] Creating base SwiftUI module

frugoman:
Creating SwiftUI module


Rakefile
```diff
@@ -56,7 +57,7 @@ def repeat_on_fail(command, run_count = 1)
   end
 
   puts "Running process. This is attempt #{run_count} of #{MAX_TEST_REPEATS}."
-  `#{command}`
+  sh command
```
frugoman, line 60: shows the logs clearer (it's also used all across the Rakefile as well)

Rakefile
```diff
@@ -85,8 +86,8 @@ namespace :git do
 end
 
 desc "Check for build-time errors and warnings"
-task :analyze do
-  sh "set -o pipefail && xcodebuild -workspace #{EXAMPLE_WORKSPACE} -scheme \"#{EXAMPLE_SCHEMA}\" SWIFT_VERSION=#{SWIFT} -sdk #{BUILD_SDK} -destination \"#{DESTINATION}\" ONLY_ACTIVE_ARCH=YES analyze 2>&1 | xcpretty"
+task :analyze, [:scheme] do |t, args|
```
frugoman, line 89: parametrised the task to reuse in each scheme

Rakefile
```diff
@@ -124,8 +124,24 @@ task :git_checks do
...
+
+task :ci_uikit do
+  task(:all_checks).invoke(EXAMPLE_SCHEME)
+  task(:all_checks).reenable
+end
+
+task :ci_swiftui do
+  task(:all_checks).invoke(SWIFTUI_SCHEME)
+  task(:all_checks).reenable
+end
```
frugoman, line 137: because ruby and the Rakefile are weird, its much easier to have 2 tasks that send parameters like this, instead of having a single ci task and have the param being sent by the github action yml.

Rakefile
```diff
@@ -124,8 +124,24 @@ task :git_checks do
   abort red 'Git branch is not up to date please pull' unless branch_up_to_date
 end
 
-task ci: [:erase_devices, :all_checks]
-task all_checks: [:lint, :analyze, :test]
+task ci: [:erase_devices, :lint, :ci_uikit, :ci_swiftui]
+
+task :ci_uikit do
+  task(:all_checks).invoke(EXAMPLE_SCHEME)
+  task(:all_checks).reenable
```
frugoman, line 131: need to call `reenable` to avoid Rake not calling the `all_checks` tasks a second time (even if it has different params!!!)

Rakefile
```diff
@@ -161,30 +177,55 @@ task :release_no_checks do
...
-  puts "Updating podspec."
-  contents = File.read(PODSPEC)
-  contents.gsub!(/s\.version\s*=\s(:?'|")\d+\.\d+\.\d+(-\w+\.\d)?(:?'|")/, "s.version          = \"#{version_string}\"")
-  File.open(PODSPEC, 'w') { |file| file.puts contents }
-  abort red "Podspec should have been updated with the new version, but it wasn't." unless file_is_dirty(PODSPEC)
 
+  task(:update_version_in_podspec).invoke(UIKIT_PODSPEC, version_string)
+  task(:update_version_in_podspec).reenable
+  task(:update_version_in_podspec).invoke(SWIFTUI_PODSPEC, version_string)
+  task(:check_changelog_version).invoke(version_string)
```
frugoman, line 184: only need to call `reenable` on tasks that will be called more than once. This is not the case so no need to call it everywhere :)

Rakefile
```diff
@@ -161,30 +177,55 @@ task :release_no_checks do
...
+  abort red "Installing pods in the Example project failed" unless install_pods_in_example_project
+  task(:push_tag).invoke(version_string)
+  task(:push_cocoapods_trunk).invoke(UIKIT_PODSPEC)
+  task(:push_cocoapods_trunk).reenable
+  task(:push_cocoapods_trunk).invoke(SWIFTUI_PODSPEC)
+
+  puts green("🎉 All went well. Version #{version_string} published.")
+end
+
+task :check_changelog_version, [:version] do |t, args|
```
frugoman, line 194: created these small tasks to make the `relese_no_checks` task simpler to read

Backpack-SwiftUI.podspec
```diff
@@ -0,0 +1,45 @@
...
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+Pod::Spec.new do |s|
+  s.name             = 'Backpack-SwiftUI'
+  s.version          = "48.1.0"
```
gert-janvercauteren, line 19: Are we able to pull in the version from a central location? Then we can make sure all pods use the same version when releasing

olliecurtis (approved):
LGTM!

PR #1266, opened 2022-04-27
[KOA-4720] adding source kitten to generate swift docs with jazzy

frugoman:
Jazzy states in their docs that for combined projects with ObjC and Swift, there's an extra step needed for swift code to be picked up to generate docs. This step is to first create a file that defines how to scan the swift code. SourceKitten plays this role and it's the recommended (by Jazzy) tool for it.

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,6 +55,9 @@ jobs:
 
       - name: Install clang format
         run: brew install clang-format@14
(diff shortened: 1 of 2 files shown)


marianeum (approved):
i approve just for the name alone

PR #1273, opened 2022-05-03
Adding spacing tokens generation for swiftui

frugoman:
Adding spacing tokens generation for SwiftUI

Implemented an API for the tokens that's friendly with SwiftUI, which improves discoverability as well.

It can now be used like this 👍 :
```swift
Text("Hi!").padding(.lg)
Text("Hey!").padding(.top, .base)
Text("Alo").frame(width: .lg, height: .base)
Text("Ciao").frame(maxWidth: .lg, maxHeight: .xxl)
```
Instead of like this 👎 
```swift
Text("Hi!").padding(BPKSpacingLg)
Text("Hey!").padding(.top, BPKSpacingBase)
Text("Alo").frame(width: BPKSpacingLg, height: BPKSpacingBase)
Text("Ciao").frame(maxWidth: BPKSpacingLg, maxHeight: BPKSpacingXxl)
```

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


Backpack-SwiftUI/Spacing/Classes/Generated/Spacing.swift
```diff
@@ -0,0 +1,94 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+ 
+import SwiftUI
+
+public struct Spacing {
```
gert-janvercauteren, line 21: Would it be a bad idea to call this BPKSpacing? (To stay on brand).
Also wondering if we could have done with an extension on CGFloat, that way we wont need to extend any of the initializers, which would reduce toil in the future

frugoman: This gives us more control on how to apply the spacing to the components.
For SwiftUI native components, we need to create the extensions that you see in this file. But for our BackpackSwiftUI components, it gives us full control of the component and blocks misusage by design.
For example, if our components receive a cgfloat for setting a padding, this makes it possible to set a random number that's not been defined by our guidelines, which goes agains our goal of standardising the product

frugoman: keen to discuss further!

frugoman: There is a case for files that will have this problem, like Color for example, there we wouldn't have a choice but to name it BPKColor. Stablishing BPK as a prefix may be the way to go for consistency across backpack supported tokens/components

gulpfile.js
```diff
@@ -104,7 +125,7 @@ const generateIconNames = (done) => {
 };
 
 gulp.task('generate-icon-names', generateIconNames);
-gulp.task('template', gulp.series('generate-icon-names', generateTemplates(parseTokens(tokens))));
+gulp.task('template', gulp.series('generate-icon-names', generateTemplates(parseUIKitTokens(tokens)), generateSwiftUI(parseSwiftUITokens(tokens))));
```
olliecurtis, line 128: Do we want to rename `generateTemplates` to something else maybe UIKit or ObjC related naming to avoid confusion and future maintenance?

scripts/gulp/spacings.js
```diff
@@ -29,20 +31,24 @@ const VALID_SPACINGS = new Set([
   'icontext',
 ]);
 
-const spacings = (properties, formatName, getLegibleName) => _.chain(properties)
+const filterValidAndMap = (properties, formatName) => _.chain(properties)
```
olliecurtis, line 34: The function name has changed here could we either add some comments above for documentation on what this function does or change the name so its understandable what this does as the new name `filterValidAndMap` is not very descriptive  reading this 'valid', what is valid code, values, numbers?

olliecurtis (changes requested):
Looks good, just a couple of minor documentation/code comments

frugoman:
Will implement the BPK prefix for the tokens and components, suggested by @gert-janvercauteren

gulpfile.js
frugoman: good catch!

Backpack-SwiftUI/Spacing/Classes/Generated/Spacing.swift
frugoman: fyi, I adopted the BPK prefix suggestion!

marianeum (approved):

olliecurtis (approved):
LGTM!

PR #1294, opened 2022-05-13
KOA-5065 BPKButton component

frugoman:
Adding the BPKButton Component to SwiftUI

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


Backpack-SwiftUI/Button/Classes/README.md
```diff
@@ -0,0 +1,14 @@
+# Backpack / Button
+
+## Usage
+
+### SwiftUI
+
+```swift
+import Backpack_SwiftUI
+
+BPKButton("Tap Me!") {
```
marianeum, line 10: might be good to add some more details here - ie what the different props are

Backpack-SwiftUI/Button/Classes/BPKButton.swift
```diff
@@ -0,0 +1,167 @@
...
+    }
+    
+    private func currentState(isPressed: Bool) -> BPKButton.CurrentState {
+        if !enabled { return .disabled }
+        if loading { return .loading }
+        if isPressed { return .highlighted }
+        return .regular
+    }
+    
+    private var isIconOnly: Bool {
```
marianeum, line 86: Maybe this should be more explicit, rather than assumed based on the title availability? The API also feels a little weird having to specify leading/trailing for icon only

How does the accessibility text work? For compose we have created separate constructors for the different types to make the accessibility text required for icon only & make the API more explicit

frugoman: I've addressed this, there's now 2 different inits, there's no way now to create a button without content, either text, text with icon or icon only (with accessibilityLabel required if icon only)

marianeum: I guess technically you could still fake it by providing an empty title (and then end up without an accessibility label)?

marianeum (approved):

marianeum (approved):

PR #1302, opened 2022-05-17
KOA-5101 adding switch component

frugoman:
Adding Switch component to SwiftUI

+ [x] `UNRELEASED.md`
+ [x] `README.md`
+ [x] Tests


Backpack-SwiftUI/Switch/Classes/BPKSwitch.swift
```diff
@@ -0,0 +1,112 @@
...
+    var containerSize: CGSize {
+        self == .large ? CGSize(width: 52, height: 32) : CGSize(width: 40, height: 20)
+    }
+    
+    var thumbSize: CGFloat {
+        self == .large ? 28 : 16
+    }
+}
+
+struct ColoredToggleStyle: ToggleStyle {
```
marianeum, line 56: oof is there no existing switch component in swift ui?

frugoman: There's the Toggle, but it does not satisfy our design demands haha
you can only change the `on` state color, but not the `off` one and you can not change the size of it

Backpack-SwiftUI/Switch/Classes/BPKSwitch.swift
```diff
@@ -0,0 +1,112 @@
...
+    
+    public init(isOn: Binding<Bool>, size: Size = .large) {
+        self._isOn = isOn
+        self.size = size
+    }
+    
+    public var body: some View {
+        Toggle(isOn: $isOn) {
+            Group {}
+        }.toggleStyle(ColoredToggleStyle(size: size))
```
gert-janvercauteren, line 37: Could we move this 1 line down?

Backpack-SwiftUI/Switch/Classes/BPKSwitch.swift
```diff
@@ -0,0 +1,112 @@
...
+    }
+    
+    var containerSize: CGSize {
+        self == .large ? CGSize(width: 52, height: 32) : CGSize(width: 40, height: 20)
+    }
+    
+    var thumbSize: CGFloat {
+        self == .large ? 28 : 16
+    }
+}
```
gert-janvercauteren, line 54: None of this could be 'standardized' into Backpack spacing tokens I would assume?

Backpack-SwiftUI/Switch/Classes/BPKSwitch.swift
```diff
@@ -0,0 +1,112 @@
...
+        HStack {
+            if isOn {
+                Spacer()
+            }
+            Circle()
+                .frame(width: size, height: size)
+                .foregroundColor(.white)
+                .shadow(.sm)
+                .animation(.easeOut(duration: 0.2))
+                .padding(2)
```
gert-janvercauteren, line 84: Could we do .sm / 2 ?

Backpack-SwiftUI/Switch/Classes/BPKSwitch.swift
marianeum: It might be worth double checking with Adam as we generally said we want to rely on system components where possible - particularly for things like switches etc
Is the existing switch component configurable this way?

gert-janvercauteren (commented):
Could you please add:
- a README.md for the component
- a snapshot UI test for the docs?

Backpack-SwiftUI/Switch/Classes/BPKSwitch.swift
frugoman: the existing one (BPKSwitch for UIKit) is not configurable (apart from the on state color).
I'll ask Adam about it!

marianeum: yeah if the existing one isn't configurable i wouldn't go through the effort of configuring this one as it then won't match the uikit one

Backpack-SwiftUI/Switch/Classes/BPKSwitch.swift
frugoman: This was removed

Backpack-SwiftUI/Switch/Classes/BPKSwitch.swift
```diff
@@ -0,0 +1,54 @@
...
+    
+    public init(isOn: Binding<Bool>) {
+        self._isOn = isOn
+    }
+    
+    public var body: some View {
+        Toggle(isOn: $isOn) {
+            Group {}
+        }
+        .labelsHidden()
```
marianeum, line 32: one thing we've done on android (for all toggleable controls) is provide a version with the label (with our text style) and then another which adds a content slot. 
the main reason for that is that on android you generally tap the whole row, rather than just the toggle to switch mode and it allows us to set accessibility rules on the whole content, rather than just the toggle. not sure if that's similar on iOS or if you generally only tap the switch to toggle

frugoman: Yes, actually the native Toggle (Switch) on SwiftUI has a Label to treat it as a row. I did it this way to make it work same as it was on UIKit, but adding a version that can receive a label (or any other view) may be a good idea, since it's the way it will actually be used

marianeum (approved):

PR #1324, opened 2022-06-09
KOA-4949 Update FloatingPanel from 1.6.6 to 2.5.2

gert-janvercauteren:
Update the floating panel layout to the latest release (2.5.2)

+ [ ] `UNRELEASED.md`
+ [ ] `README.md`
+ [ ] Tests


Backpack/BottomSheet/Classes/BackpackFloatingPanelController.swift
```diff
@@ -87,7 +87,7 @@ private extension BPKFloatingPanelController {
 
             outsideSafeAreaView.topAnchor.constraint(
                 greaterThanOrEqualTo: content.view.topAnchor,
-                constant: layout.insetFor(position: .half) ?? 0
```
gert-janvercauteren, line 90: Not sure what to think of this / do with this....

Backpack/BottomSheet/Classes/BottomSheet.swift
```diff
@@ -213,7 +217,7 @@ public final class BPKBottomSheet: NSObject {
     /// It can be useful, for example, when changing the inner constraints of the `contentViewController`
     /// and bottom sheet needs to be resized to fit the content.
     public func updateLayout() {
-        floatingPanelController.updateLayout()
+        floatingPanelController.invalidateLayout()
     }
```
frugoman, line 221: Is this even needed? shouldn't constraints just work without this?

Backpack/BottomSheet/Classes/BottomSheet.swift
```diff
@@ -179,7 +181,9 @@ public final class BPKBottomSheet: NSObject {
             assertionFailure("present(_:animated:completion:) not compatible with persistent presentation style")
             return
         }
-        if let scrollView = floatingPanelController.scrollView {
+        
+        // ToODO --> WHAT IS THIS?
+        if let scrollView = floatingPanelController.trackingScrollView {
```
frugoman, line 186: why do we have this?
can we just remove it and leave it as """""native"""""? aka: leave the scrollview offset as is

Backpack/BottomSheet/Classes/BottomSheet.swift
```diff
@@ -225,26 +229,20 @@ public final class BPKBottomSheet: NSObject {
...
-        func backdropAlphaFor(position: FloatingPanelPosition) -> CGFloat {
-            return Constants.backdropAlpha
-        }
-    }
-    
     public func floatingPanel(
         _ viewController: FloatingPanelController,
         layoutFor newCollection: UITraitCollection
-    ) -> FloatingPanelLayout? {
+    ) -> FloatingPanelLayout {
```
frugoman, line 235: this was nillable without the need for it?? yeez.

It's a breaking change tho, and we need to treat this as a major release

frugoman (commented):
Looks pretty good, just super small comments/questions.
dont forget the UNRELEASED file

Backpack/BottomSheet/Classes/BottomSheet.swift
gert-janvercauteren: Oh no, we need this for our custom bottom view. I need to remove the todo. woops

Backpack/BottomSheet/Classes/BottomSheet.swift
gert-janvercauteren: If we don't do this, then the 'resize' example stops working

Backpack/BottomSheet/Classes/BottomSheet.swift
gert-janvercauteren: The library has changed the protocol so the layout is no longer optional. 
This should not be consumed by consumers I assume. 

In any case we probably want to go for a major update, because we jumped to v2 of floating panel

frugoman (approved):

PR #1327, opened 2022-06-13
KOA-5165 setup release pipeline

gert-janvercauteren:
Setup release pipeline on CI

+ [ ] `UNRELEASED.md`
+ [ ] `README.md`
+ [ ] Tests

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -1,0 +1,40 @@
+name: PR
+
+on:
+  pull_request:
+    branches: [ main ]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+defaults:
+  run:
+    shell: bash -l {0}
+
+jobs:
+
+  Build:
+    name: Build
+    permissions:
+      statuses: write
+      pull-requests: write
+    uses: ./.github/workflows/_build.yml
+
+  Dependabot:
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: write
+    if: ${{ github.actor == 'dependabot[bot]' }}
+    steps:
+      - name: Fetch Dependabot metadata
+        id: dependabot-metadata
+        uses: dependabot/fetch-metadata@v1.3.1
+        with:
+          github-token: "${{ secrets.GITHUB_TOKEN }}"
+      - name: Add bpk label
+        if: contains(steps.dependabot-metadata.outputs.dependency-names, 'bpk-')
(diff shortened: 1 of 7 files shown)


olliecurtis (approved):
LGTM

marianeum (commented):
Nice! I think this covers the whole release process on CI - is everything else around this already set up? (eg environment + secrets, access rights)
Also does iOS need to have the versioning in code? That was one thing we had to remove on Android. You'll also have to create a release of the latest tag otherwise it'll generate release notes from the start of the repository

.github/workflows/pr.yml
```diff
@@ -0,0 +1,40 @@
...
+jobs:
+
+  Build:
+    name: Build
+    permissions:
+      statuses: write
+      pull-requests: write
+    uses: ./.github/workflows/_build.yml
+
+  Dependabot:
```
gert-janvercauteren, line 24: remove

marianeum: The reason this exists is so that updates to the backpack JS updates are included in the release notes as js updates are excluded by default. This is particularly useful for the svg updates for example

.github/workflows/release.yml
```diff
@@ -0,0 +1,77 @@
...
+      run: brew install jq
+    
+    - name: Install cocoapods
+      run: gem install cocoapods
+
+    - name: Check version in Podspec
+      run: |
+        set -eo pipefail
+        echo Target version: ${{ github.event.inputs.tag }}
+        make check-version VERSION=${{ github.event.inputs.tag }}
```
gert-janvercauteren, line 39: update podspecs.

frugoman:
> Nice! I think this covers the whole release process on CI - is everything else around this already set up? (eg environment + secrets, access rights) Also does iOS need to have the versioning in code? That was one thing we had to remove on Android. You'll also have to create a release of the latest tag otherwise it'll generate release notes from the start of the repository

Yes, it needs to be in the podspec files, it's not optional!

frugoman (approved):

.github/workflows/pr.yml
gert-janvercauteren: _keep_

gert-janvercauteren:
Guess I was wrong :D. 
We can use environment variables to push in the version when running a release! No need for extra complexity

UNRELEASED.md
```diff
@@ -1,29 +0,0 @@
-# Unreleased
-> Place your changes below this line.
-
-## How to write a good changelog entry
```
marianeum, line 4: You might also want to update the changelog file to point to the releases

marianeum (commented):
do you plan to clean up the old release flow separately?

PR #1335, opened 2022-06-16
KOA-5168 Install cocoapods from bundle

gert-janvercauteren:
Run bundle install, so we can use our dependencies when releasing docs

+ [ ] `README.md`
+ [ ] Tests

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -27,18 +27,18 @@ jobs:
       with:
         ref: main
     
-    - name: Install cocoapods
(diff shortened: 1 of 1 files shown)


frugoman (approved):
Why not using bundle? it would inforce that the version on CI is the same than locally

gert-janvercauteren:
> Why not using bundle? it would inforce that the version on CI is the same than locally

This PR will use bundle (kind of glossed over it when setting it up at first)

PR #1343, opened 2022-06-22
pod installing before building uikit

frugoman:
We weren't pod installing before building uikit, so in some cases, the project can be out of date.

+ [ ] `README.md`
+ [ ] Tests

### DIFF
--- a/.github/workflows/_build.yml
+++ b/.github/workflows/_build.yml
@@ -175,6 +175,13 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-pods-
 
+      - name: Bundle Install
+        run: bundle install --jobs 4 --retry 3
+
+      - name: Pod Install
+        run: bundle exec pod install
+        working-directory: Example
+
       - name: Build and Test
         run: ./scripts/ci ${{ matrix.variant }}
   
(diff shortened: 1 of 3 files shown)


.github/workflows/_build.yml
```diff
@@ -175,6 +175,13 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-pods-
 
+      - name: Bundle Install
```
gert-janvercauteren, line 178: We probably also want to do npm install, because we're caching the node_modules down below.

.github/workflows/_build.yml
```diff
@@ -35,108 +32,63 @@ jobs:
...
-        run: "pkill Simulator || xcrun simctl erase all"
+      - name: nvm install
+        run: nvm install
 
-      - name: Pod Install
-        run: bundle exec pod install
-        working-directory: Example
+      - name: NPM ci
+        if: steps.cache_node_modules.outputs.cache-hit != 'true'
+        run: nvm use && npm ci
```
gert-janvercauteren, line 60: Should this cache hit check be on the step below? What does npm ci do?

frugoman: `npm ci` install dependencies respecting package-lock.json, installing those exact versions.

This needs to run before running any npm command, so needs to be done here, as the very next step is running `npm run build`

the output of `npm ci` is the same as the cache_node_modules thing would cache

PR #1344, opened 2022-06-22
KOA-5354 Enable shadowless cards

frugoman:
Cards without shadow
Given that the apis of SwiftUI and UIKit differ, the solution is different on the 2 systems, with SwiftUI being the preferred one.

+ [x] `README.md`
+ [x] Tests


Backpack-SwiftUI/Card/Classes/BPKCard.swift
```diff
@@ -43,7 +43,7 @@ public struct BPKCard<Content: View>: View {
         }
     }
 
-    private var elevation: Elevation
+    private var elevation: Elevation?
```
gert-janvercauteren, line 46: Is there a specific reason to make the elevation nullable?
If the user doesn't provide an elevation, we will fallback to 'default'.

frugoman: I thought about it like if you want no elevation, this could be represented as nil, as a lack of being a shadow.

gert-janvercauteren: I feel it might be better to be explicit about it.
considering:
- Not supplying = .default
- Making it nil = no shadow
- Passing in .none = no shadow

If we make it non nullable, then we fallback to default and people supply .none to hide the shadow.

Backpack-SwiftUI/Shadow/Classes/Generated/BPKShadow.swift
```diff
@@ -53,3 +53,14 @@ public extension View {
         )
     }
 }
+
+extension View {
+    @ViewBuilder
+    func shadow(_ shadow: BPKShadow?) -> some View {
```
frugoman, line 59: This is internal on purpose

gert-janvercauteren (commented):
Code looks good!

Please update the README.md for both SwiftUI and UIKit components :D

gert-janvercauteren (approved):
LGTM

PR #1345, opened 2022-06-22
always run pod and bundle install

frugoman:
In some cases the cache is not correct and installation needs to happen, it will be fast though, since useful dependencies would be cached (but reinstalled)

+ [ ] `README.md`
+ [ ] Tests


gert-janvercauteren (commented):
When are we seeing this issue? I would assume a 'cache hit' means that the lockfiles matched, so no 'need' for installing the dependencies?

PR #1393, opened 2022-08-18
Add skeleton loading component

unclehuahua:
https://www.figma.com/file/yN0hFyZlKL0Jwbpi0rEKYT/Backpack-Beta?node-id=1338%3A12364
Slack:  #tmp-loading-skeleton-backpack

- Add Skeleton loading component(Image, headline, bodytext, circle, shimmer) to BPK-UIKit
- Add Skeleton loading component(Image, headline, bodytext, circle, shimmer) to BPK-SwiftUI

https://user-images.githubusercontent.com/13347837/186146609-e0c48750-3e5f-4d88-8851-29d24ee38b98.mp4

+ [x] `README.md`
+ [x] Tests
+ [x] Docs (either update [backpack-docs](https://github.com/Skyscanner/backpack-docs) now, or create a follow up ticket) : https://gojira.skyscanner.net/browse/PAN-2175


Backpack/Skeleton/Classes/BPKCommonSkeleton.swift
```diff
@@ -0,0 +1,44 @@
...
+
+public class BPKCommonSkeleton: UIView {
+
+    public override init(frame: CGRect) {
+        super.init(frame: frame)
+        setupSubviews()
+    }
+    
+    public required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
```
frugoman, line 30: any reason why not to support this init? I think we can just call super and then setupSubviews, no?
```swift
public required init?(coder: NSCoder) {
  super.init(coder: coder)
  setupSubviews()
}
```

Example/Backpack/SwiftUI/Components/Skeleton/SkeletonExampleView.swift
```diff
@@ -0,0 +1,124 @@
...
+                    .foregroundColor(.textPrimaryColor)
+            }
+            Spacer()
+        }
+        .padding()
+    }
+    
+    private func makeBodytextSkeletonRow() -> some View {
+//        HStack(alignment: .bottom) {
+//            Spacer()
```
frugoman, line 113: can you remove this commented code?

Backpack/Skeleton/Classes/BPKSkeleton.swift
```diff
@@ -0,0 +1,175 @@
...
+        }
+        if type == .circle && size != .none {
+            skeletonView.layer.cornerRadius = viewSize.height / 2.0
+        }
+        if type == .circle && size == .none {
+            skeletonView.layer.cornerRadius = min(customSize.height, customSize.width) / 2.0
+        }
+    }
+
+    public static func startShimmer(view: UIView) {
```
frugoman, line 125: Just fully realised this.
Can we make the skeleton view shimmer automatically? The use of a static method for this is not optimal also, since it can be called several times adding layers on top of other layers repeatedly.

I think this layer and animations could be part of the BPKSkeleton itself and then have an instance method (non-static) to set the skelenton's frame if needed

```swift
class BPKSkeleton {
  func setup() {
     startShimmer()
  }
  func startShimmer() {
    self.gradientLayer = makeLayer() // all the logic here
    self.layer.addSublayer(gradientLayer) // Called on self, aka the BPKSkeleton
  }

  public func updateSize(frame) {
    self.gradientLayer.frame = frame
  }
}
```

frugoman: using layoutsubviews to update the layer's frame is recommended, because it will automatically honour the skeleton's size, set from autolayouts on the user's side.

frugoman (changes requested):
Hey, I think we can improve the way UIKit's skeleton works, dropping the need for static methods

Backpack/Skeleton/Classes/BPKSkeleton.swift
unclehuahua: I see, I did this previously, but it not meet the requirement. I did this before and it looks like this: 

https://user-images.githubusercontent.com/13347837/186140640-bd9ade76-883b-4e9f-9b5a-309a2c034b4d.mp4


Because this way will add shimmer animation to every Skeleton instance, image skeleton has a shimmer animation, circle has a shimmer animation, etc.

But UI requirement is only 1 animation in the parent view. Here is the [Figma ](https://www.figma.com/proto/SeFyp0yJXEFUVAlAmrU7xc/Skeleton-Loader?page-id=2%3A282&node-id=728%3A6271&viewport=1462%2C31%2C1.8&scaling=contain&starting-point-node-id=4%3A5262)

frugoman: ah, indeed I didn't know that was the requirement!
My suggestion would not work as intended then

Backpack-SwiftUI/Skeleton/Classes/BPKSkeleton.swift
```diff
@@ -0,0 +1,131 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import SwiftUI
+import Backpack_Common
+
+public struct BPKSkeleton: View {
+    private var type: BPKSkeleton.ViewType = .image
+    private var size: BPKSkeleton.Size = .none
```
frugoman, line 24: This could be simplified, adding a `customSize(CGSize)` case on the Size enum. we won't need an extra `customSize` property here plus the additional init won't have to exist!

logic down this struct will be simplified also

Backpack-SwiftUI/Skeleton/Classes/BPKSkeletonStyle.swift
```diff
@@ -0,0 +1,34 @@
...
+        case `default`, rounded
+    }
+}
+
+internal extension BPKSkeleton.Style {
+    var cornerRadius: CGFloat {
+        switch self {
+        case .rounded:
+            return BPKCornerRadius.sm.value
+        default:
```
frugoman, line 30: could this be like this instead?
```swift
case .`default`:
```

Backpack-SwiftUI/Skeleton/Classes/BPKSkeleton.swift
unclehuahua: Good catch, this is more Swifty! You are really an iOS expert!

frugoman: ☺️

Backpack-SwiftUI/Skeleton/Classes/BPKSkeletonStyle.swift
frugoman: It's more a nice to have, just in case we add more cases in the future and we dont 'forget' to handle this switch

Backpack-SwiftUI/Skeleton/Classes/BPKShimmer.swift
```diff
@@ -0,0 +1,78 @@
...
+ * limitations under the License.
+ */
+
+import SwiftUI
+
+struct BPKShimmer: ViewModifier {
+    @State private var phase: CGFloat = 0
+    var duration = 1.0
+    var bounce = false
+    var delay = 0.2
```
gert-janvercauteren, line 25: Are these adjustable by the end user? I feel like we rather make these private let?

Backpack-SwiftUI/Skeleton/Classes/BPKSkeleton+Size.swift
```diff
@@ -0,0 +1,64 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import SwiftUI
+
+internal extension BPKSkeleton.Size {
```
gert-janvercauteren, line 21: Instead of using BPKSpacing, should we consider using 'hardcoded' values?

Backpack-SwiftUI/Skeleton/Classes/BPKSkeleton.swift
```diff
@@ -0,0 +1,103 @@
...
+    private var type: BPKSkeleton.ViewType = .image
+    private var size: BPKSkeleton.Size = .default
+    private var style: BPKSkeleton.Style = .default
+    private var color: Color = Color(.skyGrayTint06.darkVariant(.blackTint02))
+
+    @State private var phase: CGFloat = 0
+    let duration = 1.0
+    let delay = 0.2
+    
+    public init(
```
gert-janvercauteren, line 32: Could we make this init private? This will probably 'force' consumers to use the static functions in the extension.

(Please also add a comment that says this above the init)
"Please use the static functions to get the preferred skeleton" (or something along those lines)

Backpack-SwiftUI/Skeleton/Classes/BPKSkeleton.swift
```diff
@@ -0,0 +1,103 @@
...
+ */
+
+import SwiftUI
+import Backpack_Common
+
+public struct BPKSkeleton: View {
+    private var type: BPKSkeleton.ViewType = .image
+    private var size: BPKSkeleton.Size = .default
+    private var style: BPKSkeleton.Style = .default
+    private var color: Color = Color(.skyGrayTint06.darkVariant(.blackTint02))
```
gert-janvercauteren, line 26: Can all of these be `let`

Backpack-SwiftUI/Skeleton/Classes/BPKSkeleton.swift
```diff
@@ -0,0 +1,103 @@
...
+
+    @State private var phase: CGFloat = 0
+    let duration = 1.0
+    let delay = 0.2
```
gert-janvercauteren, line 30: Are these unused?

Backpack-SwiftUI/Skeleton/Classes/BPKSkeleton.swift
```diff
@@ -0,0 +1,103 @@
...
+               .clipped()
+            
+        case .circle:
+            Circle()
+               .fill(color)
+               .frame(width: size.circle.width, height:  size.circle.height)
+               .clipped()
+            
+        case .bodytext:
+            VStack(alignment: .leading, spacing: BPKSpacing.sm.value * 2.5) {
```
gert-janvercauteren, line 64: Would a normal BPKSpacing.md be sufficient? (it's x2 of BPKSpacing.sm)

Backpack-SwiftUI/Skeleton/Classes/BPKSkeleton.swift
```diff
@@ -0,0 +1,103 @@
...
+                makeTextRow(size: size.bodytext, multiplier: 0.86)
+                makeTextRow(size: size.bodytext, multiplier: 1.0)
+                makeTextRow(size: size.bodytext, multiplier: 0.57)
+            }
+        }
+    }
+    
+    private func makeTextRow(size: CGSize, multiplier: Double) -> some View {
+        let rowHeight = abs(size.height - BPKSpacing.sm.value * 5) / 3.0
+        return RoundedRectangle(cornerRadius: BPKCornerRadius.xs.value / 2)
```
gert-janvercauteren, line 74: Can we just use hardcoded values for this? Doing these complex mathematics on BPKSpacing loses all the value we gain from them

Backpack-SwiftUI/Skeleton/Classes/BPKSkeletonViewType.swift
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+extension BPKSkeleton {
+    public enum ViewType {
+        case image, headline, circle, bodytext
+    }
+}
```
gert-janvercauteren, line 23: Can we make these internal, so people need to use the static functions of the Skeleton?

Backpack/Skeleton/Classes/BPKShimmerLayer.swift
```diff
@@ -0,0 +1,22 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Foundation
+
+@objc
+public class BPKShimmerLayer: CAGradientLayer { }
```
gert-janvercauteren, line 22: Can we add a comment stating this custom class was made to make sure we don't remove 'default' CAGradientLayers when cleaning up the shimmer?

Backpack/Skeleton/Classes/BPKSkeleton.swift
```diff
@@ -0,0 +1,162 @@
...
+        setup()
+    }
+    
+    public override init(frame: CGRect) {
+        super.init(frame: frame)
+    }
+    
+    public required init?(coder: NSCoder) {
+        super.init(coder: coder)
+    }
```
gert-janvercauteren, line 71: Should these call setup as well? If we don't call setup, then we might miss setting up the skeleton.

Backpack/Skeleton/Classes/BPKSkeleton.swift
```diff
@@ -0,0 +1,162 @@
...
+        BPKColor.dynamicColor(withLightVariant: BPKColor.white, darkVariant: BPKColor.black)
+    private var customSize: CGSize = .zero
+    
+    public var type: BPKSkeletonType = .image {
+        didSet {
+            updateType()
+        }
+    }
+    
+    public var size: BPKSkeletonSize = .default
```
gert-janvercauteren, line 36: Do we need to call updateStyle when the size is changed?

Backpack/Skeleton/Classes/BPKSkeleton.swift
```diff
@@ -0,0 +1,162 @@
...
+        switch self.type {
+        case .bodytext:
+            self.skeletonView = BPKTextSkeleton(size: viewSize)
+            self.skeletonView.backgroundColor = .clear
+        default:
+            self.skeletonView = BPKCommonSkeleton()
+        }
+    }
+    
+    internal func setup() {
```
gert-janvercauteren, line 83: Can this be private? What is the reason to make it internal?

Backpack/Skeleton/Classes/BPKSkeletonSize.swift
```diff
@@ -0,0 +1,69 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Foundation
+import UIKit
+
+public enum BPKSkeletonSize: Equatable {
```
gert-janvercauteren, line 22: Something for @frugoman, but I'm wondering if it would make sense to move this configuration to Backpack common, as both packages need this? (And we risk having different configs across UIKit / SwiftUI)

gert-janvercauteren (commented):
Looks good! Just some tiny comments around how things are setup.

Could you please add a ScrollView { } in the SwitUI example view? (This will make it look a little nicer in the demo app :D

Backpack-SwiftUI/Skeleton/Classes/BPKSkeleton.swift
unclehuahua: Nope, the spacing of UI design is 10, not 8, I can change it to `10`
image

Backpack-SwiftUI/Skeleton/Classes/BPKShimmer.swift
unclehuahua: I think yes, I exposed it, see: https://github.com/Skyscanner/backpack-ios/pull/1393/files#diff-38c5c9da4eeebd536cbba61381c3f23868e1fd2f9deb5a663fb7f690a4c590afR70

User can modify this by `.shimmering(animated: , duration: , delay:, bounce: ) `

gert-janvercauteren: Should we allow this customization? I feel like we want to control how the shimmering works
@frugoman ?

unclehuahua: Yea, I can also make it as unchangeable, it's up to you.

Backpack/Skeleton/Classes/BPKSkeletonSize.swift
frugoman: This is a good idea, although not sure how to implement it correctly so far, so I think accepting this as is on this PR is ok. we can migrate it later as a task for Koala

Backpack-SwiftUI/Skeleton/Classes/BPKShimmer.swift
frugoman: I think this should not be adjustable by the end user, only by Backpack, so that's consistent all across the product. Lets make this unchangeable

Backpack/Skeleton/Classes/BPKSkeleton.swift
frugoman: I think yes.
Also not sure if a convenience init is a good choice here. having the designited initializer with default values may be cleaner and remove the need to have 3 inits defined

```swift
public init(
  type: BPKSkeletonType = .image,
  size: BPKSkeletonSize = .default,
  style: BPKSkeletonStyle = .default
) {
  self.type = type
  self.size = size
  self.style = style
  updateType() // <--- Should this method be called inside setup()? probably yes
  setup()
  super.init(frame: frame)
}

public required init?(coder: NSCoder) {
  super.init(coder: coder)
  updateType()
  setup()
}

**Review by frugoman — APPROVED** [review]

File: Backpack/Skeleton/Classes/BPKCommonSkeleton.swift [hunk 19]
```diff
@@ -0,0 +1,45 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Foundation
+import UIKit
+
+public class BPKCommonSkeleton: UIView {
```
gert-janvercauteren, line 22: Can we remove the public from this class? We do not want to expose the common skeleton to our consumers

Backpack-SwiftUI/Skeleton/Classes/BPKShimmer.swift
```diff
@@ -0,0 +1,72 @@
...
+ * limitations under the License.
+ */
+
+import SwiftUI
+
+struct BPKShimmer: ViewModifier {
+    @State private var phase: CGFloat = 0
+    let duration = 1.0
+    let bounce = false
+    let delay = 0.2
```
gert-janvercauteren, line 25: Please mark these as private.

Backpack-SwiftUI/Skeleton/Classes/BPKSkeleton.swift
```diff
@@ -0,0 +1,99 @@
...
+ * limitations under the License.
+ */
+
+import SwiftUI
+import Backpack_Common
+
+public struct BPKSkeleton: View {
+    private var type: BPKSkeleton.ViewType = .image
+    private var size: BPKSkeleton.Size = .default
+    private var style: BPKSkeleton.Style = .default
```
gert-janvercauteren, line 25:
```suggestion
    private let type: BPKSkeleton.ViewType
    private let size: BPKSkeleton.Size
    private let style: BPKSkeleton.Style
```

We're setting these values in the initializer, there is no need to 'pre-set' them here. 
As a nice side-effect, we can now change them to let :D

Backpack/Skeleton/Classes/BPKSkeleton.swift
```diff
@@ -0,0 +1,166 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Foundation
+import UIKit
+import SwiftUI
+
```
gert-janvercauteren, line 22:
```suggestion
enum BPKSkeletonConstants {
    static let backgroundColor = BPKColor.dynamicColor(
        withLightVariant: BPKColor.skyGrayTint06,
        darkVariant: BPKColor.blackTint02
    )
    
    static let defaultColor = BPKColor.dynamicColor(
        withLightVariant: BPKColor.white,
        darkVariant: BPKColor.black
    )
}
```

And then use BPKSKeletonConstants.defaultColor where we would use BPKSkeleton.defaultColor / same for the bgColor.

Backpack/Skeleton/Classes/BPKSkeleton.swift
```diff
@@ -0,0 +1,166 @@
...
+public class BPKSkeleton: UIView {
+    static internal var bgColor: UIColor =
+        BPKColor.dynamicColor(withLightVariant: BPKColor.skyGrayTint06, darkVariant: BPKColor.blackTint02)
+    static internal var defaultColor: UIColor =
+        BPKColor.dynamicColor(withLightVariant: BPKColor.white, darkVariant: BPKColor.black)
```
gert-janvercauteren, line 27: Please remove this in favour of the above enum

Backpack/Skeleton/Classes/BPKSkeleton.swift
```diff
@@ -0,0 +1,166 @@
...
+            skeletonView.widthAnchor.constraint(equalToConstant: viewSize.width),
+            skeletonView.heightAnchor.constraint(equalToConstant: viewSize.height)
+        ])
+    }
+
+    public static func startShimmer(view: UIView) {
+        view.layer.sublayers?.filter { $0 is BPKShimmerLayer }.forEach { $0.removeFromSuperlayer() }
+        
+        let gradientLayer = BPKShimmerLayer()
+        let transperant = BPKSkeleton.defaultColor.withAlphaComponent(0).cgColor
```
gert-janvercauteren, line 118:
```suggestion
        let transparent = BPKColor.clear.cgColor
```
Tiny typo here

gert-janvercauteren (commented):
Some tiny nitpicks to clean up the API.

Backpack/Skeleton/Classes/BPKSkeleton.swift
unclehuahua: Yea, the typo haha, my bad, but the color seems I cannot modify it. I got this with clear color:

https://user-images.githubusercontent.com/13347837/186820895-33426f54-3ecb-4a67-b1ba-0df798206021.mp4

Backpack-SwiftUI/Skeleton/Classes/BPKSkeleton.swift
unclehuahua: Perfect catch! Thank you! sorry for my un-pro swift!

gert-janvercauteren (approved):
Looks good to me, thanks for your contribution 🙇

PR #1420, opened 2022-09-14
[BPKSkeleton] Add removeShimmer method and update bodytext's size.

unclehuahua:
### Changes

 - Add `removeShimmer` method
 - Update bodytext's height to 40

+ [x] `README.md`
+ [x] Tests


gert-janvercauteren (approved):
Looks good to me

Backpack/Skeleton/README.md
```diff
@@ -32,4 +32,5 @@ let bodytext = BPKSkeleton(type: .bodytext)
 let custom = BPKSkeleton(type: .bodytext, size: .custom(size: CGSize(width: 260, height: 60)))
 
 BPKSkeleton.startShimmer(view: yourView) // Make sure call this after yourView rendered.
+BPKSkeleton.removeShimmer(view: yourView)
```
frugoman, line 35: Can you add a small comment here similar to the one above saying that you should call this when your data's came back or something so that's clear when and why to call this `removeShimmer` method?

unclehuahua: I just imagine that the user might need this, this method is not required anything, I think the method name has explained all. It don't need `when or other things`, since I have checked `if shimmer exist, then remove`.
I think that people might need this method after added shimmer animation. For example, I add shimmer to a hotel card, but now I want to remove the shimmer, so I can call this?

frugoman: 👍

frugoman (approved):

PR #1459, opened 2022-10-24
[KOA-5373] assertSnapshot helper function to use in all tests

frugoman:
KOA-5373 - Epic

created a helper function for taking and comparing snapshots. To be used by all tests

### DIFF
--- a/Backpack/Tests/SnapshotTests/Utils/AssertSnapshot.swift
+++ b/Backpack/Tests/SnapshotTests/Utils/AssertSnapshot.swift
@@ -24,12 +24,12 @@ import SnapshotTesting
 func assertSnapshot(
     _ view: UIView,
     modes: [UIUserInterfaceStyle] = [.light, .dark],
+    isRecording: Bool = false,
     file: StaticString = #file,
     testName: String = #function,
     line: UInt = #line
 ) {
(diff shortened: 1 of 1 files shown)


Backpack/Tests/SnapshotTests/Utils/AssertSnapshot.swift
```diff
@@ -0,0 +1,54 @@
...
+
+func assertSnapshot(
+    _ view: UIView,
+    modes: [UIUserInterfaceStyle] = [.light, .dark],
+    file: StaticString = #file,
+    testName: String = #function,
+    line: UInt = #line
+) {
+    let identifier: (UIUserInterfaceStyle) -> String = { mode in mode == .dark ? "dark-mode" : "light-mode" }
+    isRecording = false
```
gert-janvercauteren, line 32: Could we let control of isRecording reside in the testing files? This would give us a finer control when re-recording components.
It also follows the docs for the SnapshotTesting library closer.

frugoman: I'll add it as a parameter to the function, with a default value of `false`

gert-janvercauteren (approved):

PR #1463, opened 2022-10-24
[KOA-5645] Fix iOS docs site

frugoman:
KOA-5645
Docs were not being generated in the right folder, nor with the right parameters. It's fixed now!

The tool is not picking up the """""enums""""" that we do in obj-c, like `BPKSPacing.h` because they're not really enums.. we should migrate those over to proper objc enums or just move them to swift enums!! but that's a different change to this

### DIFF
--- a/scripts/build-docs
+++ b/scripts/build-docs
@@ -40,6 +40,7 @@ function buildDocsHtmlFiles() {
   mkdir -p "$outputDir/versions/latest"
 
   buildRedirectHTML "$outputDir/404.html"
+  buildRedirectHTML "$outputDir/index.html"
   buildLatestIndex $version
 }
 
--- a/scripts/build-docs
+++ b/scripts/build-docs
@@ -40,7 +40,6 @@ function buildDocsHtmlFiles() {
   mkdir -p "$outputDir/versions/latest"
 
   buildRedirectHTML "$outputDir/404.html"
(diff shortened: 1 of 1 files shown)


scripts/build-docs
```diff
@@ -3,35 +3,35 @@
...
+  sourcekitten doc -- -workspace $EXAMPLE_WORKSPACE -scheme "Backpack-SwiftUI" >swiftDoc-swiftui.json
+  jazzy --sourcekitten-sourcefile swiftDoc-swiftui.json -o "$outputDir/versions/$version/swiftui"
   rm swiftDoc-swiftui.json
 }
 
 function buildRedirectHTML() {
   echo '<!DOCTYPE html>
   <html>
-    <meta http-equiv="refresh" content="0;url=/ios/versions/latest" />
+    <meta http-equiv="refresh" content="0;url=/ios/versions/latest/index.html" />
```
gert-janvercauteren, line 27: Will this redirect to swiftui then? Or does this need an extra swiftui or uikit in the url?

frugoman: This will redirect to the latest/index.html, which all it does itself is redirect to swiftui as default. I guess we want that to have both a 'latest' url you can go to, to always use the latest version, but also we're using this latest page for when we go to a non existant url

gert-janvercauteren (approved):

PR #1467, opened 2022-10-26
[KOA5645] Copying docs to latest folder

frugoman:
[KOA5645] Copying docs to latest folder to be able to navigate to a specific file from the docsite

### DIFF
--- a/scripts/build-docs
+++ b/scripts/build-docs
@@ -40,6 +40,7 @@ function buildDocsHtmlFiles() {
   mkdir -p "$outputDir/versions/latest"
 
   buildRedirectHTML "$outputDir/404.html"
+  buildRedirectHTML "$outputDir/index.html"
   buildLatestIndex $version
 }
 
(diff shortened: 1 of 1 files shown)


scripts/build-docs
```diff
@@ -44,6 +44,10 @@ function buildDocsHtmlFiles() {
   buildLatestIndex $version
 }
 
+function copyDocsToLatest() {
+  cp -r "$outputDir/versions/$version/" "$outputDir/versions/latest"
+}
+
```
olliecurtis, line 50: You will need to also remove the steps for the`index.html` above otherwise it will still redirect to the place you don't want :)

olliecurtis (changes requested):

scripts/build-docs
olliecurtis: Do we want to keep that though or just have it go to latest if you just go to the root anyway without a version?

olliecurtis (approved):

PR #1483, opened 2022-11-09
[KOA-5668] Separating releases into different jobs

frugoman:
KOA-5668

Separating the release of the different modules into separate workflows that can execute in parallel, but also be re-runnable in isolation if they fail for any reason

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -15,27 +15,22 @@ jobs:
         statuses: write
         pull-requests: write
       uses: ./.github/workflows/_build.yml
-  Deploy:
-    name: Release to Cocoapods
+  Deploy-Commons:
+    name: Release Commons to Cocoapods
     runs-on: macos-11
     environment: Publishing
     needs: [Build]
     steps:
-
     - name: Checkout source code
       uses: actions/checkout@v3
       with:
         ref: main
-    
+
     - name: Set Ruby version
       uses: ruby/setup-ruby@v1 # Sets ruby version via `.ruby-version`
 
     - name: Bundle Install
       run: bundle install --jobs 4 --retry 3
-      
-    - name: Pod repo update
-      run: | 
-        bundle exec pod repo update
 
     - name: Publish Pods - Backpack Common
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -15,27 +15,22 @@ jobs:
         statuses: write
         pull-requests: write
       uses: ./.github/workflows/_build.yml
-  Deploy:
-    name: Release to Cocoapods
+  Deploy-Commons:
```
olliecurtis, line 18: nit: Should this be `Common`?
```suggestion
  Deploy-Common:
```

.github/workflows/release.yml
```diff
@@ -15,27 +15,22 @@ jobs:
...
+    name: Release Commons to Cocoapods
```
olliecurtis, line 19: Same here

```suggestion
    name: Release Common to Cocoapods
```

.github/workflows/release.yml
```diff
@@ -64,6 +93,12 @@ jobs:
         COCOAPODS_TRUNK_TOKEN: ${{ secrets.COCOAPODS_TRUNK_TOKEN }}
         LIB_VERSION: ${{ github.event.release.tag_name }}
 
+  Deploy-Docs:
+    name: Release Backpack Docs
+    runs-on: macos-11
+    environment: Publishing
+    needs: [Deploy-SwiftUI, Deploy-UIKit, Deploy-Commons]
```
olliecurtis, line 100: If the above suggestions go in then this would be

```suggestion
    needs: [Deploy-SwiftUI, Deploy-UIKit, Deploy-Common]
```

olliecurtis (changes requested):

.github/workflows/release.yml
frugoman: yes!

.github/workflows/release.yml
frugoman: yes!

olliecurtis (approved):

PR #1541, opened 2023-01-18
Add elevation property for BpkDividedCard SwiftUI version

iHandle:
**Changes**
* Extract the `Elevation` enum outside of the `BPKCard` as `BPKCardElevation` so that the Divided Card can use it also.
* Add `elevation` property for BpkDividedCard SwiftUI version.
* Updat unit tests, the example screenshot and the README.md.

+ [x] `README.md`
+ [x] Tests


Backpack-SwiftUI/Card/Classes/BPKDividedCard.swift
```diff
@@ -19,20 +19,23 @@
...
     private let primaryContent: PrimaryContent
     private let secondaryContent: SecondaryContent
     private var tapAction: () -> Void = {}
 
     public init(
+        isElevated: Bool = true,
         @ViewBuilder primaryContent: () -> PrimaryContent,
         @ViewBuilder secondaryContent: () -> SecondaryContent
     ) {
+        self.isElevated = isElevated
```
gert-janvercauteren, line 32: Can you please use the same Enum for the elevation option? 
We can basically 'copy' the way BPKCard does it.

gert-janvercauteren (commented):
Can we update the background color of the page, so we can see the corners of the non elevated card? 
(Same for the screenshot tests)

Backpack-SwiftUI/Card/Classes/BPKDividedCard.swift
iHandle: Can we just use `BPKCard.Elevation` directly? If we define an elevation Enum inside `BPKDividedCard`, we have to convert `BPKDividedCard.Elevation` to `BPKCard.Elevation`.

gert-janvercauteren: Yes we can use it directly.

iHandle:
> 

If we set the background color as, for example blue, It's hard to tell the difference between elevated card and non elevated card.
image

gert-janvercauteren (approved):
LGTM, thanks for contributing!

PR #1573, opened 2023-02-10
BPKPrice

JLau-Sky:
Implemented BPKPrice as per Figma design.
https://www.figma.com/file/irZ3YBx8vOm16ICkAr7mB3/Backpack?node-id=16470%3A17021&t=hBJOAHgbVmWTKEvi-0

See snapshots below for visuals.


Backpack-SwiftUI/Price/Classes/BPKPrice.swift
```diff
@@ -0,0 +1,121 @@
...
+import SwiftUI
+import Backpack_Common
+
+public struct BPKPrice: View {
+    public enum Size {
+        case large, small
+    }
+    
+    public enum Alignment {
+        case left, right
```
frugoman, line 28: This should be renamed to leading-trailing

Backpack-SwiftUI/Price/Classes/BPKPrice.swift
```diff
@@ -0,0 +1,121 @@
...
+    @Binding public var trailingText: String?
+    @Binding public var alignment: Alignment
+    public let size: Size
+    
+    public init(
+        price: Binding<String>,
+        leadingText: Binding<String?> = .constant(nil),
+        lineThroughText: Binding<String?> = .constant(nil),
+        trailingText: Binding<String?> = .constant(nil),
+        alignment: Binding<Alignment> = .constant(.left),
```
frugoman, line 43: it seems you're not supposed to change this values inside this view, all this parameters can be just `let` without being bindings!

Backpack-SwiftUI/Price/Classes/BPKPrice.swift
```diff
@@ -0,0 +1,121 @@
...
+    private func footnoteOrCaptionFontStyle() -> BPKFontStyle {
+        return size == .large ? .footnote : .caption
+    }
+    
+    private func mapAlignment() -> HorizontalAlignment {
+        switch alignment {
+        case .left:
+            return .leading
+        case .right:
+            return .trailing
```
frugoman, line 105: this is not logically true for RTL languages

Backpack/Price/Classes/BPKPrice.swift
```diff
@@ -0,0 +1,203 @@
...
+    }()
+    
+    private lazy var containerStackView: UIStackView = {
+        let stackView = UIStackView(arrangedSubviews: [topTextStackView, priceStackView])
+        stackView.axis = .vertical
+        return stackView
+    }()
+    
+    public init(
+        price: String,
```
frugoman, line 98: What's expected to be set as price when a viewcontroller's being initialised and the data hasn't come from the backend yet? 
Forcing this in the initialiser will force clients to set dummy values until the data comes.
You can compare this component with the regular `UILabel`'s init

Backpack/Price/Classes/BPKPrice.swift
```diff
@@ -0,0 +1,203 @@
...
+        
+        super.init(frame: .zero)
+        
+        leadingTextLabel.text = leadingText
+        separatorLabel.text = "•"
+        priceLabel.text = price
+        trailingTextLabel.text = trailingText
+        
+        lineThroughTextLabel.text = lineThroughText
+        applyLineThroughStyling()
```
frugoman, line 120: this seems to be part of the setup of the view, can we move this to the `setupView` method?

Backpack/Price/Classes/BPKPrice.swift
```diff
@@ -0,0 +1,203 @@
...
+            priceStackView.axis = .horizontal
+            priceStackView.spacing = BPKSpacingSm
+        case .right:
+            containerStackView.alignment = .trailing
+            
+            if size == .small {
+                priceStackView.axis = .vertical
+                priceStackView.spacing = BPKSpacingNone
+            }
+        }
```
frugoman, line 156: should this be executed every time the properties are set? ie: if I set trailingText after this got executed, this should be handled as well.
I recommend putting this logic inside a `updateLookAndFeel` method, and calling it from here and from the didSet of the exposed properties

Backpack/Tests/SnapshotTests/BPKPriceSnapshotTest.swift
```diff
@@ -0,0 +1,123 @@
...
+            price.leadingAnchor.constraint(equalTo: parentView.leadingAnchor),
+            price.trailingAnchor.constraint(equalTo: parentView.trailingAnchor),
+            price.bottomAnchor.constraint(equalTo: parentView.bottomAnchor)
+        ])
+        
+        return parentView
+    }
+    
+    // swiftlint:disable function_parameter_count
+    private func createPrice(
```
frugoman, line 66: is this method needed? it's just forwarding parameters without having default values, lets remove it if this is the case

Example/Backpack/SwiftUI/Components/Price/PriceExampleView.swift
```diff
@@ -0,0 +1,80 @@
...
+            }
+            .padding([.leading, .trailing], BPKSpacing.base.value)
+        }
+    }
+    
+    private func makePrice(
+        _ price: String = "£1830",
+        lineThroughText: String? = "£2033",
+        leadingText: String? = "App only deal",
+        trailingText: String? = "per day",
```
frugoman, line 62: nit-ish: 
this are not being set in any of the cases, could we have this values as `@State` in this view to showcase how they would be used in a more real-life case?

Example/Backpack Screenshot/Screenshots.swift
```diff
@@ -435,5 +435,10 @@ class Screenshots: BackpackSnapshotTestCase {
             saveScreenshot(component: "page indicator", scenario: "default", userInterfaceStyle: userInterfaceStyle)
             tapBackButton()
         }
+        
+        navigate(title: "Price") {
```
frugoman, line 439: can you add one like this for the SwiftUI screenshots?

frugoman (changes requested):

Backpack/Price/Classes/BPKPrice.swift
```diff
@@ -0,0 +1,203 @@
...
+    private lazy var strikeThroughTextAttributes: [NSAttributedString.Key: Any] = {
+        [
+            .foregroundColor: BPKColor.textSecondaryColor,
+            .font: BPKFont.makeFont(fontStyle: footnoteOrCaptionFontStyle()),
+            .strikethroughStyle: NSUnderlineStyle.single.rawValue,
+            .strikethroughColor: BPKColor.textSecondaryColor
+        ]
+    }()
+    
+    private lazy var priceLabel: BPKLabel = {
```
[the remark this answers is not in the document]
JLau-Sky: Done

Backpack-SwiftUI/Price/Classes/BPKPrice.swift
```diff
@@ -0,0 +1,121 @@
...
+        self.size = size
+    }
+    
+    public var body: some View {
+        VStack(alignment: mapAlignment(), spacing: BPKSpacing.none.value) {
+            HStack(spacing: BPKSpacing.sm.value) {
+                if let lineThroughText = lineThroughText {
+                    BPKText(lineThroughText, style: footnoteOrCaptionFontStyle())
+                        .strikethrough(true)
+                        .foregroundColor(.textSecondaryColor)
```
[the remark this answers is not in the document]
JLau-Sky: I'm not sure what you mean here.

Backpack-SwiftUI/Price/Classes/BPKPrice.swift
```diff
@@ -28,24 +76,24 @@ public struct BPKPrice: View {
...
+    private let price: String?
+    private let leadingText: String?
+    private let previousPrice: String?
+    private let trailingText: String?
+    private let alignment: Alignment
+    private let size: Size
     
     public init(
-        price: String,
+        price: String?,
```
frugoman, line 87: I think my other comment about the price not having a value was only relevant for UIKit, for SwiftUI handling the value as required if the component requires it would be an expected way to handle it.
@gert-janvercauteren what do you think about this? shall this property be required in SwiftUI and not required in UIKit? (see my previous comment on the uikit class)

Backpack/Price/Classes/BPKPrice.swift
```diff
@@ -154,30 +136,68 @@ public final class BPKPrice: UIView {
...
+        }
+        priceStackView.axis = .horizontal
+        priceStackView.spacing = BPKSpacingSm
+        priceStackView.alignment = .firstBaseline
+    }
+    
+    private func styleContainerStackView() {
+        [topTextStackView, priceStackView].forEach {
+            containerStackView.addArrangedSubview($0)
+        }
```
frugoman, line 191: having this logic inside the style method can be a bit confusing and hard to follow code in the future, please keep the view setup logic in the setupView method.

Backpack/Price/Classes/BPKPrice.swift
```diff
@@ -154,30 +136,68 @@ public final class BPKPrice: UIView {
...
+        topTextStackView.spacing = BPKSpacingSm
+    }
+    
+    private func stylePriceStackView() {
+        [priceLabel, trailingTextLabel].forEach {
+            priceStackView.addArrangedSubview($0)
+        }
+        priceStackView.axis = .horizontal
+        priceStackView.spacing = BPKSpacingSm
+        priceStackView.alignment = .firstBaseline
```
frugoman, line 185: This configuration code was correct before being in the initialisation code of each view. Can we put it back there?
```swift
private let priceStackView: UIStackView = {
  let stackView = UIStackView(arrangedSubviews: [priceLabel, trailingTextLabel])
  stackView.axis = .horizontal
  stackView.spacing = BPKSpacingSm
  stackView.alignment = .firstBaseline
  return stackView
}()
```

frugoman (changes requested):

Backpack-SwiftUI/Price/Classes/BPKPrice.swift
frugoman: I think as it does not make sense to have a price when there's no price, in SwiftUI that should be handled by the client.
ie:
```swift
struct MyScreen: View {
  @ObservedObject let priceViewModel: PriceViewModel
  var body: some View {
    if let price = priceViewModel.price {
       BPKPrice(price: price) 
    }
  }
}
```
I think this is the expected API. It would be nice to support it for UIKit, but UIKit cant achieve this because of how it, and the view controllers lifecycle works

frugoman (approved):

PR #1580, opened 2023-02-21
BPKFloatingNotification

pontus-ekhem:
Contributed the BPKFloatingNotification component

**Figma**: https://www.figma.com/file/irZ3YBx8vOm16ICkAr7mB3/Backpack?node-id=15741%3A16826&t=ndCZwBUc4tqZFhMO-0

LTR
LTR

RTL
RTL


Backpack/FloatingNotification/Classes/BPKFloatingNotification.swift
```diff
@@ -0,0 +1,203 @@
...
+    private var onButtonTap: (() -> Void)?
+    private var onDismissal: (() -> Void)?
+    
+    private var bottomConstraint = NSLayoutConstraint()
+    private let animator = FloatingNotificationAnimator()
+    private var notificationsQueue = [FloatingNotificationViewModel]()
+    
+    // MARK: Subviews
+    private let stackView: UIStackView = {
+        let result = UIStackView()
```
frugoman, line 39: nit: could we rename this `result` into `stack`, `iconView`, `label`, and such?

Backpack/FloatingNotification/Classes/FloatingNotificationViewModel.swift
```diff
@@ -0,0 +1,29 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Foundation
+
+struct FloatingNotificationViewModel {
```
frugoman, line 21: could we make this one public and have clients send an instance of this directly?

in the future (or now) we could have static functions that only receive certain params and give it specific names, like:
```swift
public struct FloatingNotificationViewModel {
   let ... 
   // all the properties..
   public init(...) {...}

   public static func titleOnly(_ title: String) -> FloatingNotificationViewModel {}
   public static func titleWithAction(_ title: String, action: FloatingNotificationAction) -> FloatingNotificationViewModel {}
}

// So when showing a notification, it could look like this from the client

notification.show(.titleOnly("Howdy!"))
```

Backpack/FloatingNotification/Classes/FloatingNotificationViewModel.swift
```diff
@@ -0,0 +1,29 @@
...
+    let parentView: UIView
+    let text: String
+    let buttonTitle: String?
+    let onTap: (() -> Void)?
```
frugoman, line 25: I'd suggest grouping data that needs to go together in a separate type (could be a nested struct here). this way the compiler does not allow inconsistent state to be represented, like creating a button with title but no action, or vice versa.

I'd create a FloatingNotificationViewModel.Action struct, containing a NON optional title and action, and then have the FloatingNotificationViewModel have an optional action.

Backpack/FloatingNotification/Classes/FloatingNotificationAnimator.swift
```diff
@@ -0,0 +1,74 @@
...
+        }
+        isNotificationDisplayed = true
+        upAnimator.startAnimation()
+    }
+    
+    @objc
+    private func prepareDownAnimation() {
+        downAnimator = UIViewPropertyAnimator(duration: animationDuration, curve: .easeInOut)
+        downAnimator.addAnimations { [weak self] in
+            self?.delegate?.downAnimation()
```
frugoman, line 66: what's the expected purpose of this delegate? what's the event that we want to emit to the client exactly (conceptually I mean)?

frugoman: if we go for it, check out my comment on the protocol above

Backpack/FloatingNotification/Classes/FloatingNotificationAnimator.swift
```diff
@@ -0,0 +1,74 @@
...
+
+import Foundation
+
+protocol FloatingNotificationAnimatorDelegate: AnyObject {
+    func upAnimation()
+    func downAnimation()
+    func animationDidFinish()
+}
+
+final class FloatingNotificationAnimator: NSObject {
```
frugoman, line 27: does this need to be an NSObject?

Backpack/FloatingNotification/Classes/BPKFloatingNotification.swift
pontus-ekhem: Yeah sure can do

Backpack/FloatingNotification/Classes/FloatingNotificationViewModel.swift
pontus-ekhem: Yeah sounds good. Had the same init as Android to align but happy to change it to this

Backpack/FloatingNotification/Classes/FloatingNotificationAnimator.swift
pontus-ekhem: Yes, it is needed for the `perform` delay. FYI, the perform delay is used because if we use the animator delay then animation (not visible) starts too soon in the background and does not allow interaction such as the button tap.

Backpack/FloatingNotification/Classes/FloatingNotificationViewModel.swift
pontus-ekhem: Yeah sure can do

Backpack/FloatingNotification/Classes/BPKFloatingNotification.swift
```diff
@@ -0,0 +1,186 @@
...
+        animator.delegate = self
+        alpha = 0
+        backgroundColor = BPKColor.corePrimaryColor
+        layer.cornerRadius = BPKCornerRadiusMd
+        translatesAutoresizingMaskIntoConstraints = false
+        addSubviews()
+        BPKShadow.shadowLg().apply(to: layer)
+    }
+    
+    required init?(coder: NSCoder) {
```
gert-janvercauteren, line 92:
```suggestion
    @available(*, unavailable)
    required init?(coder: NSCoder) {
```

Backpack/FloatingNotification/Classes/BPKFloatingNotification.swift
```diff
@@ -0,0 +1,186 @@
...
+        fatalError("init(coder:) has not been implemented")
+    }
+    
+    private func addSubviews() {
```
gert-janvercauteren, line 96: Could this be renamed to 'setup()'
then we can also include the 
translatesAutoresizingMaskIntoConstraints = false
and apply shadow from the init function

Backpack/FloatingNotification/Classes/BPKFloatingNotification.swift
```diff
@@ -0,0 +1,186 @@
...
+
+    private func setupConstraints(relativeTo parent: UIView) {
+        constraints.forEach { removeConstraint($0) }
+        
+        bottomConstraint = bottomAnchor.constraint(
+            equalTo: parent.safeAreaLayoutGuide.bottomAnchor,
+            constant: Constants.startBottomConstraint
+        )
+        
+        var constraints = [
```
gert-janvercauteren, line 139: Could we do something with the 'specialiist' reading layout anchors? 
What happens when we're on iPad splitview? (Any of the sizes)

Backpack/FloatingNotification/Classes/BPKFloatingNotification.swift
```diff
@@ -0,0 +1,186 @@
...
+    // MARK: Animation & rendering
+    public func show(_ viewModel: FloatingNotificationViewModel) {
+        notificationsQueue.append(viewModel)
+        renderNext()
+    }
+    
+    private func renderNext() {
+        if let nextViewModel = notificationsQueue.first, !animator.isNotificationDisplayed {
+            render(viewModel: nextViewModel)
+            animator.animateUp(hideAfter: nextViewModel.hideAfter)
```
gert-janvercauteren, line 110: When VoiceOver is on, the floating notification should not hide as long as it has focus. 
(Check out what the Youtube app does when you save/unsave a video)

Backpack/FloatingNotification/Classes/BPKFloatingNotification.swift
```diff
@@ -0,0 +1,186 @@
...
+        
+        animator.delegate = self
+        alpha = 0
+        backgroundColor = BPKColor.corePrimaryColor
+        layer.cornerRadius = BPKCornerRadiusMd
+        translatesAutoresizingMaskIntoConstraints = false
+        addSubviews()
+        BPKShadow.shadowLg().apply(to: layer)
+    }
+
```
gert-janvercauteren, line 91: Should we mark the init(frame:) as unavailable as well?

Backpack/FloatingNotification/Classes/BPKFloatingNotification.swift
pontus-ekhem: ok sure but will keep `translatesAutoresizingMaskIntoConstraints = false` within the definition of the components as I find it neater

Backpack/FloatingNotification/Classes/BPKFloatingNotification.swift
pontus-ekhem: Nice one, I'll base it on the max width we accept

frugoman:
would it make sense to treat the notification view as the 'holder' of the view model? so when you call `show` you dont pass parameters, you just call it to show itself.

other approach could be to have a FloatingNotification.show static method, right now it doesn't seem to make much sense to treat the notification as an instance (we can have a quick call about the topic)

Backpack/FloatingNotification/Classes/BPKFloatingNotification.swift
pontus-ekhem: This one was tricky but added it in

Backpack/Carousel/Classes/CarouselPageViewController.swift
```diff
@@ -45,11 +45,11 @@ final class CarouselPageViewController: UIPageViewController {
     
     override func viewDidLoad() {
         super.viewDidLoad()
-        dataSource = self
         delegate = self
     }
     
     func set(images: [UIView], animated: Bool = true) {
+        dataSource = images.count > 1 ? self : nil
         pages.removeAll()
```
pontus-ekhem, line 53: Disabling scrolling for when there is only not image as mentioned here: https://skyscanner.slack.com/archives/C04NDB73D8S/p1676459461414229?thread_ts=1676046304.364759&cid=C04NDB73D8S

frugoman (approved):

PR #1584, opened 2023-02-27
BPKCarousel: Disabling scrolling when setting one image

pontus-ekhem:
As discussed and agreed here: https://skyscanner.slack.com/archives/C04NDB73D8S/p1676459363098889?thread_ts=1676046304.364759&cid=C04NDB73D8S


gert-janvercauteren (approved):
LGTM.

PR #1590, opened 2023-03-02
Don't generate deprecated colors in SwiftUI

gert-janvercauteren:
We are generating deprecated colors in SwiftUI, which is not what we intended. SwiftUI adoption is starting and we want to make sure no one uses these colors by accident. 

+ [ ] `README.md`
+ [ ] Tests


marianeum (commented):
it might be worth not generating static colours at all for swift ui as we should only really use dynamic ones (we ended up doing the same in compose as we just had a handful of static colours left)

gert-janvercauteren:
> it might be worth not generating static colours at all for swift ui as we should only really use dynamic ones (we ended up doing the same in compose as we just had a handful of static colours left)

Fair point. I'll change the PR to totally get rid of them :D

marianeum (approved):

PR #1599, opened 2023-03-09
SONIC-1385: fix missing addSubview when titleView & title are set

dennis-li-skyscanner:
# Changes
- Fix titleView & TitleLabel removeFromSuperView & addSubView logic during updateLookAndFeel
  - When the title is set, the old titleView should be removed 
  - When the titleView is set, the old titleLabel should be removed 
  - When the titleView is set, the old titleView should be removed 

+ [x] `README.md`
+ [x] Tests


Backpack/Rating/Classes/BPKRating.swift
```diff
@@ -234,6 +237,14 @@ public class BPKRating: UIView {
         }
     }
 
+    private func updateTitleSubtitleStackViewSubviews() {
+        let titleViewUpdated = titleSubtitleStackView.arrangedSubviews != [titleView, subtitleLabel]
```
frugoman, line 241: I think this boolean represents if the stack needs to be updated, rather than if it was updated. can you rename it to reflect this?

Backpack/Rating/Classes/BPKRating.swift
```diff
@@ -208,6 +209,8 @@ public class BPKRating: UIView {
 
         updateSize()
         ratingValueLabel.text = ratingScale.displayedValue(from: value)
+
+        updateTitleSubtitleStackViewSubviews()
```
frugoman, line 213: I dont think you need to call this from here, the title and subtitle labels are added at setup time and are never removed

Backpack/Rating/Classes/BPKRating.swift
```diff
@@ -234,6 +237,14 @@ public class BPKRating: UIView {
         }
     }
 
+    private func updateTitleSubtitleStackViewSubviews() {
+        let titleViewUpdated = titleSubtitleStackView.arrangedSubviews != [titleView, subtitleLabel]
+        if titleViewUpdated {
+            titleSubtitleStackView.arrangedSubviews.forEach { $0.removeFromSuperview() }
+            [titleView, subtitleLabel].forEach(titleSubtitleStackView.addArrangedSubview(_:))
+        }
```
frugoman, line 245: I dont quite understand why is this needed, do we really need this?

frugoman (changes requested):

Backpack/Rating/Classes/BPKRating.swift
dennis-li-skyscanner: I updated it in this commit 6a552388575a0253b740462a2c7bf4006b24fab9

PR #1600, opened 2023-03-10
[KOA-5727] Fixing elevated pannels without padding

frugoman:
KOA-5727


gert-janvercauteren (changes requested):
Can we discard the changes to the package.lock?

gert-janvercauteren (approved):

PR #1622, opened 2023-03-30
[KOA-5994] Making the title not required in bpkrating

frugoman:
KOA-5994

### DIFF
--- a/Backpack/Rating/Classes/BPKRating.swift
+++ b/Backpack/Rating/Classes/BPKRating.swift
@@ -43,9 +43,9 @@ public class BPKRating: UIView {
         }
     }
 
-    public var title: String {
+    public var title: String? {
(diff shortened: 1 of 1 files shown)


Backpack/Rating/Classes/BPKRating.swift
```diff
@@ -43,9 +43,9 @@ public class BPKRating: UIView {
         }
     }
 
-    public var title: String {
+    public var title: String? {
```
marianeum, line 46: this just makes the title nullable, but doesn't hide the view if it is - is the spacing still correct everywhere?

gert-janvercauteren: @frugoman There's also a SwiftUI version we need to take care of :D

PR #1626, opened 2023-04-03
[KOA-5078] Adding text field swiftui component

frugoman:
| Day | Night |
| --- | --- |
|iPhone-swiftui_text-field___default_lm|iPhone-swiftui_text-field___default_dm|


Backpack-SwiftUI/Badge/Classes/BPKBadge.swift
```diff
@@ -70,34 +70,6 @@ public struct BPKBadge: View {
     }
 }
 
-// Helper functions to apply the outline style
```
frugoman, line 73: Moving this to its own file so that's reusable!

Backpack-SwiftUI/TextField/Classes/BPKTextField.swift
```diff
@@ -0,0 +1,110 @@
...
+    /// Creates a `BPKTextField`.
+    ///
+    /// - Parameters:
+    ///   - placeholder: The placeholder text to display when the text field is empty.
+    ///   - text: The text to display in the text field.
+    ///   - style: The style of the text field. Defaults to `bodyDefault`.
+    public init(
+        placeholder: String = "",
+        _ text: Binding<String>,
+        style: BPKFontStyle = .bodyDefault
```
gert-janvercauteren, line 44: Do we allow textfields with different fonts?

Backpack-SwiftUI/TextField/Classes/BPKTextField.swift
```diff
@@ -0,0 +1,110 @@
...
+    }
+    
+    public func inputState(_ state: State) -> BPKTextField {
+        var result = self
+        result.state = state
+        return result
+    }
+}
+
+private extension TextField {
```
gert-janvercauteren, line 87: For 'alignment' should we make this 'fileprivate' ?

gert-janvercauteren (commented):
Looking good! Just some tiny remarks. 

Wondering if the documentation would benefit from an example with 'form validation'. (And maybe the example app as well, to jumpstart our consumers.

Example/Backpack/SwiftUI/Components/TextField/TextFieldExampleView.swift
```diff
@@ -0,0 +1,49 @@
...
+import SwiftUI
+import Backpack_SwiftUI
+
+struct TextFieldExampleView: View {
+    @State var text = "Value"
+    
+    var body: some View {
+        ScrollView {
+            VStack {
+                BPKTextField(.constant(""))
```
gert-janvercauteren, line 29: Can we add an example of an empty field, with a placeholder?

Backpack-SwiftUI/TextField/Classes/BPKTextField.swift
```diff
@@ -0,0 +1,106 @@
...
+            .overlay(accessory)
+            .disabled(state.isDisabled)
+    }
+    
+    private var accessory: some View {
+        HStack {
+            Spacer()
+            if let icon = state.icon {
+                if case let .clear(accessibilityLabel, action) = state {
+                    BPKIconView(icon.icon)
```
gert-janvercauteren, line 63: The BPKIconView needs to change how it 'create the image' to use the 'decorative' initializer.

We also need to consider to either:
- Wrap this into a button
- Add a 'trait' of `.isButton`

Backpack-SwiftUI/TextField/Classes/BPKTextField.swift
```diff
@@ -0,0 +1,106 @@
...
+                        .foregroundColor(icon.color)
+                        .accessibilityLabel(accessibilityLabel)
+                        .onTapGesture(perform: action)
+                } else {
+                    BPKIconView(icon.icon)
```
gert-janvercauteren, line 68: This icon needs to be ignored, so it doesn't get focus
`accessibilityHidden(true)`

Backpack-SwiftUI/TextField/Classes/BPKTextField.swift
```diff
@@ -0,0 +1,106 @@
...
+            .background(.surfaceDefaultColor)
+            .outline(state.borderColor, cornerRadius: .sm)
+            .overlay(accessory)
+            .disabled(state.isDisabled)
+    }
+    
+    private var accessory: some View {
+        HStack {
+            Spacer()
+            if let icon = state.icon {
```
gert-janvercauteren, line 61: When 'activating' the clear button, VoiceOver gets stuck and can't move forward. We need to double check if we can resolve this or if this is just a VoiceOver bug.

Backpack-SwiftUI/TextField/Classes/BPKTextField.swift
```diff
@@ -0,0 +1,106 @@
...
+    ///   - text: The text to display in the text field.
+    public init(
+        placeholder: String = "",
+        _ text: Binding<String>
+    ) {
+        self.placeholder = placeholder
+        self._text = text
+    }
+    
+    public var body: some View {
```
gert-janvercauteren, line 47: We need to make sure the VoiceOver focus is around the full text field and not just the input part.
Can we investigate how to achieve this on SwiftUI?

Backpack-SwiftUI/TextField/Classes/BPKTextField.swift
```diff
@@ -0,0 +1,106 @@
...
+        TextField(placeholder, text: $text)
+            .font(style: .bodyDefault)
+            .foregroundColor(state.textColor)
+            .padding(.md)
+            .background(.surfaceDefaultColor)
```
gert-janvercauteren, line 52: We probably need to 'clip' the textfield as the background is now showing outside of the outline line. (Rectangle button, with rounded outline)

Example/Backpack/SwiftUI/Components/TextField/TextFieldExampleView.swift
frugoman: done

Backpack-SwiftUI/TextField/Classes/BPKTextField.swift
frugoman: done

Backpack-SwiftUI/TextField/Classes/BPKTextField.swift
frugoman: done

Backpack-SwiftUI/TextField/Classes/BPKTextField.swift
frugoman: This is now fixed!

Backpack-SwiftUI/TextField/Classes/BPKTextField.swift
frugoman: I tried several ways to achieve this, and I can't achieve it.
I think it may be the way SwiftUI handles view renders behind the scenes, as it treat it like this because there's a view on top of it (the button or icons)
I've tried changing the zIndex, using a ZStack, setting the accessibilitySortPriority, playing around with accessibilityChildren props, but I cant.
It works when I set accessibilitySortPriority to the field, but then the button cant be selected anymore!

Same applies to the order of how the button is focused BEFORE the field itself, I can't get to do it.
Maybe is worth creating a ticket to tackle this?

Backpack-SwiftUI/TextField/Classes/BPKTextField.swift
```diff
@@ -0,0 +1,106 @@
...
+    ///
+    /// - Parameters:
+    ///   - placeholder: The placeholder text to display when the text field is empty.
+    ///   - text: The text to display in the text field.
+    public init(
+        placeholder: String = "",
+        _ text: Binding<String>
+    ) {
+        self.placeholder = placeholder
+        self._text = text
```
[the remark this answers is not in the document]
frugoman: I'm not sure I understand your concern, the localisation is handled by the client. As far as our components go, localisation is a transparent thing that end up returning a string in the end, which will end up being sent to us..

That said, we could maybe check out something to add (either in backpack itself or in Main app) to be able to swiftyfy localisable strings a bit more with the use of components! I'd do it in a separate ticket though!

Backpack-SwiftUI/TextField/README.md
```diff
@@ -0,0 +1,47 @@
...
+```swift
+BPKTextField("Placeholder", text: $text)
+```
+
+### Changing the State
+    
+```swift
+BPKTextField(text: $text)
+    .inputState(.error)
+    .inputState(.valid)
```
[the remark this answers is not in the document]
frugoman: our text field does not do validation logic or anything really, is up to the client to define what state the field is in.
The design for this does not specify error labels to be added.
UIKit's BPKTextField also does not such handlings.
We could propose this to design though, as I think is how may be more widely used in Forms.
I'd argue then, that it's a feature for FormFields, more than input fields, but that could be debatable haha (BPKForm question mark?)

Backpack-SwiftUI/TextField/Classes/BPKTextField.swift
frugoman: (check the other response about issues with VO)

Backpack-SwiftUI/TextField/Classes/BPKTextField.swift
```diff
@@ -0,0 +1,110 @@
...
+    private var accessory: some View {
+        HStack {
+            Spacer()
+            if let icon = state.icon {
+                if case let .clear(accessibilityLabel, action) = state {
+                    Button(action: action) {
+                        BPKIconView(icon.icon)
+                            .foregroundColor(icon.color)
+                    }
+                        .accessibilityLabel(accessibilityLabel)
```
gert-janvercauteren, line 68: Focus order on this button is still wrong (Clear first, then field). It should focus on the field and then the clear button.

Sent you something on Slack on how we could achieve this in a different way, with correct order.

Example/Backpack/SwiftUI/Components/TextField/TextFieldExampleView.swift
```diff
@@ -0,0 +1,61 @@
...
+import Backpack_SwiftUI
+
+struct TextFieldExampleView: View {
+    @State var text = "Value"
+    @State var validationFieldText = ""
+    
+    var body: some View {
+        ScrollView {
+            VStack {
+                BPKTextField(.constant(""))
```
gert-janvercauteren, line 30: Having a constant like this will 'break' the example, as a consumer is unable to input anything and the textfield 'blinks' when trying to force input in.

can we consider another @State for this, so it doesn't 'feel broken'?

gert-janvercauteren (commented):
The clear button is not behaving as expected with VoiceOver.

Example/Backpack/SwiftUI/Components/TextField/TextFieldExampleView.swift
frugoman: yeap

Backpack-SwiftUI/TextField/Classes/BPKTextField.swift
```diff
@@ -0,0 +1,110 @@
...
+        placeholder: String = "",
+        _ text: Binding<String>
+    ) {
+        self.placeholder = placeholder
+        self._text = text
+    }
+    
+    public var body: some View {
+        HStack {
+            TextField(placeholder, text: $text)
```
gert-janvercauteren, line 49: One final thing I overlooked when reviewing.
The placeholder text color is incorrect, can we implement something like the following so we can
- Control the color
- Control the font (Haven't checked if this would be ok with our custom Skyscanner font)

https://stackoverflow.com/a/57715771

gert-janvercauteren: This is probably splitting hairs as well, but the disabled text color might not be up-to-spec as well.

gert-janvercauteren (commented):
Accessibility is now working as expected! (Nice!!).
Found some issue we may or may not want to tackle with color contrast (and placeholder colors)

gert-janvercauteren (approved):
After discussing, this component is ok.

PR #1640, opened 2023-04-13
BPKSlider setups

frugoman:
|Day|Night|
|---|---|
|iPhone-swiftui_slider___all_lm|iPhone-swiftui_slider___all_dm|


Backpack-SwiftUI/ExampleComponent/README.md
frugoman, line 1: I removed this file while at this

Backpack-SwiftUI/Slider/Classes/SliderThumbView.swift
```diff
@@ -18,16 +18,28 @@
  
 import SwiftUI
 
-public struct ExampleComponent: View {
-    @State var content: String
-
-    public var body: some View {
-        Text(content)
```
frugoman, line 25: I repurposed the example component files. Those were created just to create the SwiftUI Backpack module, back when we did that and we forgot to remove it after having actual components!

olliecurtis: Should we still have example component files, so if there is a new component contributed people have a base to work from?

This is what we have on Web to help set people up rather than guessing from a premade component and have to undo

olliecurtis (approved):

Backpack-SwiftUI/ExampleComponent/README.md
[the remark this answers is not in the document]
frugoman: yeah, it's expected! see other response

Backpack-SwiftUI/Slider/Classes/SliderThumbView.swift
frugoman: This may be a good point, I think that we may benefit more from a 'createComponent' script or similar, as we do on Android.

This Example Component is not that though, it was just created as an empty component for the library to be created when we created the SwiftUI Module. It's not to be used as an example nor a template!

PR #1711, opened 2023-07-26
NEUT-1548: Add Snippet component in SwiftUI

darioroa:

+ [x] `README.md`
+ [x] Tests


Backpack-SwiftUI/Snippet/Classes/BPKSnippet.swift
```diff
@@ -0,0 +1,104 @@
...
+                    .lineLimit(2)
+                    .accessibilityAddTraits(.isHeader)
+            }
+            if let description {
+                BPKText(description, style: .subheading)
+                    .padding(.bottom, .xl)
+            }
+            if let bodyText {
+                BPKText(bodyText, style: .bodyDefault)
+                    .lineLimit(5)
```
darioroa, line 72: Adding a limit here, as on mobile long paragraphs should be discouraged, but the value itself might be a bit low, to be discussed.

Backpack-SwiftUI/Snippet/Classes/BPKSnippet.swift
```diff
@@ -0,0 +1,104 @@
...
+            }
+        }
+        .frame(maxWidth: .infinity)
+        .onTapGesture {
+            onClick?()
+        }
```
darioroa, line 78: Using `onTapGesture`, although something like a `Button` could provide more A11y functionality out of the box (such as voiceover indication that the card is tappable as a whole). Happy to re-vise.

Backpack-SwiftUI/Snippet/Classes/BPKSnippet.swift
```diff
@@ -0,0 +1,104 @@
...
+    let onClick: (() -> Void)?
+    
+    public init(
+        image: Image,
+        accessibilityLabel: String,
+        headline: String? = nil,
+        description: String? = nil,
+        bodyText: String? = nil,
+        imageOrientation: ImageOrientation = .landscape,
+        onClick: (() -> Void)? = nil
```
AxelCB, line 37: Totally not on you/your PR, but we need to find a convention on how. we are calling these actions. Click or tap? I see mixed usages across backpack

AxelCB (commented):
Just got one question. On the screenshots for each image orientation, the portrait ones are not showing body text while the other ones are. Is that because due to the orientation that text view got pushed down and is outside the visible screen?

Backpack-SwiftUI/Snippet/Classes/BPKSnippet.swift
darioroa: That's correct, the `portrait` mode is tall enough to not show the bottom in an iPhone SE size:

https://github.com/Skyscanner/backpack-ios/assets/26660989/cf0f1f74-d151-443a-a291-8d9da8a174a0


> Totally not on you/your PR, but we need to find a convention on how. we are calling these actions. Click or tap? I see mixed usages across backpack

Agreed, just followed API table as strictly as possible

LexSwed: The table might have been agreed on with Web in mind. In browsers, both, mouse clicks and touch screen pointers are using same `click` event, hence the name.

Backpack-SwiftUI/Snippet/Classes/BPKSnippet.swift
```diff
@@ -0,0 +1,104 @@
...
+                .foregroundColor(.clear)
+                .background(
+                    image
+                        .resizable()
+                        .background(.surfaceHighlightColor)
+                        .aspectRatio(contentMode: .fill)
+                        .accessibilityLabel(accessibilityLabel)
+                )
+                .aspectRatio(aspectRatio(for: imageOrientation), contentMode: .fit)
+                .clipShape(RoundedRectangle(cornerRadius: .md, style: .continuous))
```
frugoman, line 60: Should this all be just `image` with a clipshape? here we have a RoundedRectangle that then also sets a clipshape of RoundingRectangle

Backpack-SwiftUI/Snippet/Classes/BPKSnippet.swift
AxelCB: I agree that given that the whole snippet is tappable (and not just a section of it) the tap gesture handler could be set from outside directly. Would we need to coordinate this across platforms? 🤔

darioroa: 🤔 Interesting. I would argue for keeping it, as it would make logic for tapping (which is more related to views, and thus to the domain of Backpack) more unified. Otherwise it could encourage a pattern where this component is used with `onTapGesture` in some places, or nested inside a `Button` in others, which knowing how SwiftUI behaves and changes across iOS versions, platforms and devices could save us some unexpected results.

Backpack-SwiftUI/Snippet/Classes/BPKSnippet.swift
darioroa: Good catch! 
Using the `image` directly (instead of being used as the "background" of this rectangle) is tricky because the image itself has a `fill` requirements, while the actual shape we want is to preserve the defined aspect ratio, needing another layer. For instance, we could use a GeometryReader like so: 

```swift
GeometryReader { geo in
    image
        .resizable()
        .scaledToFill()
        .frame(width: geo.size.width)
}
.aspectRatio(aspectRatio(for: imageOrientation), contentMode: .fill)
.clipShape(RoundedRectangle(cornerRadius: .md, style: .continuous))
```

Backpack-SwiftUI/Snippet/Classes/BPKSnippet.swift
darioroa: Update: it seems like the `onClick` function might have been more aimed for the desktop version, let's double check with design 👍

darioroa: Double checked, indeed it is specified to include the `onClick` action, even on apps

frugoman (approved):
Looks good!

Backpack-SwiftUI/Snippet/Classes/BPKSnippet.swift
frugoman: Lets keep it as is for now!

PR #1713, opened 2023-07-28
NEUT-1549: Card List - SwiftUI

AxelCB:

+ [x] `README.md`
+ [x] Tests


Backpack-SwiftUI/CardList/Classes/BPKCardList.swift
```diff
@@ -0,0 +1,131 @@
...
+ * limitations under the License.
+ */
+
+import SwiftUI
+
+public struct BPKCardList<Element: Identifiable, Content: View>: View {
+    private let title: String
+    private let description: String?
+    private let layout: BPKCardListLayout
+    private let initiallyShownCards: Int
```
gert-janvercauteren, line 25: nit: Do we want to include 'count' in the naming. So the naming is more precise?

Backpack-SwiftUI/CardList/Classes/BPKCardList.swift
```diff
@@ -0,0 +1,131 @@
...
+                }
+                .padding(.horizontal)
+            }
+        }
+    }
+
+    private func stackLayout(with accessory: BPKCardListLayout.Accessory?) -> some View {
+        cardListSkeleton(with: accessory?.sectionHeaderButton) {
+            let cardsShown = (showingAllCards) ? elements.count : min(initiallyShownCards, elements.count)
+            let filteredCards = elements[0..<cardsShown]
```
gert-janvercauteren, line 71: nit: Should this bit of logic be extracted?

Backpack-SwiftUI/CardList/Classes/BPKCardList.swift
```diff
@@ -0,0 +1,131 @@
...
+            ForEach(filteredCards) { element in
+                cardForElement(element)
+            }
+                .padding(.horizontal)
+            accessoryView(accessory: accessory)
+                .padding(.horizontal)
```
gert-janvercauteren, line 77: Can we make sure we use BPKSpacing in the `.padding` modifier? Not sure if the default is what we want (It's probably 16, but not tied to Backpack)

Backpack-SwiftUI/CardList/Classes/BPKCardList.swift
```diff
@@ -0,0 +1,131 @@
...
+        }
+    }
+
+    private func sectionHeader(with sectionHeaderButton: BPKCardListLayout.SectionHeaderAction?) -> some View {
```
gert-janvercauteren, line 81: nit: should we pass in the title and description as well? 
Considering we're now both using vars from the component and what is unwrapped from input params

Backpack-SwiftUI/CardList/Classes/BPKCardList.swift
```diff
@@ -0,0 +1,131 @@
...
+        if let sectionHeaderButton {
+            return BPKSectionHeader(title: title, description: description) {
+                BPKButton(
+                    icon: sectionHeaderButton.icon,
+                    accessibilityLabel: sectionHeaderButton.accessibilityLabel,
+                    action: sectionHeaderButton.action)
+                .buttonStyle((sectionHeaderButton.style == .default) ? .primary : .primaryOnDark)
```
gert-janvercauteren, line 88: Did we create a special section header button? Worried that now the button styling is here and in the SectionHeader component. 

If we ever change which styles are used, we risk having to change in two places

Backpack-SwiftUI/CardList/Classes/BPKCardList.swift
```diff
@@ -0,0 +1,131 @@
...
+            }
+            .padding()
+        } else {
+            return BPKSectionHeader(title: title, description: description)
+                .padding()
```
gert-janvercauteren, line 93: These `.padding` probably need to add a BPKSpacing token to make it explicit. The system default is 16, but not linked to our design system

Backpack-SwiftUI/CardList/Classes/BPKCardList.swift
```diff
@@ -0,0 +1,131 @@
...
+        }
+    }
+
+    private func cardListSkeleton<LayoutContent: View>(
+        with sectionHeaderButton: BPKCardListLayout.SectionHeaderAction?,
```
gert-janvercauteren, line 98: nit: Should the naming more closely align? As the param states 'button', but the type is 'action'

Backpack-SwiftUI/CardList/Classes/BPKCardListAccessory.swift
```diff
@@ -0,0 +1,38 @@
...
+
+import SwiftUI
+
+extension BPKCardListLayout {
+    public enum Accessory {
+        case sectionHeaderButton(SectionHeaderAction)
+        case expand(expandText: String, collapseText: String, sectionHeaderButton: SectionHeaderAction? = nil)
+        case footerButton(AccessoryAction)
+        
+        var sectionHeaderButton: SectionHeaderAction? {
```
gert-janvercauteren, line 27: nit: can we rename it to sectionHeaderAction ?

Backpack-SwiftUI/Tests/CardList/__Snapshots__/BPKCardListViewTests/testCardListRail.light-mode.png
gert-janvercauteren, line 1: It seems that the shadow if the cards is clipped at the bottom, is this the case in 'real' usage also?

Backpack-SwiftUI/Tests/CardList/__Snapshots__/BPKCardListViewTests/testCardListStackWithExpandAccessoryAndSectionHeaderButton.dark-mode.png
gert-janvercauteren, line 1: Can we test what happens if the title & subtitle are larger than the horizontal space?
We should also consider implications when font sizes scale (but that's not something to fix right now)

gert-janvercauteren (commented):
Thanks for contributing!
Left some comments, which I would like you to address and some `nit: `, which you can ignore if you don't agree :D 

Looking great overall! 
One small question: Where did we pull the images from and what is the licensing on them?

Backpack-SwiftUI/CardList/Classes/BPKCardList.swift
AxelCB: I thought that as well but was worried it could get too long
"initially" is a bit long though haven't found any good synonym

Backpack-SwiftUI/CardList/Classes/BPKCardList.swift
AxelCB: That's a fair point. I'll check with design, I understood the style restriction here only applied to the button, not the whole section header, but might have misunderstood

Backpack-SwiftUI/Tests/CardList/__Snapshots__/BPKCardListViewTests/testCardListStackWithExpandAccessoryAndSectionHeaderButton.dark-mode.png
AxelCB: I'll add a test for that 😄 
Wouldn't mind checking out/testing how font sizes scale, but shouldn't we include it on the assert snapshots util as we currently have for light / dark mode?

AxelCB:
> Thanks for contributing! I left some comments, which I would like you to address and some `nit: `, which you can ignore if you don't agree :D
> 
> There are some usages of `.padding()` without any BPKSpacing tokens, could you make sure they are using one, I might not have caught all of the uses in my review 🙇
> 
> One small question: Where did we pull the images from, and what is the licensing on them?
> 
> Looking great overall! 🎉

Thanks a lot for such a quick review!
I've been going through the comments, hopefully can send an update today.
About the images, I actually took them from the Example app assets, they just weren't included in the test bundle. But given that, I don't really know about licensing since they were already there. Should we remove them and replace them with other ones?

gert-janvercauteren:
> 

As long as we can sync the docs screenshots with what is in Figma, same images & content it's all ok for me :D

Backpack-SwiftUI/Tests/CardList/__Snapshots__/BPKCardListViewTests/testCardListRail.light-mode.png
AxelCB: That's a tricky one actually. Yes, it happens on 'real' / example usage also, and it's caused due to how `BPKCard` has defined it's shadow and size. The shadow is offset from the card's intrinsic content size, so it overflows a bit. The card itself doesn't have any padding to account for it (the shadow is applied to the wrapping button but the padding option is only applied to the content within that button), therefore if CardList doesn't include any vertical padding for it's cards content, it eats up that shadow space/offset.

Personally, I think `BPKCardList` shouldn't be responsible for that, `BPKCard` should be aware that it needs some extra space for it's shadow, and add the necessary padding to show it properly (shouldn't be more than 4 points as far as I checked).
What do you think?

Backpack-SwiftUI/CardList/Classes/BPKCardList.swift
gert-janvercauteren: `defaultVisibleCardCount` could be a contender, but naming things is **hard** 😅

Backpack-SwiftUI/CardList/Classes/BPKCardList.swift
```diff
@@ -0,0 +1,134 @@
...
+                (showingAllCards) ? collapsingText : expandingText
+            ) {
+                withAnimation {
+                    showingAllCards.toggle()
+                }
+            }
+            .buttonStyle(.link)
+            .stretchable()
+        }
+    }
```
gert-janvercauteren, line 133: Should we make an extension on the Accessory enum to provide a `var accessory`, this way we can switch over all cases and have all layouts neatly organised. 

as for calling `showingAllCards.toggle()` we can resolve that with a button.action closure

gert-janvercauteren (commented):
Thanks for the adjustments! 
Just ran it on the simulator and it's looking great. 

Left some comments around performance and the scrollbar showing on top of the content. 
Let me know if there's anything I can help out with!

Backpack-SwiftUI/CardList/Classes/BPKCardList.swift
```diff
@@ -0,0 +1,134 @@
...
+    private func stackLayout(with accessory: BPKCardListLayout.Accessory?) -> some View {
+        cardListSkeleton(with: accessory?.sectionHeaderAction) {
+            ForEach(filteredCards) { element in
+                cardForElement(element)
+            }
+            .padding(.horizontal, .base)
+            accessoryView(accessory: accessory)
+                .padding(.horizontal, .base)
+        }
+    }
```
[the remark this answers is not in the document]
AxelCB: Let me take a look at it, there's currently no API limitation but design wise, there should never be more than 12 elements in the list

Backpack-SwiftUI/CardList/Classes/BPKCardList.swift
AxelCB: > Should we make an extension on the Accessory enum to provide a var accessory, this way we can switch over all cases and have all layouts neatly organised.

Not sure I understood you correctly, `Accessory` enum is already the accessory. We could add a computed variable to `Layout` to provide an accessory if it applies, but not sure how it would improve.
Using `if case` instead of `switch` here we decided with @frugoman to include section header button as an extra accessory option, and using `switch` forced us to provide `EmptyView` as it's content.

> as for calling showingAllCards.toggle() we can resolve that with a button.action closure
Can add a closure to configure it, but what I understood is that the expand accessory option provides a built in functionality that's not as configurable

Backpack-SwiftUI/CardList/Classes/BPKCardListLayout.swift
```diff
@@ -0,0 +1,24 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import SwiftUI
+
+public enum BPKCardListLayout {
+    case rail(sectionHeaderAction: SectionHeaderAction? = nil)
```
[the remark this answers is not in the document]
AxelCB: Checked with design, I'll remove it

Backpack-SwiftUI/Tests/CardList/__Snapshots__/BPKCardListViewTests/testCardListRail.light-mode.png
AxelCB: Super low priority, but what do you folks think of the above? @frugoman @gert-janvercauteren 
Just want to clear out which component's responsibility would that be 😬

gert-janvercauteren: If we add it to the card, this might cause issues for the non elevated card. 
I would rather have if on the rail for now, to 'play it safe'. But keen to hear what Nico thinks.

AxelCB: Thanks for the early reply! 😄 
I don't mind adding it but I think even when choosing elevated card, one might not always remember to add padding for this, and being that `BPKCard` can know if it's elevated (hence, having shadow), it can selectively add the padding when needed. But of course, that's extra complexity we might not want 😅 
So let's wait for Nico's opinion as well 👀

Backpack-SwiftUI/CardList/Classes/BPKCardList.swift
AxelCB: After discussing with design, we decided to remove this style option for now
Ideally we may want to allow any styling option for the accessory button, but went with only `.primary` for a first version

Backpack-SwiftUI/CardList/Classes/BPKCardList.swift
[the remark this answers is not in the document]
AxelCB: @gert-janvercauteren I checked those docs and they recommend only switching to `LazyVStack` / `LazyHStack` if there are performance issues. Apple doesn't include many details, but it says:

- Lazy stacks trade some degree of layout correctness for performance

Do we still want to go with the lazy ones?

Backpack-SwiftUI/CardList/Classes/BPKCardList.swift
AxelCB: I never applied any change for this. I wouldn't mind changing it, but I also used that name because it's the one on the API table which I guessed other platforms would use.
Any strong opinions here? We could always rename it letter, but both options are fine with me 😄

Backpack-SwiftUI/CardList/Classes/BPKCardListSectionHeaderAction.swift
```diff
@@ -0,0 +1,37 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import SwiftUI
+
+public extension BPKCardListLayout {
+    struct SectionHeaderAction {
+        let icon: BPKIcon
```
gert-janvercauteren, line 23: Should this use the BPKButton.Icon like BPKCardListAccessoryAction

Backpack-SwiftUI/CardList/Classes/BPKCardList.swift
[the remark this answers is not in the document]
gert-janvercauteren: Reading further into the docs, we can keep the current approach and switch over once we do hit these performance issues. 

Thanks for digging deeper as well!

gert-janvercauteren (commented):
Some small-ish comments on properties. 

Those docs screenshots look absolutely amazing!

Backpack-SwiftUI/CardList/Classes/BPKCardListSectionHeaderAction.swift
AxelCB: Personally, even though it looks weird that they're different, I think it makes more sense like this.
SectionHeader button by design shouldn't allow text, only icons (that came up as a requirement after it was contributed so not implemented yet), and therefore using `BPKButton.Icon` is rather pointless since both leading and trailing positions would be the same. We would be ignoring it since we should only use the `BPKButton` init without title

The accessory button is more flexible though, and could have both title, and icon in any position

Backpack-SwiftUI/Tests/CardList/__Snapshots__/BPKCardListViewTests/testCardListRail.light-mode.png
frugoman: I'd say lets add it to the rail. I think we explored adding it to the rail before?
its the parent that needs to set the spacing for it's children so it would be fine.

darioroa:
Hi there, checking in to see if we need any more modifications here, as I see tests are passing already (thanks @gert-janvercauteren 🎉!)

PR #1715, opened 2023-07-28
[LMN-25] Multi select chips group

frugoman:
Introducing Multi Select Chips Group, with support for adding a sticky chip (icon only) for the `rail` type of group

https://www.figma.com/file/irZ3YBx8vOm16ICkAr7mB3/Backpack?node-id=30119%3A41732&mode=dev

image
image


frugoman:
Continued in: https://github.com/Skyscanner/backpack-ios/pull/1717

PR #1739, opened 2023-08-15
NEUT-1549: Remove background colour from section header

AxelCB:
Background colour shouldn't be set within section header component but be set by the container, so removing it from inside

+ [x] `README.md`
+ [x] Tests


gert-janvercauteren (approved):
Thanks for making this change! 🚀

PR #1779, opened 2023-09-15
Adding a swiftui version of the dialog used as a zstack of views

frugoman:
Adding a SwiftUI version of the dialog component, which can be used as a ZStack of views instead of a view controller. The SwiftUI dialog uses the same content and actions views as the UIKit dialog, but wraps them in a custom modifier that handles the presentation logic.

This is meant to be used in cases where presenting a view controller is not an option. The example use case for Hotels is that we need to show a dialog on a screen that's currently presenting a screen, making a new presentation impossible.

This, being just a view, has a couple of drawbacks, like presentation events for accessibility, which needs to now be done manually from the client or the scrim not being able to be drown on top of a tab bar controller or a nav bar.
So, the usage of this component is not encouraged, but there are some cases where this a different option is not possible, like the above mentioned case.

### DIFF
--- a/Backpack-SwiftUI/Dialog/Classes/BPKDialogView.swift
+++ b/Backpack-SwiftUI/Dialog/Classes/BPKDialogView.swift
@@ -18,9 +18,13 @@
 
 import SwiftUI
 
+/// This extension provides a set of convenience methods to display a dialog that works as a view.
+/// It's usage is discouraged in favour of the `BPKDialog` view. Usage of this view should be limited to cases where
+/// presenting a new view controller is not possible. (`BPKDialog` uses a view controller to present the dialog
+/// behind the scenes).
 public extension View {
     /// Displays a success dialog with a title, text and a list of buttons.
-    func bpkSuccessDialog(
+    func bpkSuccessDialogView(
         presented: Binding<Bool>,
         icon: BPKIcon,
         title: String,
@@ -36,7 +40,7 @@ public extension View {
                 buttons.append(BPKDialogButton(button: linkButton, style: .link))
             }
         }
-        return self.modifier(UIKitDialogContainerViewModifier(isPresented: presented, dialogContent: {
+        return self.modifier(DialogContainerViewModifier(isPresented: presented, dialogContent: {
             DialogWithIconContent(
                 icon: icon,
                 iconColor: .coreAccentColor,
@@ -47,7 +51,7 @@ public extension View {
     }
     
     /// Displays a warning dialog with a title, text and a list of buttons.
-    func bpkWarningDialog(
+    func bpkWarningDialogView(
         presented: Binding<Bool>,
         icon: BPKIcon,
         title: String,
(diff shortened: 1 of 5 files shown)


Backpack-SwiftUI/Dialog/Classes/DialogContainerViewModifier.swift
```diff
@@ -30,57 +30,40 @@ struct DialogContainerViewModifier<DialogContent: View>: ViewModifier {
...
-                                Spacer(minLength: .lg)
-                                dialogContent
-                                    .frame(maxWidth: 400)
-                                Spacer(minLength: .lg)
-                            }
-                        }
+        ZStack {
+            content
+            if isPresented {
+                ZStack {
```
MBaldo83, line 37: If we already have our content in a ZSgtack, why do we need a second one here?

Backpack-SwiftUI/Dialog/Classes/BPKDialogView.swift
```diff
@@ -0,0 +1,171 @@
...
+
+import SwiftUI
+
+/// This extension provides a set of convenience methods to display a dialog that works as a view.
+/// It's usage is discouraged in favour of the `BPKDialog` view. Usage of this view should be limited to cases where
+/// presenting a new view controller is not possible. (`BPKDialog` uses a view controller to present the dialog
+/// behind the scenes).
+public extension View {
+    /// Displays a success dialog with a title, text and a list of buttons.
+    func bpkSuccessDialogView(
```
MBaldo83, line 27: @frugoman As I understand it, making the accessibility announcement that the dialog has been presented is now the responsibility of the consumer since we are not using `UIViewController.present`. Would it be possible to require a localised string in each of these interfaces that is the voice-over announcement to use when presented, and then `DialogContainerViewModifier` would have the responsibility to make the announcement. This ensures anyone using the interface has correct voice-over call outs. WDYT?

Backpack-SwiftUI/Dialog/Classes/DialogContainerViewModifier.swift
frugoman: This is because we need to group the content and apply certain modifiers to the group. It so happens that the content of the dialog is also a zstack but that's a ''''''detail'''''' of the dialog content.

Backpack-SwiftUI/Dialog/Classes/BPKDialogView.swift
frugoman: @gert-janvercauteren what do you reckon would be the best approach? the dialog component does not know what the content of the dialog is, so having the event in the component seems weird.
(what the screen reader needs to say when a dialog is presented is not a specific string, but rather focus on one given part of the screen, and read that.)

I'm not sure forcing an interface that requires a string or similar is the way to go here

gert-janvercauteren (approved):
LGTM!

PR #1791, opened 2023-10-03
Added Label Flares to BPKRangeSlider

MBaldo83:
Added feature to present a flared label above the thumb circle of a range slider

### Snapshots
testRangeSliderWithThumbnailLabels dark-mode
testRangeSliderWithThumbnailLabels light-mode
testRangeSliderWithThumbnailLabels rtl

+ [x] `README.md`
+ [x] Tests


Backpack-SwiftUI/Slider/Classes/BPKRangeSlider.swift
```diff
@@ -43,20 +59,22 @@ public struct BPKRangeSlider: View {
...
+        self.thumbnailLabels = thumbnailLabels
     }
     
     public var body: some View {
         GeometryReader { geomentry in
             sliderView(sliderSize: geomentry.size)
         }
         .fixedSize(horizontal: false, vertical: true)
-        .frame(height: thumbSize)
+        .frame(height: height)
```
MBaldo83, line 77: @frugoman honestly I got a bit stuck here, and I think there's probably a better solution. 

The issue I faced is that previously the height was hardcoded to the `thumbSize`. Actually if you added a border to this view, it was not really in the right place even before my changes I think due to the views being centred vertically in the ZStack. But anyway, the complication with the the changes here is that the height of the View is dependant on whether we add the thumbnail labels or not. _Also_, the height of the thumbnail labels is not constant due to dynamic type. So what I have done is set the height from within the sliderView ZStack. Se line 142. The height is calculated there based on intrinsic size of the ZStack + the offset if required.

I think there must be a better way to do this to be honest, but GeometryReader seems to be necessary here, but also requires a height to be set in order to be used within other views in a simple way, so I'd love to hear your thoughts / ideas on this!

frugoman: hmmm lets pair on it tomorrow, I dont quite understand why this is even needed tbh

frugoman (approved):
Changes look good!! Thanks for contributing and refactoring the existing bit to make it better!

PR #1802, opened 2023-10-18
Emitting event for slider changes

frugoman:
Emitting an event when the user stops dragging the slider or range slider. This is useful for applications that need to react to the slider value changes

The changes are:

* Added a parameter onDragEnded to BPKSlider and BPKRangeSlider that takes a closure that will be called with the new value or range when the user stops dragging.
* Added a call to onDragEnded in the SliderThumbView when the drag gesture ends.
* Updated the README.md file with examples of how to use the onDragEnded parameter.

### DIFF
--- a/Backpack-SwiftUI/Slider/Classes/SliderThumbView.swift
+++ b/Backpack-SwiftUI/Slider/Classes/SliderThumbView.swift
@@ -22,6 +22,7 @@ struct SliderThumbView: View {
     let size: CGFloat
     let offset: CGFloat
     let onDrag: (DragGesture.Value) -> Void
+    let onDragEnded: () -> Void
     
     var body: some View {
         Circle()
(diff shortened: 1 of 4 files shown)


Backpack-SwiftUI/Slider/Classes/SliderThumbView.swift
```diff
@@ -34,12 +35,15 @@ struct SliderThumbView: View {
                     .onChanged { value in
                         onDrag(value)
                     }
+                    .onEnded { _ in
```
marianeum, line 38: is this also called when using the accessibility controls?

frugoman: good catch! no, it would not, i'll call the event when the accessibiltiy methods are executed!

marianeum (approved):

PR #1809, opened 2023-11-07
OGM-1926: Disable month selection for last month

simon-o:
OGM-1926

We want to disable the last month button as it select the actual month instead. 
To fix this issue we decided to disable it for now.

Simulator Screen Shot - iPhone 8 - 2023-11-07 at 10 00 59

+ [ ] `README.md`
+ [ ] Tests

### DIFF
--- a/Backpack/Calendar/Classes/BPKCalendar.m
+++ b/Backpack/Calendar/Classes/BPKCalendar.m
@@ -455,7 +455,7 @@ CGFloat const BPKCalendarDefaultCellHeight = 44;
     NSDate *minDate = [self.minDate dateForCalendar:self.gregorian];
 
     NSDateComponents *comps = [self.gregorian components:NSCalendarUnitMonth fromDate:minDate toDate:monthDate options:0];
-    return comps.month >= 0;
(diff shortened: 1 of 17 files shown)


melwaraki (approved):

Backpack/Calendar/Classes/BPKCalendar.m
```diff
@@ -453,9 +453,11 @@ - (void)selectWholeMonth:(BPKSimpleDate *)month {
 - (BOOL)isWholeMonthButtonEnabledForMonth:(BPKSimpleDate *)month {
```
gert-janvercauteren, line 453: Could you please add a test for this logic? 
Not 100% sure if there was a test before, but it feels like we have hit some issues with 'wholeMonth' before.

gert-janvercauteren (dismissed):
I don't fully understand the use case. 
Could we just add logic to check if the maxDate is bigger (or equal) than the last day in the month we show the button for? 

This will allow us to have whole month only on months that have a full month of days available to select.

gert-janvercauteren (approved):

PR #1818, opened 2023-11-14
[LMN-20] SwiftUI Calendar component

frugoman:
Initial implementation for the Calendar component purely on SwiftUI

Following: [Figma's Calendar](https://www.figma.com/file/irZ3YBx8vOm16ICkAr7mB3/Backpack?node-id=30428%3A56087&mode=dev)

currently supporting: 
* Single selection
* Range selection
* Adding accessory action for each month

image
image
image

Remaining features to match UIKit's implementation will be added on a needed-basis, like:
* Multiple selection
* Select whole month accessory action (working within the component)
* Support for adding additional info on each day in calendar (like prices)


Backpack-SwiftUI/Calendar/Classes/BPKCalendar.swift
```diff
@@ -0,0 +1,129 @@
...
+        calendar: Calendar,
+        validRange: ClosedRange<Date>
+    ) {
+        _currentlyShownMonth = State(initialValue: validRange.lowerBound)
+        self.validRange = validRange
+        self.calendar = calendar
+        self.selectionType = selectionType
+
+        monthHeaderDateFormatter = DateFormatter()
+        monthHeaderDateFormatter.dateFormat = "MMMM yyyy"
```
MBaldo83, line 55: Does this date format work for RTL and LTR languages?

Backpack-SwiftUI/Calendar/Classes/AccessibilityConfigurations/RangeAccessibilityConfigurations.swift
```diff
@@ -0,0 +1,61 @@
...
+///     as the first date in the range.
+///   - endSelectionState: The label provided to assistive technologies informing a user that a date is selected
+///     as the second date in the range.
+///   - betweenSelectionState: The label provided to assistive technologies informing a user that a date lies
+///     between the first and second selected dates.
+///   - startAndEndSelectionState: The label provided to assistive technologies informing a user that a date
+///     is selected as both the first and second date in the range.
+///   - returnDatePrompt: The prompt provided to assistive technologies informing a user that they should now
+///     select a second date.
+public struct RangeAccessibilityConfigurations {
```
MBaldo83, line 35: Should this struct live within the calendar component or be a general Backpack utility? Just asking about which folder this should live in for discoverability purposes. If it's in here, there's less chance of it being re-used elsewhere

Backpack-SwiftUI/Tests/Calendar/BPKCalendarTests.swift
```diff
@@ -0,0 +1,122 @@
...
+            )
+            .frame(width: 320, height: 720)
+        )
+    }
+    
+    func test_rangeCalendarDayCells() {
+        let calendar = Calendar.current
+        let date = calendar.date(from: .init(year: 2023, month: 11, day: 8))!
+        
+        assertSnapshot(
```
MBaldo83, line 110: 👍 very useful to have all the states tested, and visualised through the snapshot 😄

Backpack-SwiftUI/Calendar/Classes/CalendarBadge.swift
```diff
@@ -0,0 +1,33 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import SwiftUI
+
+struct CalendarBadge: View {
```
MBaldo83, line 21: Is it worth having snapshot tests for this as well?

MBaldo83: Unless I missed these...

frugoman: This already appears in the calendar snapshot, I don't think its worth it to have it's own, given also how simple it is!

Backpack-SwiftUI/Calendar/Classes/AccessibilityConfigurations/RangeAccessibilityConfigurations.swift
frugoman: It's something that's specific for the Calendar, so living within the Calendar folder is the right place I think. It should not be reused, it's super specific to this!

Backpack-SwiftUI/Calendar/Classes/CalendarMonthHeader.swift
```diff
@@ -0,0 +1,88 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import SwiftUI
+
+/// CalendarMonthHeader is a view that displays the month name and an optional accessory action.
+/// Also exposes a binding to the currently shown month.
+struct CalendarMonthHeader: View {
```
MBaldo83, line 23: Should this have a snapshot test as well, it looks like it contains some layout logic for different states

Backpack-SwiftUI/Calendar/Classes/Core/CalendarContainer.swift
```diff
@@ -0,0 +1,75 @@
...
+
+import SwiftUI
+
+struct CalendarContainer<MonthContent: View>: View {
+    let calendar: Calendar
+    let validRange: ClosedRange<Date>
+    @ViewBuilder let monthContent: (_ month: Date) -> MonthContent
+    
+    var body: some View {
+        ScrollView {
```
MBaldo83, line 27: From what I can tell, you're not doing any cell recycling of the months, I guess that's because there's a limit of 12 months, so it's not necessary, is that correct?

Backpack-SwiftUI/Calendar/Classes/Core/CalendarMonthGrid.swift
```diff
@@ -0,0 +1,106 @@
...
+    let validRange: ClosedRange<Date>
+
+    @ViewBuilder let dayCell: (Date) -> DayCell
+    @ViewBuilder let emptyLeadingDayCell: () -> EmptyLeadingDayCell
+    @ViewBuilder let emptyTrailingDayCell: () -> EmptyTrailingDayCell
+
+    private let daysInAWeek = 7
+    
+    var body: some View {
+        // Sunday is the first day of the week, so we need to offset the days to make Monday the first day
```
MBaldo83, line 37: I think you don't need the 1st line of this comment

Backpack-SwiftUI/Calendar/Classes/Range/Cells/RangeSelectionCalendarDayCell.swift
```diff
@@ -0,0 +1,57 @@
...
+    var body: some View {
+        if let selection, selection.contains(date) {
+            if selection.lowerBound == selection.upperBound {
+                LowerAndUpperBoundSelectedCell(calendar: calendar, date: date)
+            } else if date == selection.lowerBound {
+                LowerBoundSelectedCell(calendar: calendar, date: date)
+            } else if date == selection.upperBound {
+                UpperBoundSelectedCell(calendar: calendar, date: date)
+            } else {
+                InbetweenSelectionCell(calendar: calendar, date: date)
```
MBaldo83, line 35: I'm trying to consider the pros / cons of making individual cells like this versus have a view modifier to apply these various states to a common calendar cell.
I guess the way you have it here with distinct View structs gives some simplicity, and clarity when reading this, but it implies more duplication, eg we have to create a `BPKText()` view in each cell, and apply mostly the same styling to it each time. 
Just wondering if you considered making view modifiers to represent each of these cell types, which could be applied to a common view, or if you think that would end up being too painful / not working?

Backpack-SwiftUI/Calendar/Classes/BPKCalendar.swift
frugoman: It's the same format we used before! So yeah it's the way we expect it to be!

Backpack-SwiftUI/Calendar/Classes/CalendarMonthHeader.swift
frugoman: It's also covered by the other snapshots! I dont think it needs its own!

Backpack-SwiftUI/Calendar/Classes/Core/CalendarContainer.swift
frugoman: the months themselves are not recycled, but the day cells are. if both the day cells and months are dynamically resolved while scrolling, that produces jumps in the UI while scrolling, because a new month will load but then more cells need to render (the days), pushing content up more.

Later on, when we can support using the Grid component with iOS 16 (only LazyVGrid is available for ios 14 :/ ) we'll be able to switch to recycling months and not the days, which would be preferred!

So we're kinda cornered here by the ios versions we support!

Backpack-SwiftUI/Calendar/Classes/Core/CalendarMonthGrid.swift
frugoman: ahh good catch!

Backpack-SwiftUI/Calendar/Classes/Range/Cells/RangeSelectionCalendarDayCell.swift
frugoman: I considered it yes. I ended up with different views for each day cell type to make them not depend on each other too much and have more flexibility when adding more cell types (for example, the component needs to support (not in this PR) adding prices to the day cells)

So flexibility here is more worth having than some duplication.

Duplication is not really happening much though here (except for maybe the fact it's using a BPKText with with lineLimit=1) and a modifier would be slightly harder to use for this usecase, in general, conditionally adding modifiers should not happen, as re-composing should be the approach taken... aaaaanyways haha the different structs approach is prefered for this case

gert-janvercauteren (commented):
I noticed that the day headings use a different format compared to the existing calendar. Should we align Mon / Tue / Wed between both UIKit and SwiftUI?

There's also a bit of funkyness when a selection happens, which leads to a double read out of a cell, it feels like it redraws a large portion of the view when the state changes.

frugoman:
> I noticed that the day headings use a different format compared to the existing calendar. Should we align Mon / Tue / Wed between both UIKit and SwiftUI?
> 
> There's also a bit of funkyness when a selection happens, which leads to a double read out of a cell, it feels like it redraws a large portion of the view when the state changes.

About showing the months Mon/Tue/Wed or M/T/W, I followed Figma for this, which shows days using only one letter!

on the selection funkyness with VoiceOver, I'm not really sure what's going on? I've debugged every component taking part here and nothing I could find gets re-rendered.

PR #1825, opened 2023-11-29
Making Flow stack view public

frugoman:
Making the `FlowStackView` public so that it can be used from outside Backpack.

### DIFF
--- a/Backpack-SwiftUI/Utils/Classes/FlowStackView.swift
+++ b/Backpack-SwiftUI/Utils/Classes/FlowStackView.swift
@@ -18,20 +18,25 @@
 
 import SwiftUI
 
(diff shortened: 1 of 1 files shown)


gert-janvercauteren (commented):
Please rename the FlowStackView to include the BPK prefix. 
All public components should bear this prefix.

Also please add documentation (a README.md) as it is now for public use :D

AxelCB:
Besides the documentation/README that was raised earlier, do we need tests and screenshots for this component now that it's public?
I imagine it's being tested through Chip Group, but we might want individual tests and screenshots for this component

gert-janvercauteren (approved):

PR #1834, opened 2023-12-08
Conform BPKAppSearchModalResults to Equatable

iHandle:
Conform BPKAppSearchModalResults to Equatable
Even if we can do it outside the module by making all the properties public, but it's verbose.

+ [ ] `README.md`
+ [ ] Tests


Backpack-SwiftUI/AppSearchModal/Classes/BPKAppSearchModalResults.swift
```diff
@@ -112,6 +112,18 @@ public struct BPKAppSearchModalAction {
     }
 }
 
+extension BPKAppSearchModalContent.Shortcut: Equatable {
```
frugoman, line 115: Could you add conformance to Equatable in the struct definition instead? so you dont have to implement this manually (which we can forget to update if we add a new property! )

frugoman (approved):

PR #1836, opened 2023-12-08
SONIC-2126: BPKCarousel page indicator should update when delegate = nil

dennis-li-skyscanner:
# Changes
- Fix the issue which swiping BPKCarousel doesn't update the pageIndicator when its delegate is nil

+ [ ] `README.md`
+ [ ] Tests

### DIFF
--- a/Backpack/Carousel/Classes/BPKCarousel.swift
+++ b/Backpack/Carousel/Classes/BPKCarousel.swift
@@ -26,11 +26,7 @@ public protocol BPKCarouselDelegate: AnyObject {
 public final class BPKCarousel: UIView {
     private let internalCarousel: BPKInternalCarousel
     
-    public weak var delegate: BPKCarouselDelegate? {
-        didSet {
-            internalCarousel.delegate = self
-        }
-    }
+    public weak var delegate: BPKCarouselDelegate?
     
     public var currentImage: Int { internalCarousel.currentImage }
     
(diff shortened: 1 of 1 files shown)


frugoman (approved):
Nice one

Backpack/Carousel/Classes/BPKCarousel.swift
```diff
@@ -26,11 +26,7 @@ public protocol BPKCarouselDelegate: AnyObject {
 public final class BPKCarousel: UIView {
     private let internalCarousel: BPKInternalCarousel
     
-    public weak var delegate: BPKCarouselDelegate? {
-        didSet {
-            internalCarousel.delegate = self
-        }
-    }
```
frugoman, line 33: I think you'd still want to do this if someone sets a delegate for this.

frugoman (changes requested):
Nice one

Backpack/Carousel/Classes/BPKCarousel.swift
dennis-li-skyscanner: My understanding is the `internalCarousel.delegate` would never change after the Carousel is initialized.

When the external delegate (BPKCarouselDelegate) is updated, the function below calls it 

```
 public func onImageChange(index: Int) {
        pageIndicator.currentIndex = index
        delegate?.onImageChange(index: index)
    }
```

dennis-li-skyscanner: Yeah. I reverted my change The didSet block is unchanged now

PR #1842, opened 2023-12-20
fix: LMN-925 BPKAppSearchModal recreating TextField when changing from loading to content

MBaldo83:
In the view BPKAppSearchModal we draw a BPKTextField when the state is loading or content.
Previously this was done through a switch statement, so SwiftUI was unable to identify that we wanted this to be the _same_  BPKTextField.

Now for both loading & content states, we render the same BPKTextField using an 'if' statement based on content state instead of within the 'switch' statement, this informs Swift UI that it doesn't need to redraw the BPKTextField when changing between loading and content states.

### After Fix Issue Resolved
App Search Modal - NOT Losing Focus

### Before Fix Showing Issue
App Search Modal - Losing Focus

+ [ ] `README.md`
+ [ ] Tests


frugoman (approved):
Looks Good!

PR #1846, opened 2023-12-22
Update BPKText to use LocalizedStringKey for text

frugoman:
This pull request updates the BPKText struct to use LocalizedStringKey for the text parameter instead of a plain string. It also adds a test for styled text in BPKText.

image


Backpack-SwiftUI/Tests/Text/BPKTextTests.swift
```diff
@@ -40,10 +40,14 @@ class BPKTextTests: XCTestCase {
...
     
+    func test_withStyledText() {
+        assertSnapshot(BPKText("Hello **Backpack**, I hope you're having a _great_ day!", style: .heading1))
+    }
+    
     func test_customColor() {
         assertSnapshot(
             BPKText("Hello Backpack, I hope you're having a great day!")
-                .foregroundColor(.primaryColor)
+                .foregroundColor(.coreAccentColor)
```
frugoman, line 50: deprecated colour!

Backpack-SwiftUI/Tests/Text/BPKTextTests.swift
```diff
@@ -40,10 +40,14 @@ class BPKTextTests: XCTestCase {
         assertSnapshot(BPKText("Hello Backpack, I hope you're having a great day!", style: .heading1))
     }
     
+    func test_withStyledText() {
+        assertSnapshot(BPKText("Hello **Backpack**, I hope you're having a _great_ day!", style: .heading1))
```
gert-janvercauteren, line 44: Does this markdown formatting translate to iOS 14 as well?

frugoman: It does not, so It'd be up for the consumer in this case to determine if they should or should not use it.
Although we're getting rid of 14 now, so it does not worry me a whole lot! but up for debate!

Backpack-SwiftUI/Text/Classes/BPKText.swift
```diff
@@ -25,15 +25,15 @@ import SwiftUI
...
+    private var text: LocalizedStringKey
     private var style: BPKFontStyle
     
     private var textColor = Color(BPKColor.textPrimaryColor)
     private var lineLimit: Int? = 1
     private var strikethrough: Bool = false
     
     public init(_ text: String, style: BPKFontStyle = .bodyDefault) {
-        self.text = text
+        self.text = .init(text)
```
gert-janvercauteren, line 36: Can we use the other LocalizedStringKey(text) init to make it clearer?

gert-janvercauteren (approved):

PR #1857, opened 2024-01-09
SONIC-1959: Change FloatingNotificationView to public 

dennis-li-skyscanner:
# Change FloatingNotificationView to public for updating the accessibilityElements array 

## The issue I am fixing
- I am showing a BPKFloatingNotifcation from a view where it has a fixed array of `view.accessibilityElements = [navigationHeaderView, tableView]`
- When I call `showNotification`, the button is not selectable by voice-over because the newly added FloatingNotificationView is not an accessibilityElement.
```swift
BPKFloatingNotification.show(
            .titleWithIconAndAction(
                parentView: view, // <- view has fixed accessibilityElements array
                title: title,
                iconName: .tickCircle,
                action: .action(title: actionTitle, action: action)
            ))
```
- So, I added the following function to add the following code to update the accessibilityElements array
  - ```swift
      public override func viewDidLayoutSubviews() {
        super.viewDidLayoutSubviews()

        // ⭐️ the line below requires public FloatingNotificationView
        let accessibilityElementsContainsFloatingNotification = view.accessibilityElements?.contains(where: { $0 is FloatingNotificationView })
        // Set the floating Notification view as accessible
        if view.subviews.count > view.accessibilityElements?.count ?? 0 &&
            accessibilityElementsContainsFloatingNotification == false,
           let floatingNotificationView = view.subviews.filter({ $0 is FloatingNotificationView }).first { // ⭐️ This also requires public FloatingNotificationView
            // Add floatingNotificationView when it is not in the accessibilityElements
            view.accessibilityElements?.append(floatingNotificationView)
        } else if view.subviews.count < view.accessibilityElements?.count ?? 0 &&
                    accessibilityElementsContainsFloatingNotification == true {
            // Remove floatingNotificationView when it is removed from the view hierarchy
            view.accessibilityElements?.removeAll { $0 is FloatingNotificationView } // ⭐️ This also requires public FloatingNotificationView
        }
    }
     ```
- As a result, I would like to change FloatingNotificationView's access level to `public`

----
## Other Consideration 
An alternative solution is to move the accessibilityElements update logic in backpack-ios. 

For example:
 - update parentView's  accessibilityElements after `addSubView()` 
   - `viewModel.parentView.accessibilityElements?.append(self)`
 - update parentView's  accessibilityElements before `removeFromSuperView()`
   - `superview.accessibilityElements?.removeAll { $0 == self }`

Let me know if you prefer the current solution "setting a view class to public" OR adding the `accessibilityElements` append & remove logic in Backpack-ios.

+ [x] `README.md`
+ [x] Tests

### DIFF
--- a/Backpack/FloatingNotification/Classes/FloatingNotificationView.swift
+++ b/Backpack/FloatingNotification/Classes/FloatingNotificationView.swift
@@ -19,7 +19,7 @@
 import Foundation
 import UIKit
 
-final class FloatingNotificationView: UIView {
+public final class FloatingNotificationView: UIView {
     
     private let shouldAnimateDown: ((FloatingNotificationView) -> Void)
     private let hiddenBottomConstraintConstant: CGFloat


gert-janvercauteren:
Hi @dennis-li-skyscanner,
Thanks for raising a PR, have you considered changing the show(_) function to return the floatingNotification view (as UIView).

This was we don't have to expose internal UI and you can achieve similar results as the example code you added.

dennis-li-skyscanner:
I updated the file to return an UIView. 

Thank for the suggestion!

dennis-li-skyscanner:
The same changes is also tested in the backpack-ios sample project with the following temporary changes.

```swift 
        // Lock the accessibilityElements like FlightsConfigs
        view.accessibilityElements = [stackView]
```


```swift 
// update accessibilityElements when appear  and disappear 
        let lastFloatingNotificationView = BPKFloatingNotification.show(
            .titleWithAction(
                parentView: view,
                title: "Saved",
                action: .action(title: "View", action: action),
                didDismiss: { [weak self] in
                    guard let strongSelf = self else { return }
                    strongSelf.view.accessibilityElements = [strongSelf.stackView]
                }
            ))

        // update accessibilityElements when appear
        if let lastFloatingNotificationView = lastFloatingNotificationView {
            view.accessibilityElements = [stackView, lastFloatingNotificationView]
        }

```
Screenshot 2024-01-15 at 10 51 55


## Before 
(View button is not selectable by voice-over when the accessibilityElements is fixed)
https://github.com/Skyscanner/backpack-ios/assets/102601094/a477d4f5-9a25-4c4f-93bf-3ad5c7ebca1e



## After 
https://github.com/Skyscanner/backpack-ios/assets/102601094/d563ab78-fe6a-42df-b6f6-240937c85552

gert-janvercauteren (approved):
LGTM!

PR #1868, opened 2024-01-26
Update snapshot test results on CI to reduce toil

gert-janvercauteren:
# Snapshot tests

CI runs on intel and our local devices run on M1. This causes friction with snapshot tests and leads to flaky results. We introduce a way for CI to retake snapshots and then require manual verification by the engineer.

Once CI pushes the screenshot, we still need to push an empty commit to trigger CI and finish the PR.

+ [ ] `README.md`
+ [ ] Tests


.github/workflows/_build.yml
```diff
@@ -133,10 +133,75 @@ jobs:
...
+        id: screenshotTestsUIKit
+        run: ./scripts/ci uikit test
+        continue-on-error: true
+
+      - name: Run SwiftUI tests
+        id: screenshotTestsSwiftUI
+        run: ./scripts/ci swiftui test
+        continue-on-error: true
+
+      - name: Commit changes
```
gert-janvercauteren, line 188: This needs extra checks (if PR + file changed)

gert-janvercauteren: + only add .png files and folders.
+ maybe move all snapshots to an easier location for deletion?

frugoman: you can git add like this I believe: `git add */**/__Snapshots__/*.png`

.github/workflows/_build.yml
```diff
@@ -175,10 +174,76 @@ jobs:
...
+      - name: Select Xcode 14
+        run: sudo xcode-select --switch /Applications/Xcode_14.2.app
+      
+      - name: Screenshot Tests
+        id: screenshotTests
+        run: |
+          ./scripts/record_screenshot_tests.sh
+
+          changedFiles=`git status --porcelain` && echo "CHANGED_FILES=${changedFiles//$'\n'/'%0A'}" >> $GITHUB_OUTPUT
+        continue-on-error: true
```
marianeum, line 222: is this needed because the step "fails" when snapshots change? what happens if there's a build error? we currently have a (slightly similar/related) issue on android where if that happens we delete all snapshots as none get generated. we don't have `continue-on-error` but the step didn't fail as it was missing the `set -e` at the top

.github/workflows/_build.yml
```diff
@@ -175,10 +174,76 @@ jobs:
...
+
+      - name: Check snapshot changes
+        id: checkSnapshotChanges
+        if: ${{ github.event_name != 'pull_request' }}
+        run: ./scripts/lint
+
+      - name: Commit changes
```
marianeum, line 229: another thing to consider as an option which is something i wanna consider on android as well: you could have two modes: by default you just use the snapshot tests as tests (ie just verify nothing has changed) and share the test output in a nice format - most snapshot testing libraries have that by default. then you can add a trigger for actually re-recording the snapshots if intentional. that way you get fewer unintentional changes - especially with flaky changes.
again depends on your tooling, duration etc - on android we switched to the current approach as the tests were very slow, so waiting for the initial failure to then trigger the actual change would take forever, but it also results in that confusing behaviour where because the screenshots changed your CI status is pending for all steps. Now that snapshot tests are a lot faster & stable the standard approach of treating CI as a test by default might be better again

.github/workflows/_build.yml
```diff
@@ -175,10 +174,76 @@ jobs:
...
+          git checkout $GITHUB_HEAD_REF
+          git add */**/__Snapshots__/*.png
+          git diff-index --quiet HEAD || git commit -m "Updated snapshots"
+          git push
+
   UploadArtifacts:
     name: Upload Artifacts
     runs-on: macos-latest
-    needs: [TestPods]
+    needs: [TestPods, Screenshots]
```
frugoman, line 246: so you need to upload these? wouldn't they be part of the pr already? (or it's in case they fail probably right?)

scripts/ci
```diff
@@ -30,9 +30,11 @@ function runTests() {
 function check_modules() {
     if [ $2 == "test" ]; then
         if [ $1 == "uikit" ]; then
-            runTests "Backpack-Snapshot-Tests"
             runTests "Backpack-Unit-Tests"
 
+        elif [ $1 == "uikitsnapshot" ]; then
```
frugoman, line 35: should this be also called from the workflow that runs the snapshot tests?

.github/workflows/_build.yml
```diff
@@ -132,15 +132,14 @@ jobs:
...
-
   TestPods:
     name: Testing Pods
     runs-on: macos-latest
     needs: [Setup]
     strategy:
       fail-fast: false
       matrix:
-        schemes: [uikit, swiftui]
+        schemes: [uikit]
```
frugoman, line 142: was this intentional?

.github/workflows/_build.yml
gert-janvercauteren: Just want to create a 'flow' where upload artifacts is the last step.

.github/workflows/_build.yml
gert-janvercauteren: This was, as SwiftUI doesn't have a separate 'snapshot' target. 
Should we introduce that? 

UIkit: SnapshotTest / UnitTest
SwiftUI: SnapshotTest / UnitTest

scripts/record_screenshot_tests.sh
```diff
@@ -0,0 +1,25 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+set -e
+
+find . -type d -name "__Snapshots__" -exec rm -r {} +;
+
+# Get the directory of the currently running script
+SCRIPT_DIR="$(dirname "$(readlink -f "$0")")"
+
+"$SCRIPT_DIR/ci" uikitsnapshot test || true;
```
gert-janvercauteren, line 24: @frugoman We run uikitsnapshot here as we only want to run snapshot tests

gert-janvercauteren:
take snapshot

PR #1890, opened 2024-03-06
Add NavBar component and tests

frugoman:
Figma: https://www.figma.com/file/irZ3YBx8vOm16ICkAr7mB3/Backpack-Components?type=design&node-id=4522-2521&mode=design&t=UP3vTrAyFNtG15Zg-0

This pull request adds the NavBar component and its corresponding tests to the project.

|Mode|Default|Transparent|Surface Contrast|
|---|---|---|---|
|Day Mode|iPhone SE (3rd generation)-swiftui_navbar___default_lm|iPhone SE (3rd generation)-swiftui_navbar___transparent_lm|iPhone SE (3rd generation)-swiftui_navbar___surfaceContrast_lm|
|Night Mode|iPhone SE (3rd generation)-swiftui_navbar___default_dm|iPhone SE (3rd generation)-swiftui_navbar___transparent_dm|iPhone SE (3rd generation)-swiftui_navbar___surfaceContrast_dm|


frugoman:
The snapshots tests can't capture the navbar's functionality well, apparently because it's not really a straight view, but it needs more context, a scene or window maybe. We can only use snapshots to validate the style of the background of the navbar (just a little bit though). I'd keep them just for that, but the value they provide is limited and we'll need to be extra careful while making changes to it in the future

gergana-st-ivanova (approved):

Backpack-SwiftUI/NavigationBar/Classes/BPKNavigationView.swift
```diff
@@ -0,0 +1,140 @@
...
+        self.leadingItems = leadingItems
+        self.trailingItems = trailingItems
+        self.content = content
+        
+        let appearance = UINavigationBarAppearance()
+        appearance.configureWithOpaqueBackground()
+        appearance.titleTextAttributes = [.foregroundColor: BPKColor.statusSuccessFillColor.value]
+        appearance.largeTitleTextAttributes = [.foregroundColor: BPKColor.statusSuccessFillColor.value]
+        UINavigationBar.appearance().standardAppearance = appearance
+        UINavigationBar.appearance().scrollEdgeAppearance = appearance
```
mcomisso, line 46: Question on these lines, wouldn't all navigation bars in the app change after we set this?

frugoman: My understanding is NO, it should not change other navbars in the app. Testing in the Backpack Sample App, this does not change other navbars. It would only affect navigation bars that has not been configured before hand. Besides this, in the Skyscanner app we currently have 0 native navigation bars (to my knowledge), they're all custom views placed on top of the view controllers.

If after releasing this change we see it messes with the app in any way we should make a release removing this change!

brunomdac (approved):

PR #1892, opened 2024-03-08
[TEMPURA-1242] Remove references to UIApplication.shared

darioroa:
# UIApplication References
As we want to use Backpack on application extensions, it is required that references to the `shared` property of `UIApplication` are removed, otherwise Backpack fails to compile.

Changes introduced: 
- `UIKitDialogContainerViewModifier` now has an optional `rootViewController` argument
- `LegacyBottomSheetContainerViewModifier` now has an optional `rootViewController` argument
- Both example views pass the same `rootViewController` previously declared inside each of those modifiers.

+ [ ] `README.md`
+ [ ] Tests

### DIFF
--- a/Backpack-SwiftUI/BottomSheet/Classes/BPKBottomSheet.swift
+++ b/Backpack-SwiftUI/BottomSheet/Classes/BPKBottomSheet.swift
@@ -41,6 +41,7 @@ public extension View {
         closeButtonAccessibilityLabel: String? = nil,
         title: String? = nil,
         action: BPKBottomSheetAction? = nil,
+        rootViewController: UIViewController?,
         @ViewBuilder bottomSheetContent: @escaping () -> BottomSheetContent
     ) -> some View {
         if #available(iOS 16.0, *) {
@@ -64,7 +65,8 @@ public extension View {
                     closeButtonAccessibilityLabel: closeButtonAccessibilityLabel,
                     title: title,
                     action: action,
-                    bottomSheetContent: bottomSheetContent
+                    bottomSheetContent: bottomSheetContent,
+                    rootViewController: rootViewController
                 )
             )
         }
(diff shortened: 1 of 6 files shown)


Backpack-SwiftUI/Dialog/Classes/BPKDialog.swift
```diff
@@ -27,6 +27,7 @@ public extension View {
         text: String,
         confirmButton: BPKDialogButton,
         secondaryActions: BPKDialogSecondaryActions? = nil,
+        rootViewController: UIViewController? = nil,
```
frugoman, line 30: I wonder if these need to be required (non optional), as if you call this BPKDialog with a nil rootViewController it will just not be shown. If this is the case, the API would be allowing broken behaviour to be defined.

When this was internal code, then it made more sense for the method that returned it to be optional, but now that this is exposed we should allow for no invalid state to be sent (ie, a nil rootViewController)

Backpack-SwiftUI/BottomSheet/Classes/BPKBottomSheet.swift
```diff
@@ -41,6 +41,7 @@ public extension View {
         closeButtonAccessibilityLabel: String? = nil,
         title: String? = nil,
         action: BPKBottomSheetAction? = nil,
+        rootViewController: UIViewController?,
```
frugoman, line 44: left a comment below about the optionality of this, but also: Can we rename the property to be `presentingController` or similar to that? it would reflect what the VC will be used for. `rootViewController` alludes to what we used internally, but may not be the case for clients.

darioroa: Completely agree, much clearer!

Backpack-SwiftUI/Dialog/Classes/BPKDialog.swift
frugoman: yeah I think we need to make it a breaking change here and fix the parts in the app that use this (only 1 I believe in Hotels?)

darioroa: Updated: 
be2494a72253b7e7b25cea483a76febc6a2f2dae

Backpack-SwiftUI/BottomSheet/Classes/BPKBottomSheet.swift
darioroa: Updated 👍 
af27a9d4f4c7e6b9c82c4aa5bbbcbc299aec703c

frugoman (approved):
Looks good! We'll then need to make the changes in the clients after this is released!

PR #1893, opened 2024-03-08
[LMN-844] Add BPKRatingBar SwiftUI component

frugoman:
This pull request adds the BPKRatingBar SwiftUI component to the project. The BPKRatingBar allows users to rate items on a scale and provides customization options for the scale and style of the rating bar.
|Day Mode|Night Mode|
|---|---|
|iPhone-swiftui_rating-bar___default_lm|iPhone-swiftui_rating-bar___default_dm|


Backpack-SwiftUI/RatingBar/Classes/BPKRatingBar+Style.swift
```diff
@@ -0,0 +1,41 @@
...
+        
+        var ratingBarBackgroundColor: BPKColor {
+            switch self {
+            case .default:
+                return BPKColor(value: UIColor(red: 0.88, green: 0.89, blue: 0.9, alpha: 1))
+                    .darkVariant(.surfaceElevatedColor)
+            case .onContrast:
+                return BPKColor(value: .white)
+                    .darkVariant(BPKColor(value: UIColor(red: 0.14, green: 0.2, blue: 0.27, alpha: 1)))
+            }
```
gert-janvercauteren, line 38: Should we use internal tokens for these? I'm worried this will make it hard to change the design colours later on

Backpack-SwiftUI/RatingBar/Classes/BPKRatingBar.swift
```diff
@@ -0,0 +1,146 @@
...
+                GeometryReader { proxy in
+                    Capsule()
+                        .foregroundColor(.corePrimaryColor)
+                        .frame(width: proxy.size.width * CGFloat(calculateValue()))
+                }
+            }
+            .frame(height: .md)
+        }
+        .accessibilityElement(children: .combine)
+        .accessibilityLabel(accessibilityLabel)
```
gert-janvercauteren, line 77: This might need to be `.ignore`. Setting it to combine, but then override the label might lead to future problems

Backpack-SwiftUI/RatingBar/Classes/BPKRatingBar.swift
```diff
@@ -0,0 +1,146 @@
...
+        self.style = style
+    }
+    
+    public var body: some View {
+        VStack(spacing: .md) {
+            HStack(alignment: .lastTextBaseline, spacing: .zero) {
+                BPKText(label, style: .footnote)
+                Spacer()
+                if showScale {
+                    BPKText("\(rating)", style: .heading5)
```
gert-janvercauteren, line 59: Do we add any rounding? 4.559595 is probably not what we want to display. 
We can be smart about this offering no decimal, 1, 2, <unlimited> ?

Backpack-SwiftUI/RatingBar/README.md
```diff
@@ -0,0 +1,57 @@
...
+- `default` (default)
+- `onContrast`
+
+### On Contrast
+
+```swift
+BPKRatingBar(
+    label: "Rating",
+    value: 4.5,
+    accessibilityLabel: "Rating of 4.5",
```
gert-janvercauteren, line 54: Should it be "Rating of 4.5 out of 5" to ensure the label examples include the scale?

Backpack-SwiftUI/RatingBar/Classes/BPKRatingBar.swift
```diff
@@ -0,0 +1,146 @@
...
+    public init(
+        label: String,
+        rating: Float,
+        accessibilityLabel: String,
+        ratingScale: RatingScale = .zeroToFive,
+        showScale: Bool = true,
+        style: Style = .default
+    ) {
+        self.label = label
+        self.rating = rating
```
gert-janvercauteren, line 46: We probably want to add some clamping to keep the rating between the scale? I saw the tests with a rating of -2.5 or 6.0/5 and they don't make sense from a customer perspective

Example/Backpack/SwiftUI/Components/RatingBar/RatingBarExampleView.swift
```diff
@@ -0,0 +1,65 @@
...
+            )
+            BPKRatingBar(
+                label: "Rating",
+                rating: 2.5,
+                accessibilityLabel: "Rating",
+                showScale: false,
+                style: style
+            )
+        }
+        .padding()
```
gert-janvercauteren, line 57: Can we make the accessibility label more descriptive of the actual use. 
If we just do 'rating' we set ourselves up for a11y failures when people fail to understand this will be the only thing that is read out. 

I would even venture and say that we might even miss the leading label as well. 
Should this a11y label only cover the 4.0/5 bit and then we group the label + score as an a11y label?

gert-janvercauteren (commented):
Left some a11y related comments. My main concern is that the a11y label is not clear how it is used. Can we make the api more foolproof?

gert-janvercauteren:
@frugoman Can we also make the docs screenshot match figma more and give better examples of usage? (best to ensure we sync across all 3 platforms)

Backpack-SwiftUI/RatingBar/Classes/BPKRatingBar.swift
```diff
@@ -0,0 +1,146 @@
...
+        self.showScale = showScale
+        self.style = style
+    }
+    
+    public var body: some View {
+        VStack(spacing: .md) {
+            HStack(alignment: .lastTextBaseline, spacing: .zero) {
+                BPKText(label, style: .footnote)
+                Spacer()
+                if showScale {
```
gert-janvercauteren, line 58: According to the API table, showScale only hides the `/5` bit and not the entire 4.5/5 :D

Backpack-SwiftUI/RatingBar/Classes/BPKRatingBar.swift
frugoman: I'll round to 1 decimal, to match Figma!

Backpack-SwiftUI/RatingBar/Classes/BPKRatingBar.swift
frugoman: Sounds right! thanks!

Backpack-SwiftUI/RatingBar/Classes/BPKRatingBar.swift
frugoman: Correct!!!

Backpack-SwiftUI/RatingBar/Classes/BPKRatingBar.swift
frugoman: I saw this, but ended up not handling it within backpack to let this logic be in the client.
I will handle it within backpack now though as I see the value having it!

Backpack-SwiftUI/RatingBar/README.md
frugoman: nice one yeah

Example/Backpack/SwiftUI/Components/RatingBar/RatingBarExampleView.swift
frugoman: I'd say lets wrap the entire thing, to allow for features to customise what they say. But also the 'out of' part would still be needed to be provided, I believe one prop is easier to deal with this

gergana-st-ivanova (approved):
Looks good to me! Will be interesting to see how that is used. 🎉

PR #1901, opened 2024-03-15
Create BPKImageGalleryPreview component

frugoman:
[Figma Link](https://www.figma.com/file/yN0hFyZlKL0Jwbpi0rEKYT/Backpack-Beta?type=design&node-id=5116-27794&mode=design)

This component works pretty much like the gallery, but also supports tapping on the current image out of the box.
Also pushes the page indicator up, as it's meant to be used with an overlayed view, as a header

image

|Day mode|Night Mode|
|---|---|
|image|image|


brunomdac (approved):
Looking good 😎

PR #1904, opened 2024-03-18
Fixing scrolling issue and cleanup

frugoman:
[No jira]
Cleaning up the Navbar's code and also fixing an issue where the scrollview clips the bottom content because we're manually dealing with the offsets.


gergana-st-ivanova (approved):
Looks good!

PR #1921, opened 2024-03-27
OGM-2374: New component - Navigation Tab Group

melwaraki:
Similar to a recent PR on [backpack-android](https://github.com/Skyscanner/backpack-android/pull/1902), this PR introduces a new NavigationTab component along with a NavigationTabGroup component.

It's similar to the current Chip component, but with a few changes in design and logic. 

| Light Mode | Dark Mode |
|------------|-----------|
| [Simulator Screenshot - iPhone 14 Pro - 2024-03-26 at 20 55 34](https://github.com/Skyscanner/backpack-ios/assets/25390815/645af166-18d6-40b8-a42c-adb2cded4241) | [Simulator Screenshot - iPhone 14 Pro - 2024-03-26 at 20 55 37](https://github.com/Skyscanner/backpack-ios/assets/25390815/739aab0f-3c87-4123-96b9-9c9fc89ca6fc) |

+ [ ] `README.md`
+ [x] Tests


Backpack-SwiftUI/NavigationTab/Classes/BPKNavigationTab.swift
```diff
@@ -0,0 +1,80 @@
...
+            )
+        )
+        .accessibilityAddTraits(selected ? [.isSelected] : [])
+        .if(!BPKFont.enableDynamicType, transform: {
+            $0.sizeCategory(.large)
+        })
+    }
+}
+
+#Preview {
```
frugoman, line 67: We're using Xcode 14 in Github Actions so this will fail, you'll need an 'old school' Preview Provider hehe

Backpack-SwiftUI/NavigationTab/Classes/BPKNavigationTab.swift
```diff
@@ -0,0 +1,80 @@
...
+    ) {
+        self.text = text
+        self.icon = icon
+        self.selected = selected
+        self.style = style
+        self.onClick = onClick
+    }
+    
+    public var body: some View {
+        Button(action: onClick) {
```
frugoman, line 44: Can you refactor this to just use a BPKChip? it looks the exact same except for the corner radious, which you can make the current BPKChipStyle struct receive that size as a parameter so you dont need to duplicate that file (or this one for that matter)

frugoman (commented):
From looking at android's code, it seems that you don't want to expose single tab views, but the entire tab itself, as in, receiving a list of tab items and then the bpk component would render the list of 'chips' to make the whole tab.

Keen to chat about this if needed!

Backpack-SwiftUI/NavigationTab/Classes/BPKNavigationTab.swift
melwaraki: Ah gg okay haha I'll switch it to the good ol' way

Backpack-SwiftUI/NavigationTab/Classes/BPKNavigationTab.swift
frugoman: no, lets NOT do this.

Backpack-SwiftUI/NavigationTab/Classes/BPKNavigationTabStyle.swift
```diff
@@ -0,0 +1,23 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import SwiftUI
+
+public enum BPKNavigationTabStyle {
```
frugoman, line 21: Can you make this an extension of the BPKNavigationTab? and then just call it `Style`

```swift
extension BPKNavigationTab {
  enum Style {
    .....
  }
}

File: Backpack-SwiftUI/NavigationTabGroup/Classes/BPKNavigationTabGroup.swift [hunk 4]
```diff
@@ -0,0 +1,112 @@
...
+            selected: selectedIndex == index,
+            style: style
+        ) {
+            onItemClick(index)
+        }
+    }
+}
+
+public extension BPKNavigationTabGroup {
+    struct NavigationTabItem: Hashable {
```
frugoman, line 66: why make this hashable?

Backpack-SwiftUI/NavigationTab/Classes/BPKNavigationTab.swift
```diff
@@ -0,0 +1,80 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import SwiftUI
+
+struct BPKNavigationTab: View {
```
frugoman, line 21: could you move these files into a single NativationTab folder? both the group and the tab into a single folder.

Backpack-SwiftUI/NavigationTab/Classes/BPKNavigationTab.swift
```diff
@@ -0,0 +1,80 @@
...
+        self.onClick = onClick
+    }
+    
+    public var body: some View {
+        Button(action: onClick) {
+            HStack(spacing: .md) {
+                if let icon {
+                    BPKIconView(icon)
+                }
+                Text(text)
```
frugoman, line 49: could you use a BPKText here instead? I see the Chip uses a Text as well, but dont quite remember why.

Backpack-SwiftUI/NavigationTab/Classes/BPKNavigationTab.swift
```diff
@@ -0,0 +1,80 @@
...
+
+struct BPKNavigationTab: View {
+    
+    let text: String
+    let icon: BPKIcon?
+    let selected: Bool
+    let style: BPKNavigationTabStyle
+    let onClick: () -> Void
+    
+    public init(
```
frugoman, line 29: this shouldn't be public

Backpack-SwiftUI/NavigationTabGroup/Classes/BPKNavigationTabGroup.swift
```diff
@@ -0,0 +1,112 @@
...
+    private let tabs: [NavigationTabItem]
+    private let style: BPKNavigationTabStyle
+    private let onItemClick: (_ index: Int) -> Void
+    
+    @Binding private var selectedIndex: Int?
+    
+    public init(
+        tabs: [NavigationTabItem],
+        style: BPKNavigationTabStyle = .default,
+        selectedIndex: Binding<Int?>,
```
frugoman, line 32: is it really possible to have a nil selection?  as in.. is taht an actual use case that makes sense to have? if not, lets not make this optional, if it is, then this is fine

Example/Backpack/SwiftUI/Components/NavigationTabGroup/NavigationTabGroupExampleView.swift
```diff
@@ -0,0 +1,61 @@
...
+        BPKNavigationTabGroup(
+            tabs: tabs,
+            style: .onDark,
+            selectedIndex: $selectedNavigationTab) { _ in }
+            .padding()
+            .background(.surfaceContrastColor)
+    }
+}
+
+#Preview {
```
frugoman, line 59: same as the other one

Backpack-SwiftUI/NavigationTabGroup/Classes/BPKNavigationTabGroup.swift
frugoman: I see on figma there's a design for having nothing selected as an example, but in a real use case I don;t think this makes sense, I'd consider making it non-optional

Backpack-SwiftUI/NavigationTabGroup/Classes/BPKNavigationTabGroup.swift
frugoman, line 66: could you rename it to just `Item` its a nested type so we know it's a NavigationTabItem already

Backpack-SwiftUI/Text/Classes/BPKText.swift
```diff
@@ -31,18 +31,26 @@ public struct BPKText: View {
     private var textColor = Color(BPKColor.textPrimaryColor)
     private var lineLimit: Int? = 1
     private var strikethrough: Bool = false
+    private var ignoreTextColorAndInheritFromParent: Bool = false
```
melwaraki, line 34: I'm not sure how you feel about this @frugoman? But I assume this is why BPKChip used Text rather than BPKText. We were forcing a `textColor` to be set rather than letting BPKText inherit from its parent just like a regular Text would. At least that's how it seems to me

frugoman (approved):
Looks good!! Thanks for contributing!

melwaraki:
Cloned into a non-fork repo: https://github.com/Skyscanner/backpack-ios/pull/1929

PR #1923, opened 2024-03-27
DON-384: Adjust place selector text field

yurareutskiy:
[Figma link](https://www.figma.com/file/yN0hFyZlKL0Jwbpi0rEKYT/Backpack-Beta?type=design&node-id=23566-3180&mode=design&t=S39eLd5xnE48YT7q-4)

- Add the new component BPKSearchInputSummary (naming and requirements came from the Design Team) based on BPKTextField with customisable `inputPrefix` which will allow us to set an additional visual enhancement for different verticals and their place selectors.
- `inputPrefix` can be in two state: icon, customText
- The new component will enable to develop a search input field without adding more complexity to existing BPKTextField
- (additional) fix setting a light mode for screenshot generating by setting .lightMode in case there is missing "FORCE_DARK_MODE" flag

| Light Mode | Dark Mode |
| --- | --- |
| Simulator Screenshot - iPhone SE (3rd generation) - 2024-03-26 at 22 46 15 | Simulator Screenshot - iPhone SE (3rd generation) - 2024-03-26 at 22 47 08 |
| Simulator Screenshot - iPhone SE (3rd generation) - 2024-03-26 at 22 46 38 | Simulator Screenshot - iPhone SE (3rd generation) - 2024-03-26 at 22 46 53 |

+ [ ] `README.md`
+ [ ] Tests


Backpack-SwiftUI/TextField/Classes/BPKTextField.swift
```diff
@@ -22,13 +22,19 @@ import SwiftUI
...
+    }
+    
     struct Icon {
         let icon: BPKIcon
         let color: BPKColor
     }
     
     @Binding private var text: String
     private let placeholder: String
+    private let prefixState: PrefixState?
```
frugoman, line 37: as above, the property should be just `prefix`

Backpack-SwiftUI/TextField/Classes/BPKTextField.swift
```diff
@@ -76,6 +86,22 @@ public struct BPKTextField: View {
...
+            switch prefixState {
+            case .text(let prefixText):
+                BPKText(prefixText, style: .bodyDefault)
+                    .foregroundColor(.textSecondaryColor)
+            case .searchIcon:
+                BPKIconView(BPKIcon(name: "search"))
+                    .foregroundColor(.textPrimaryColor)
+                    .accessibilityHidden(true)
+            case nil:
+                EmptyView()
```
frugoman, line 100: It's recommended to avoid using `EmptyView()`, please use `@ViewModifier` instead! Also you don't need an HStack here, you can do like so:

```swift
@ViewBuilder
private var prefixView: some View {
  if let prefix {
    switch prefix {
      case .text(let prefixText):
        BPKText(prefixText, style: .bodyDefault)
          .foregroundColor(.textSecondaryColor)
      case .searchIcon:
        BPKIconView(BPKIcon(name: "search"))
            .accessibilityHidden(true)
        }
    }
}

File: Example/Backpack/SceneDelegate.swift [hunk 3]
```diff
@@ -34,6 +34,8 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
         window.rootViewController = rootViewController(isUITestRun: isUITestRun)
         if ThemeHelpers.forceDarkMode() {
             window.overrideUserInterfaceStyle = .dark
+        } else {
+            window.overrideUserInterfaceStyle = .light
```
frugoman, line 38: I think this shouldn't be committed

Backpack-SwiftUI/Tests/AppSearchModal/BPKAppSearchModalTests.swift
```diff
@@ -54,14 +66,16 @@ class BPKAppSearchModalTests: XCTestCase {
     // MARK: - Helpers
     private func givenSut(
         with results: BPKAppSearchModalResults,
-        inputState: BPKAppSearchModal.TextFieldState = .default
+        inputState: BPKAppSearchModal.TextFieldState = .default,
+        prefixState: BPKTextField.PrefixState? = nil
```
gert-janvercauteren, line 70: @yurareutskiy the test fails to compile on this line. The naming has probably changed.

gert-janvercauteren (commented):
Can you make a recording how the prefix sounds with a screen reader? :D

Backpack-SwiftUI/TextField/Classes/BPKTextField.swift
yurareutskiy: Not using EmptyView() anymore, it was in a draft version. Agree about HStack, but the problem is in opaque types (because of the compiler can't know which type will be using in runtime here). However, I remembered about @ViewBuilder attribute which should help here with creating "conditional tree" of views. Thanks for noticing that.

Example/Backpack/SceneDelegate.swift
yurareutskiy: @frugoman Ah, that's a good question. 
I can remove it if you think it shouldn't be here. However, this line is helping with stability for snapshots. As we're setting only dark mode when there is a dark mode parameter in arguments. However, it can be a dark mode already on the device by default sometime, which makes screenshot for a light mode captured in a dark mode.
So, this line is just for better reliability here. I even noticed there was a wrong screenshot in the repo which was fixed by this change.
But if there is other complications which I don't see here, just let me know please!
image

Backpack-SwiftUI/Tests/AppSearchModal/BPKAppSearchModalTests.swift
yurareutskiy: Thanks! Still not sure why it was failing before I renamed that, but amazing as it's working:)

Backpack-SwiftUI/AppSearchModal/Classes/BPKAppSearchModal.swift
```diff
@@ -21,6 +21,7 @@ import SwiftUI
 public struct BPKAppSearchModal: View {
     let title: String
     @Binding var inputText: String
+    let inputPrefix: BPKSearchInputSummary.InputPrefix
```
frugoman, line 24: should this not be optional?

Example/Backpack/SceneDelegate.swift
frugoman: snapshots are now taken on CI actually! so not really a problem anymore 😁

frugoman: ah you mean the screenshots, yeah then this is fine!

Backpack-SwiftUI/SearchInputSummary/Classes/BPKSearchInputSummary.swift
```diff
@@ -0,0 +1,136 @@
...
+
+import SwiftUI
+
+/// A control that displays an editable text interface for search inputs.
+///
+/// Use `inputState(_ state: State)` to change the state of the text field.
+public struct BPKSearchInputSummary: View {
+    public enum InputPrefix {
+        case text(String)
+        case icon
```
frugoman, line 27: this should receive a BPKIcon as parameter, so that clients can define what icon they use (clients of this component may be within backpack, but at least it would be configurable from outside this very component)

Backpack-SwiftUI/AppSearchModal/Classes/BPKAppSearchModal.swift
frugoman: Alright, they should all have a prefix 👍

Backpack-SwiftUI/AppSearchModal/Classes/BPKAppSearchModal.swift
```diff
@@ -34,42 +35,42 @@ public struct BPKAppSearchModal: View {
...
+                AppSearchModalContentView(state: content, onScroll: onScroll(_:))
                     .padding(.top, .md)
             case .error(let error):
                 AppSearchModalErrorView(state: error)
                     .padding(.horizontal, .md)
             }
         }
+        .onAppear {
+            inputFieldIsFocussed.toggle()
+        }
```
frugoman, line 73: I'm not sure we want this within the component, for example, hotels does not want this. it should be removed from here and leave it up to the clients defining this

frugoman: (if this is something you want to have)

Backpack-SwiftUI/SearchInputSummary/Classes/BPKSearchInputSummary.swift
```diff
@@ -0,0 +1,136 @@
...
+    }
+    
+    @ViewBuilder
+    private var prefixView: some View {
+        switch inputPrefix {
+        case .text(let prefixText):
+            BPKText(prefixText, style: .bodyDefault)
+                .foregroundColor(.textSecondaryColor)
+        case .icon:
+            BPKIconView(BPKIcon(name: "search"))
```
frugoman, line 99: the icon should be passed by the ones using this component, shouldnt be hardcoded within. also, you should use the constant for the icon rather than a string for it

Backpack-SwiftUI/AppSearchModal/Classes/BPKAppSearchModal.swift
yurareutskiy: It seems, we it has not to be nil, otherwise it would be a regular BPKTextField according to requirements

frugoman (commented):
Looks good over all! just a couple of small modifications I think would be good to have

Backpack-SwiftUI/SearchInputSummary/Classes/BPKSearchInputSummary.swift
yurareutskiy: hmm, that's a good idea I guess to make it more flexible

Backpack-SwiftUI/AppSearchModal/Classes/BPKAppSearchModal.swift
yurareutskiy: yeah, 100% should be removed 😅

frugoman (approved):
Looks pretty good!! thanks for contributing!!

Backpack-SwiftUI/SearchInputSummary/Classes/BPKSearchInputSummary.swift
```diff
@@ -0,0 +1,135 @@
...
+            if case let .clear(accessibilityLabel, action) = state {
+                Button(action: action) {
+                    BPKIconView(icon.icon)
+                        .foregroundColor(icon.color)
+                }
+                    .accessibilityLabel(accessibilityLabel)
+            } else {
+                BPKIconView(icon.icon)
+                    .foregroundColor(icon.color)
+                    .accessibilityHidden(true)
```
novinfard-skyscanner, line 86: 👍

Backpack-SwiftUI/SearchInputSummary/Classes/BPKSearchInputSummary.swift
```diff
@@ -0,0 +1,135 @@
...
+            prefixView
+                .accessibilityHidden(true)
+            TextField(placeholder, text: $text)
+                .font(style: .bodyDefault)
+                .foregroundColor(state.textColor)
+                .disabled(state.isDisabled)
+                .focused($focused)
+            accessory
+        }
+        .padding(.md)
```
novinfard-skyscanner, line 68: Based on the design I think the padding is:
.padding(.left, .base)
.padding(.right, .md)
.padding(.vertical, .md)

frugoman: I thin it's actually `.padding(.horizontal, .base).padding(.vertical, BPKSpacing.md.value + BPKSpacing.sm.value)`  (we don't have a token for value 12 hehe) could you raise the pr so I can approve it myself?

novinfard-skyscanner: The padding is 16 (.base) from left [leading], and 12 (.md + .sm) for all other sides.
Screenshot 2024-03-28 at 16 42 37
Screenshot 2024-03-28 at 16 42 24

Please correct me if I read the Figma file incorrectly. we will make the PR for this change after that, no worries

PR #1964, opened 2024-05-22
[BPKBarChart] ACC-455/ACC-456 Resolve Accessibility defects

gert-janvercauteren:
# BPKBarChart

- Hide the key items (price/no price) as they add clutter
- Make month titles heading
- Fix accessibility label and value for each bar

| Before | After |
| --- | --- |
| [Video before](https://github.com/Skyscanner/backpack-ios/assets/728889/3423049c-5f68-4512-9b22-4fc8bc4e9c99) | [Video after](https://github.com/Skyscanner/backpack-ios/assets/728889/4955ba5f-b40c-461d-8b0f-1e6dcd7c9e54) |

+ [ ] `README.md`
+ [ ] Tests

### DIFF
--- a/Backpack/BarChart/Classes/BarChartBar.swift
+++ b/Backpack/BarChart/Classes/BarChartBar.swift
@@ -20,10 +20,10 @@ import UIKit
 
 @objcMembers
 @objc(BPKBarChartBar)
-public final class BPKBarChartBar: UIControl {
+public final class BPKBarChartBar: UIView {
 
     /// Whether the bar is selected or not
-    public override var isSelected: Bool {
+    public var isSelected: Bool = false {
         didSet {
             updateBarColor()
             updateminiFlareViewAppearance()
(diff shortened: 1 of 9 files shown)


Backpack/BarChart/Classes/BarChartCollectionViewCell.swift
```diff
@@ -26,7 +26,6 @@ internal final class BPKBarChartCollectionViewCell: UICollectionViewCell {
     lazy public var barChartBar: BPKBarChartBar = {
         let view: BPKBarChartBar = BPKBarChartBar()
         view.translatesAutoresizingMaskIntoConstraints = false
-        view.isUserInteractionEnabled = false
```
gert-janvercauteren, line 29: This would add 'dimmed' to the barchart read out for each bar

Backpack/BarChart/Classes/BarChartBar.swift
```diff
@@ -20,10 +20,10 @@ import UIKit
 
 @objcMembers
 @objc(BPKBarChartBar)
-public final class BPKBarChartBar: UIControl {
+public final class BPKBarChartBar: UIView {
```
gert-janvercauteren, line 23: Make the bar a simple UIView, we only used UIControl to facilitate isSelected as a property

mcomisso: Any cons in leaving this a UIControl? Given is an intractable element and UIControl inherits from UIView, seems alright to leave it like that?

gert-janvercauteren: The problem is that we then need to set userInteractionEnabled = false on it, to allow selection of the cell. 
There was no point in having it as a UIControl, if we do and disable interaction it will announce as disabled :/

mcomisso (approved):

PR #1974, opened 2024-05-31
DON-386: Add accessibility support to BPKNudger

yurareutskiy:
In the scope of improving feature tests of the traveller selector there is essential change for BPKNudger, which is using as a stepper component in TravellerSelector.

The problem is some small components like increment and decrement buttons are unavailable for clicking in UITests due to combining those elements into one Nudger accessibility element.

Those changes make smaller elements accessible without any changes for screen reader for actual accessibility users.  

+ [ ] `README.md`
+ [ ] Tests

### DIFF
--- a/Backpack-SwiftUI/Nudger/Classes/BPKNudger.swift
+++ b/Backpack-SwiftUI/Nudger/Classes/BPKNudger.swift
@@ -33,6 +33,7 @@ public struct BPKNudger: View {
     private var maxValue: Int
     private var step: Int
     private let minWidth: CGFloat = BPKSpacing.lg.value
+    @State var accessibilityPrefix: String?
     
     /// Creates a `BPKNudger`.
     /// - Parameters:
@@ -45,7 +46,9 @@ public struct BPKNudger: View {
     ///     Must be greater than `min`.
     ///   - step: The step value of the `BPKNudger`.
     ///     Defaults to `1`.
-    public init(value: Binding<Int>, min: Int, max: Int, step: Int = 1) {
+    ///   - accessibilityPrefix: The string which would be added to accessibility identifiers of `BPKNudger` components.
+    ///     Defaults to `nil`.
+    public init(value: Binding<Int>, min: Int, max: Int, step: Int = 1, accessibilityPrefix: String? = nil) {
         self.title = nil
         self.subtitle = nil
         self.icon = nil
@@ -53,6 +56,7 @@ public struct BPKNudger: View {
         minValue = min
         maxValue = max
         self.step = step
+        self.accessibilityPrefix = accessibilityPrefix
         self._value = value
     }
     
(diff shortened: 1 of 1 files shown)


Backpack-SwiftUI/Nudger/Classes/BPKNudger.swift
```diff
@@ -163,6 +174,13 @@ public struct BPKNudger: View {
         value = max(value - step, minValue)
         updateButtonStates()
     }
+    
+    private func accessibilityLabel(for label: String) -> String {
```
novinfard-skyscanner, line 178: Not an issue, just naming wise:
```private func accessibilityIdentifierl(for label: String) -> String {```

File: Backpack-SwiftUI/Nudger/Classes/BPKNudger.swift [hunk 2]
```diff
@@ -45,14 +46,17 @@ public struct BPKNudger: View {
     ///     Must be greater than `min`.
     ///   - step: The step value of the `BPKNudger`.
     ///     Defaults to `1`.
-    public init(value: Binding<Int>, min: Int, max: Int, step: Int = 1) {
+    ///   - accessibilityPrefix: The string which would be added to accessibility identifiers of  `BPKNudger` components.
+    ///     Defaults to `nil`.
+    public init(value: Binding<Int>, min: Int, max: Int, step: Int = 1, accessibilityPrefix: String? = nil) {
```
gert-janvercauteren, line 51: I would solve this using a modifier instead of adding it to the initializers.
There's a big chance we might miss this in future updates and a modifier would make it more robust api wise.

Backpack-SwiftUI/Nudger/Classes/BPKNudger.swift
```diff
@@ -33,6 +33,7 @@ public struct BPKNudger: View {
     private var maxValue: Int
     private var step: Int
     private let minWidth: CGFloat = BPKSpacing.lg.value
+    @State var accessibilityPrefix: String?
```
frugoman, line 36: I'd make this private

Backpack-SwiftUI/Nudger/Classes/BPKNudger.swift
```diff
@@ -163,12 +174,27 @@ public struct BPKNudger: View {
         value = max(value - step, minValue)
         updateButtonStates()
     }
+    
+    private func accessibilityIdentifier(for label: String) -> String {
+        if let prefix = accessibilityPrefix {
+            return "\(prefix)_\(label)"
```
frugoman, line 180: is the underscore needed? or should a space be better? also, how does this work for rtl languages? I'm not sure this is the right way to go about this. @gert-janvercauteren you thik that's a problem for voice over? or is VO smart enough?

yurareutskiy: that's a identifier, not a label, so, there won't be a case with the issue here :)

Backpack-SwiftUI/Nudger/Classes/BPKNudger.swift
yurareutskiy: initially it was private, but after bringing the extension with the modifier, it seems, it's no longer possible to keep it private
unless you see another approach

PR #2001, opened 2024-07-04
AI-600: Fix TextArea frame 

AlaaAmrAmin:
`TextArea` frame's `minHeight` is set to the equivalent of 3 lines, but this does not take into consideration the padding added between the `BPKText`/`TextEditor` and the frame. The incorrect `minHeight` sometimes activates the scrolling behaviour of the `TextEditor` when a third line is added, and after a fourth line is added, it starts to expand instead of scrolling. This fix ensures that the `TextArea` expands its height in all cases.

+ [ ] `README.md`
+ [ ] Tests


frugoman (approved):
Looks good! Thanks for fixing!

PR #2016, opened 2024-07-22
Make medium bottom sheet expansion configurable

frugoman:
This pull request adds the ability to configure the expansion behavior of the medium bottom sheet in the code. Previously, the medium bottom sheet was always expansible, but now it can be set to be non-expansible if desired. This provides more flexibility in customizing the behavior of the medium bottom sheet.


marianeum (approved):
looks good - let's do the third state as a separate PR

PR #2022, opened 2024-07-26
[ChAI] [AI-641]: Contribute BPKCardCarousel to backpack-ios

mattyr3200:

+ [x] `README.md`
+ [x] Tests


Backpack-SwiftUI/CardCarousel/Classes/BPKCardCarousel.swift
```diff
@@ -0,0 +1,173 @@
...
+        self.cardCount = content.count
+        self.cardWidth = size.width * 0.8
+        self.cards = content + content
+        self.onCardChange = onCardChange
+        
+        currentCardIndex = cardCount + startingIndex
+    }
+    
+    var body: some View {
+        VStack {
```
frugoman, line 73: Can you add the spacing value to the VStack using BPK values? like so:
```swift
VStack(spacing: .md) // same value as SwiftUI uses by default

File: Backpack-SwiftUI/CardCarousel/Classes/BPKCardCarousel.swift [hunk 2]
```diff
@@ -0,0 +1,173 @@
...
+    }
+    
+    private func onDragChanged (value: DragGesture.Value) {
+        totalDrag = value.translation.width
+        isDragging = true
+    }
+    
+    private func cardIndicator() -> some View {
+        return HStack(spacing: .md) {
+            ForEach(0..<cardCount, id: \.self) { index in
```
frugoman, line 127: Would the BPKPageIndicator work for this?

Backpack-SwiftUI/CardCarousel/Classes/BPKCardCarousel.swift
```diff
@@ -0,0 +1,173 @@
...
+internal struct InternalCardCarousel<Content: View>: View {
+    private let size: CGSize
+    private let cardCount: Int
+    private let cards: [Content]
+    private let onCardChange: () -> Void
+    
+    private let dragAnimation: Animation = .snappy(duration: 0.5)
+    private let cardWidth: CGFloat
+    
+    @State private var currentCardIndex: Int
```
frugoman, line 42: This shall be a Binding instead and have this value exposed to the parent, so that the client then has access to this property and it's the client that's in control over what index is selected.

Backpack-SwiftUI/CardCarousel/Classes/BPKCardCarousel.swift
```diff
@@ -0,0 +1,173 @@
...
+        startingIndex: Int,
+        onCardChange: @escaping () -> Void
+    ) {
+        self.size = size
+        self.cardCount = content.count
+        self.cardWidth = size.width * 0.8
+        self.cards = content + content
+        self.onCardChange = onCardChange
+        
+        currentCardIndex = cardCount + startingIndex
```
frugoman, line 69: On second thought, my above suggestion about making the currentCardIndex a Binding may not apply. as your calculations are done based on the dragging gesture and not the current index itself I believe? worth exploring though!

Backpack-SwiftUI/CardCarousel/Classes/BPKCardCarousel.swift
```diff
@@ -0,0 +1,214 @@
...
+}
+
+internal struct InternalCardCarousel<Content: View>: View {
+    @Binding private var currentIndex: Int
+    @State private var currentInternalIndex: Int
+    @AccessibilityFocusState private var focusOnCard: Int?
+    private let onCardChange: () -> Void
+    private let content: [Content]
+    
+    private let size: CGSize
```
AlaaAmrAmin, line 58: Do we need to store the size? It looks like we just use it in the init

Backpack-SwiftUI/CardCarousel/Classes/BPKCardCarousel.swift
```diff
@@ -0,0 +1,214 @@
...
+        self.cardCount = content.count
+        self.cardWidth = size.width * 0.8
+        self.content = content + content
+        self.onCardChange = onCardChange
+        
+        currentInternalIndex = cardCount + 1 + currentIndex.wrappedValue
+        
+        _currentIndex = currentIndex
+        
+        focusOnCard = .init(cardCount + 1 + currentIndex.wrappedValue)
```
AlaaAmrAmin, line 88: Would this work? 
```suggestion
        focusOnCard = currentInternalIndex
```

Example/Backpack Screenshot/SwiftUIScreenshots.swift
```diff
@@ -327,6 +327,11 @@ class SwiftUIScreenshots: BackpackSnapshotTestCase {
             saveScreenshot(component: "carousel-card", scenario: "default", userInterfaceStyle: userInterfaceStyle)
         }
         
+        await navigate(title: "Card Carousel") {
```
AlaaAmrAmin, line 330: I think this one is duplicated

mattyr3200: I think all components have two parts in this file 🙂

Backpack-SwiftUI/CardCarousel/Classes/BPKCardCarousel.swift
```diff
@@ -0,0 +1,209 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import SwiftUI
+
+public struct BPKCardCarousel<Content: View>: View {
+    @Binding private var currentIndex: Int
+    private let cards: [Content]
+    private let onCardChange: () -> Void
```
AlaaAmrAmin, line 24: Thinking out loud, If I am a client and I am interested in getting notified about card changes, I would want to get the card index as well. With the current API, the `onCardChange` would be triggered, then the client would need to get the index from the `currentIndex`. I think it would be better to bundle them together. Like this: `private let onCardChange: (Int) -> Void`, and we no longer need the `currentIndex` to be a `binding`. what do you think?

frugoman: I'd say it should be the other way around. only have a binding exposed and NOT an `onCardChange`. if clients want to run logic when the current Index change, they can add an `onChange($currentIndex)` on their side! This is the approach we follow in most components!

mattyr3200: We will still need the binding for the consumer to be allowed to set the starting index and control it from their side 🙂. Happy to change though if it seems a better way of handling the card change.

mattyr3200: I will remove onCardChanged and let the client handle it

frugoman (approved):
Excellent contribution!!! Thanks!

PR #2030, opened 2024-07-30
Add original copyright to files ported from external libraries

frugoman:
This pull request adds the original copyright to the files that were ported from external libraries. This ensures that the appropriate attribution is given to the original authors of the code.


gert-janvercauteren (approved):
Thank you!

PR #2041, opened 2024-08-06
A SwiftUI version of a BPKFlightLeg using a DynamicStack

michmaml:

+ [ ] `README.md`
+ [ ] Tests

### DIFF
--- a/Backpack-SwiftUI/DynamicStack/Classes/BPKDynamicStack.swift
+++ b/Backpack-SwiftUI/DynamicStack/Classes/BPKDynamicStack.swift
@@ -1,0 +1,42 @@
+import SwiftUI
+import Backpack_Common
+
+struct BPKDynamicStack<Content: View>: View {
+    @Environment(\.dynamicTypeSize) private var dynamicTypeSize
+
+    private let horizontalAlignment: HorizontalAlignment
+    private let verticalAlignment: VerticalAlignment
+    private let spacing: BPKSpacing
+    private let threshold: DynamicTypeSize
+    @ViewBuilder var content: () -> Content
+
+    public init(
+        horizontalAlignment: HorizontalAlignment = .center,
+        verticalAlignment: VerticalAlignment = .center,
+        spacing: BPKSpacing = .none,
+        threshold: DynamicTypeSize = .accessibility1,
+        content: @escaping () -> Content
+    ) {
+        self.horizontalAlignment = horizontalAlignment
+        self.verticalAlignment = verticalAlignment
+        self.spacing = spacing
+        self.threshold = threshold
+        self.content = content
+    }
+
+    var body: some View {
+         if dynamicTypeSize >= threshold {
+             VStack(
+                 alignment: horizontalAlignment,
+                 spacing: spacing,
+                 content: content
+             )
+         } else {
+             HStack(
+                 alignment: verticalAlignment,
+                 spacing: spacing,
+                 content: content
+             )
+         }
+    }
+}
--- a/Backpack-SwiftUI/FlightLeg/Classes/BPKFlightLeg.swift
+++ b/Backpack-SwiftUI/FlightLeg/Classes/BPKFlightLeg.swift
@@ -1,0 +1,177 @@
+import SwiftUI
+import Backpack_Common
+
+struct BPKFlightLeg: View {
+    private let airlineLogoSize: CGFloat = 24.0
+
+    private let departureArrivalTime: String
+    private let nextDayArrival: String?
+    private let flightDescription: AttributedString
+    private let stopsInfo: String
+    private let highlightStopsInfo: Bool
+    private let duration: String
+    private let operatedBy: String?
+    private let warning: String?
+    @State private var carrierLogo: Image?
+
+    // String-based init
+    public init(
+        departureArrivalTime: String,
+        nextDayArrival: String? = "",
+        flightDescription: String,
+        stopsInfo: String,
+        highlightStopsInfo: Bool,
+        duration: String,
+        operatedBy: String? = "",
+        warning: String? = "",
+        carrierLogo: UIImage? = nil
+    ) {
+        self.departureArrivalTime = departureArrivalTime
+        self.nextDayArrival = nextDayArrival
+        self.flightDescription = AttributedString(flightDescription)
+        self.stopsInfo = stopsInfo
+        self.highlightStopsInfo = highlightStopsInfo
+        self.duration = duration
+        self.operatedBy = operatedBy
+        self.warning = warning
+        if let carrierLogo = carrierLogo {
+            self._carrierLogo = State(initialValue: Image(uiImage: carrierLogo))
+        } else {
+            self._carrierLogo = State(initialValue: nil)
+        }
+    }
+
+    // AttributedString-based init
+    public init(
+        departureArrivalTime: String,
+        nextDayArrival: String? = "",
+        flightDescription: AttributedString,
+        stopsInfo: String,
+        highlightStopsInfo: Bool,
+        duration: String,
+        operatedBy: String? = "",
+        warning: String? = "",
+        carrierLogo: UIImage? = nil
+    ) {
+        self.departureArrivalTime = departureArrivalTime
+        self.nextDayArrival = nextDayArrival
+        self.flightDescription = flightDescription
+        self.stopsInfo = stopsInfo
+        self.highlightStopsInfo = highlightStopsInfo
+        self.duration = duration
+        self.operatedBy = operatedBy
+        self.warning = warning
+        if let carrierLogo = carrierLogo {
+            self._carrierLogo = State(initialValue: Image(uiImage: carrierLogo))
+        } else {
+            self._carrierLogo = State(initialValue: nil)
+        }
+    }
+
+    var body: some View {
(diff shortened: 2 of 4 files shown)


Backpack-SwiftUI/DynamicStack/Classes/BPKDynamicStack.swift
```diff
@@ -0,0 +1,42 @@
...
+    private let verticalAlignment: VerticalAlignment
+    private let spacing: BPKSpacing
+    private let threshold: DynamicTypeSize
+    @ViewBuilder var content: () -> Content
+
+    public init(
+        horizontalAlignment: HorizontalAlignment = .center,
+        verticalAlignment: VerticalAlignment = .center,
+        spacing: BPKSpacing = .none,
+        threshold: DynamicTypeSize = .accessibility1,
```
frugoman, line 17: could this be a modifier instead of an initializer prop? seems like an accessory thing to set rather than a hard requirement.

frugoman: Also giving it a more explicit name, threshold of what ?

Backpack-SwiftUI/DynamicStack/Classes/BPKDynamicStack.swift
```diff
@@ -0,0 +1,42 @@
...
+    private let horizontalAlignment: HorizontalAlignment
+    private let verticalAlignment: VerticalAlignment
+    private let spacing: BPKSpacing
+    private let threshold: DynamicTypeSize
+    @ViewBuilder var content: () -> Content
+
+    public init(
+        horizontalAlignment: HorizontalAlignment = .center,
+        verticalAlignment: VerticalAlignment = .center,
+        spacing: BPKSpacing = .none,
```
frugoman, line 16: could it be the case that spacing needs to be different in vertical and horizontal enviroments? maybe you could create a small Struct that takes this as a required parameter like this:
```swift
struct HorizontalDynamicAlignment {
  let alignment: HorizontalAlignment
  let spacing: BPKSpacing
}

// and

struct VerticalDynamicAlignment {
  let alignment: VerticalAlignment
  let spacing: BPKSpacing
}
```
and then receive this instead:
```swift
public init(
    horizontalAlignment: HorizontalDynamicAlignment = .center,
    verticalAlignment: VerticalDynamicAlignment = .center,
    content: @escaping () -> Content
) {}
```
or even simpler than all this,
```swift
public init(
    horizontalAlignment: (HorizontalAlignment, BPKSpacing) = (.center, .none),
    verticalAlignment: (VerticalAlignment, BPKSPacing) = (.center, .none),
    content: @escaping () -> Content
) {}

File: Backpack-SwiftUI/FlightLeg/Classes/BPKFlightLeg.swift [hunk 3]
```diff
@@ -0,0 +1,177 @@
...
+    private let highlightStopsInfo: Bool
+    private let duration: String
+    private let operatedBy: String?
+    private let warning: String?
+    @State private var carrierLogo: Image?
+
+    // String-based init
+    public init(
+        departureArrivalTime: String,
+        nextDayArrival: String? = "",
```
frugoman, line 20: This name is not very clear at describing what's this for, it should rather be named after what's the label intention, rather than the potential content. as in, is it a comment on the view, information badge, etc. if it's exclusively for something about the next day arrival, then it should be a bit clearer on what's expected to be passed here

Backpack-SwiftUI/FlightLeg/Classes/BPKFlightLeg.swift
```diff
@@ -0,0 +1,177 @@
...
+        flightDescription: String,
+        stopsInfo: String,
+        highlightStopsInfo: Bool,
+        duration: String,
```
frugoman, line 24: should this not be a number? if not, what's the expected formate of this string?

Backpack-SwiftUI/FlightLeg/Classes/BPKFlightLeg.swift
```diff
@@ -0,0 +1,177 @@
...
+                .frame(
+                    maxWidth: .infinity,
+                    alignment: .leading
+                )
+            createDurationInfo()
+        }
+    }
+
+    @ViewBuilder
+    private func createCarrierLogo() -> some View {
```
frugoman, line 84: could you change the name of these funcitons to remove the word 'create' from them? name them as if you were composing this view as any other view. just `carrierLogo` should be good! also if it does not receive params, it should probably be a `var` instead.
A slighly better way would be to make this a little struct internal to the BPKFlightLeg to contain this view.
```swift
struct CarrierLogo: View {
 ...
}

File: Backpack-SwiftUI/FlightLeg/Classes/BPKFlightLeg.swift [hunk 6]
```diff
@@ -0,0 +1,177 @@
...
+        VStack(alignment: .leading, spacing: .sm) {
+            if let carrierLogo = carrierLogo {
```
frugoman, line 86: this if let should be a parent of the VStack I believe, else it will draw an empty VStack, which doesnt seem to make a loot of sense (maybe it does because of the padding.top you're setting?)

frugoman: Also, this if let can be written like this: 
```swift
if let carrierLogo { // if the property name is the same it can be ignored!
}

File: Backpack-SwiftUI/FlightLeg/Classes/BPKFlightLeg.swift [hunk 7]
```diff
@@ -0,0 +1,177 @@
...
+
+    private func departureArrivalTimeAttributedText() -> AttributedString {
+        var str = AttributedString(departureArrivalTime)
+        str.font = BPKFontStyle.heading5.font
+
+        return str
+    }
+
+    private func nextDayAttributedText() -> AttributedString {
+        var str = AttributedString(nextDayArrival ?? "")
```
frugoman, line 142: why create this attributedString if the nextDayArrival is empty/nil? should this not return an optional attributedString?

Backpack-SwiftUI/FlightLeg/Classes/BPKFlightLeg.swift
```diff
@@ -0,0 +1,177 @@
...
+        VStack(alignment: .trailing, spacing: .sm) {
+            BPKText(stopsInfo, style: .label3)
+                .foregroundColor(highlightStopsInfo ? .textErrorColor : .textPrimaryColor)
+
+            BPKText(duration, style: .caption)
+                .foregroundColor(.textSecondaryColor)
+        }
+    }
+
+    private func departureArrivalTimeAttributedText() -> AttributedString {
```
frugoman, line 134: all these could be `var`s instead of funcs

Backpack-SwiftUI/FlightLeg/Classes/BPKFlightLeg.swift
```diff
@@ -0,0 +1,177 @@
...
+
+    private let departureArrivalTime: String
+    private let nextDayArrival: String?
+    private let flightDescription: AttributedString
+    private let stopsInfo: String
+    private let highlightStopsInfo: Bool
+    private let duration: String
+    private let operatedBy: String?
+    private let warning: String?
+    @State private var carrierLogo: Image?
```
frugoman, line 15: does this need to be `@State`? seems it does not, a regular `let` should suffice?
also, you should store the UIImage instead of an Image and create the Image(uiImage: image) on the view's body instead I think.

Backpack-SwiftUI/FlightLeg/Classes/BPKFlightLeg.swift
frugoman: aside: Why receive an empty string when this is optional? it shold be nil instead. if it's really not intended to be optional, then it should not be, but looks like it should

Backpack-SwiftUI/FlightLeg/Classes/BPKFlightLeg.swift
```diff
@@ -0,0 +1,177 @@
...
+
+    // String-based init
+    public init(
+        departureArrivalTime: String,
+        nextDayArrival: String? = "",
+        flightDescription: String,
+        stopsInfo: String,
+        highlightStopsInfo: Bool,
+        duration: String,
+        operatedBy: String? = "",
```
frugoman, line 25: same as above, default value for these should be nil instead

Backpack-SwiftUI/FlightLeg/Classes/BPKFlightLeg.swift
```diff
@@ -0,0 +1,177 @@
...
+                stopsInfo: "Direct",
+                highlightStopsInfo: false,
+                duration: "2h 00m",
+                carrierLogo: UIImage(systemName: "airplane")
+            )
+            
+            BPKFlightLeg(
+                departureArrivalTime: "19:51 - 22:45",
+                nextDayArrival: "+1",
+                flightDescription: "LHR - SIN, SwissAir", // highlightedDescription,
```
frugoman, line 165: what makes this highlighted?

frugoman:
Closing due to inactivity, we can always reopen.

PR #2054, opened 2024-08-27
AI-638: FieldSet Component

AlaaAmrAmin:
Dispatch state updates using `.environment` instead of calling the `.updateState() `. This gives consumers the flexibility to add view modifiers to the wrapped view.

+ [ ] `README.md`
+ [ ] Tests


Backpack-SwiftUI/Select/Classes/BPKSelect.swift
```diff
@@ -51,11 +51,12 @@ extension View {
 }
 
 public struct BPKSelect: View {
+    @Environment(\.bpkFieldSetState) var wrapperState
```
frugoman, line 54:
```suggestion
    @Environment(\.bpkFieldSetState) var fieldSetState
```

frugoman (approved):
Looks good to me! it's a bit weird that we need an 'internal' starte property to hold the actual state of each component, but maybe it's just the name that's puzzling me.

PR #2066, opened 2024-09-27
DON-788: Update BPKSearchInputSummary

yurareutskiy:

+ [ ] `README.md`
+ [ ] Tests


Backpack-SwiftUI/AppSearchModal/Classes/BPKAppSearchModal.swift
```diff
@@ -76,6 +75,19 @@ public struct BPKAppSearchModal: View {
         .background(.surfaceDefaultColor)
     }
     
+    private var searchInputSummary: some View {
+        BPKSearchInputSummary(
+            placeholder: inputHint,
+            clearAction: .init(accessibilityLabel: clearAction?.accessibilityLabel ?? "", action: {
```
frugoman, line 81: it doesn't seem like this should be an optional property? if not, then the `clearAction` property should be non optional

Backpack-SwiftUI/AppSearchModal/Classes/BPKAppSearchModal.swift
```diff
@@ -76,6 +75,19 @@ public struct BPKAppSearchModal: View {
...
+                inputText = ""
```
frugoman, line 82: I wonder if this whole action should be handled by the client, rather than from the component. As in, it should be the client the one that clears out this text field, not the BPK component.
Either this or just not expose a closure for an action, if its going to be the component that clears the text (I'd not recommend this).

yurareutskiy: @frugoman thanks for your comment. In case of Date Selector we need to handle clear action on the client side, so we could perform some logic around date selections. However, we can transform it into "clearActionHandler", optional function, which would be called after clearing text out. And the clearing text would be left in the component implementation inside.
Otherwise, we need to keep it as it is.

frugoman: Right, I think then that this should be always in the client. even letting the client clear the text themselves. this releases this component of all logic around that and gives further flexibilty for the future (say for example, if we wanted to make the clear button be something else.)
another point for leaving this to the client is that now this component no longer has to 'guess' what to do with the inputState and can always be controlled by the client.

frugoman: That said, i'm not sure why you also removed the text field state. that mechanism is the one that controls the state of the field and restricted it to either be default or show the clear button.

Backpack-SwiftUI/SearchInputSummary/Classes/BPKSearchInputSummary.swift
```diff
@@ -35,22 +36,33 @@ public struct BPKSearchInputSummary: View {
     @Binding private var text: String
     @FocusState private var focused: Bool
     private let placeholder: String
-    private let inputPrefix: InputPrefix
+    private let inputPrefix: InputPrefix?
```
frugoman, line 39: wasnt this a requirement that the search input summary component will always have a prefix?

yurareutskiy: Yes, it was before
Now it should be optional :)

Backpack-SwiftUI/AppSearchModal/Classes/BPKAppSearchModal.swift
yurareutskiy: Basically, what happened is taking the clear action configuration out into the initialiser (it would be aligned with Android approach). And remove all unnecessary states (valid, error, disable). And we added a highlight state, so we could make borders highlighted even if the field is not in focus. This one is crucial for Date Selector latest design. Saying that, we can't have clear and highlight states at the same time.

Backpack-SwiftUI/SearchInputSummary/Classes/BPKSearchInputSummary.swift
frugoman: ahh okay, i just found the figma file. alright!

Backpack-SwiftUI/SearchInputSummary/Classes/BPKSearchInputSummary.swift
```diff
@@ -35,22 +36,33 @@ public struct BPKSearchInputSummary: View {
...
     private var state: State = .default
     
-    /// Creates a `BPKSearchInputSummary`.
-    ///
-    /// - Parameters:
-    ///   - placeholder: The placeholder text to display when the text field is empty.
-    ///   - inputPrefix: The prefix which would be displayed on the left of text input
-    ///   - text: The text to display in the text field.
+    private let readOnly: Bool
+    private let clearAction: ClearAction?
```
frugoman, line 43: per figma, this should not be optional

Backpack-SwiftUI/SearchInputSummary/Classes/BPKSearchInputSummary.swift
```diff
@@ -60,41 +72,38 @@ public struct BPKSearchInputSummary: View {
...
-                }
-                .accessibilityElement(children: .ignore)
-                .accessibilityLabel(accessibilityLabel)
-                .accessibilityAddTraits(.isButton)
-                .opacity(text.isEmpty ? 0.0 : 1.0)
-            } else {
-                BPKIconView(icon.icon)
-                    .foregroundColor(icon.color)
-                    .accessibilityHidden(true)
+        if let clearAction {
```
frugoman, line 97: per figma, I think this should be controlled not by checking if theres a clear aciton or not (as it should not be optional) but instead it should be by checking if the input text is not nil or empty

Backpack-SwiftUI/SearchInputSummary/Classes/BPKSearchInputSummary.swift
```diff
@@ -60,41 +72,38 @@ public struct BPKSearchInputSummary: View {
...
+            let clearIcon = BPKSearchInputSummary.Icon(icon: .closeCircle, color: .textSecondaryColor)
+            Button(action: clearAction.action) {
+                BPKIconView(clearIcon.icon)
```
frugoman, line 100: I think we can just do the definition of this here, and remove the `BPKSearchInputSummary.Icon` struct altogether, it makes no longer sense

```swift
BPKIconView(.closeCircle)
    .foregourndColor(.textSecondaryColor)

File: Backpack-SwiftUI/SearchInputSummary/Classes/BPKSearchInputSummary.swift [hunk 7]
```diff
@@ -35,22 +36,33 @@ public struct BPKSearchInputSummary: View {
...
+            self.action = action
+        }
+    }
+    
     public init(
         placeholder: String = "",
-        inputPrefix: InputPrefix = .icon(.search),
+        inputPrefix: InputPrefix? = nil,
+        clearAction: ClearAction? = nil,
+        readOnly: Bool = false,
```
frugoman, line 59: whats the expected behaviour if i set this to `readOnly = true` but also provide a clear action and I handle the click?
I'm thinking if this sould be handled by an enum, where you could either say `readonly` or `default or something` and that case receiving a clear action (again, non optional).
in the readOnly option, this component could show the icon but just not be clickable. this would make the API impossible to represent 'invalid' state.
if it even makes sense to show the (X) button when the field is readOnly. I couldn't find any design for the read only option on Figma. we should define this behaviour in figma first I thnk

Backpack-SwiftUI/SearchInputSummary/Classes/BPKSearchInputSummary.swift
```diff
@@ -113,6 +122,8 @@ public struct BPKSearchInputSummary: View {
         case .icon(let icon):
             BPKIconView(icon)
                 .foregroundColor(.textPrimaryColor)
+        case .none:
+            EmptyView()
```
frugoman, line 126: lets write this prefixView in a different way to avoid having to use EmptyView:
```swift
@ViewBuilder
private var prefixView: some View {
  if let inputPrefix {
    switch inputPrefix {
      case .text(let prefixText):
          BPKText(prefixText, style: .bodyDefault)
              .foregroundColor(.textSecondaryColor)
      case .icon(let icon):
          BPKIconView(icon)
    }
  }
}

File: Backpack-SwiftUI/SearchInputSummary/Classes/BPKSearchInputSummaryState.swift [hunk 9]
```diff
@@ -23,44 +23,7 @@ extension BPKSearchInputSummary {
...
-        var icon: BPKSearchInputSummary.Icon? {
-            switch self {
-            case .clear: return BPKSearchInputSummary.Icon(icon: .closeCircle, color: .textSecondaryColor)
-            case .valid: return BPKSearchInputSummary.Icon(icon: .tickCircle, color: .statusSuccessSpotColor)
-            case .error: return BPKSearchInputSummary.Icon(icon: .exclamationCircle, color: .statusDangerSpotColor)
-            default: return nil
-            }
-        }
+        /// Highlight borders
+        case manuallyHighlighted
```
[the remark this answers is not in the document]
frugoman: side note: this enum no longer represents state, but purely style, could you rename it to `BPKSearchInputSummaryStyle`?

Backpack-SwiftUI/AppSearchModal/Classes/BPKAppSearchModal.swift
frugoman: sorry, did you mean that we should not have highlight and clear at teh same time, or that we should and we cannot right now?

Backpack-SwiftUI/AppSearchModal/Classes/BPKAppSearchModalState.swift
```diff
@@ -23,16 +23,5 @@ extension BPKAppSearchModal {
     public enum TextFieldState {
```
frugoman, line 23: lets remove this entire enum, it no longer makes sense

Backpack-SwiftUI/AppSearchModal/Classes/BPKAppSearchModal.swift
```diff
@@ -22,20 +22,22 @@ public struct BPKAppSearchModal: View {
     let title: String
     @Binding var inputText: String
     let inputPrefix: BPKSearchInputSummary.InputPrefix
+    let clearAction: BPKSearchInputSummary.ClearAction?
     let inputHint: String
     let results: BPKAppSearchModalResults
     let closeAccessibilityLabel: String
     let onClose: () -> Void
-    private var textFieldState: TextFieldState = .default
+    @State private var textFieldState: TextFieldState = .default
```
frugoman: we should remove this property, its no longer used

Backpack-SwiftUI/AppSearchModal/Classes/BPKAppSearchModal.swift
frugoman: btw, i think youre not setting the state/style of the search input field here, to handle the manual setting of the highlughted (focused) state

yurareutskiy: We should and we cannot right now
In BPKAppSearchModal we don't need manual highlighting state

Backpack-SwiftUI/SearchInputSummary/Classes/BPKSearchInputSummary.swift
yurareutskiy: Before it was optional, I need to double check it with Adam

Backpack-SwiftUI/SearchInputSummary/Classes/BPKSearchInputSummaryState.swift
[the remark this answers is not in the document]
yurareutskiy: I don't want to have a confusion with @Focused property wrapper, which is about focus state
However, I don't like the naming as well, also I want to make it clear as much as possible
So, I'm thinking about highlightedBorder name

About BPKSearchInputSummaryStyle, yes, I agree.

Backpack-SwiftUI/SearchInputSummary/Classes/BPKSearchInputSummary.swift
yurareutskiy: If it's not optional anymore, then yes

Backpack-SwiftUI/SearchInputSummary/Classes/BPKSearchInputSummary.swift
```diff
@@ -60,41 +72,38 @@ public struct BPKSearchInputSummary: View {
                 .accessibilityHidden(true)
             TextField(placeholder, text: $text, prompt: placeholderView)
                 .font(style: .bodyDefault)
-                .foregroundColor(state.textColor)
-                .disabled(state.isDisabled)
+                .foregroundColor(.textPrimaryColor)
+                .disabled(readOnly)
```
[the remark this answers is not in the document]
yurareutskiy: Yes, I don't think either "Not enabled" should be there
It just didn't get to that draft version

Backpack-SwiftUI/SearchInputSummary/Classes/BPKSearchInputSummary.swift
yurareutskiy: Basically, that's a figma for that: https://www.figma.com/design/RkSuI77vgPXkTf2SvuEvKf/Apps---Search-controls-GC?node-id=1867-58841&t=xeqerj6azCGFcs1a-4

Read only should affect only user text input ability. But we still can set the text value outside. And user still may remove the text by clicking the clear button. So, they're independent

Backpack-SwiftUI/SearchInputSummary/Classes/BPKSearchInputSummary.swift
yurareutskiy: Good catch! thanks

Backpack-SwiftUI/AppSearchModal/Classes/BPKAppSearchModalState.swift
yurareutskiy: Yes, I agree, I'm going to

Backpack-SwiftUI/AppSearchModal/Classes/BPKAppSearchModal.swift
yurareutskiy: It's removed yes

Backpack-SwiftUI/SearchInputSummary/Classes/BPKSearchInputSummary.swift
yurareutskiy: yes, we can make it non-optional

Backpack-SwiftUI/SearchInputSummary/Classes/BPKSearchInputSummary.swift
yurareutskiy: Done

Backpack-SwiftUI/SearchInputSummary/Classes/BPKSearchInputSummary.swift
yurareutskiy: We still need this struct for a prefix, but yes, here we can simplify it and not use this struct for a clear action

frugoman (approved):
Looks good! thanks for addressing the comments!

Backpack-SwiftUI/SearchInputSummary/Classes/BPKSearchInputSummaryState.swift
```diff
@@ -20,47 +20,10 @@ import SwiftUI
 
 extension BPKSearchInputSummary {
     /// The state of the text field.
```
novinfard, line 22:
```suggestion
    /// The style of the text field.
```

Backpack-SwiftUI/SearchInputSummary/Classes/BPKSearchInputSummary.swift
```diff
@@ -35,22 +36,33 @@ public struct BPKSearchInputSummary: View {
     @Binding private var text: String
     @FocusState private var focused: Bool
     private let placeholder: String
-    private let inputPrefix: InputPrefix
-    private var state: State = .default
+    private let inputPrefix: InputPrefix?
+    private var state: Style = .default
```
novinfard, line 40:
```suggestion
    private var style: Style = .default
```
Not an issue, just to be consistent with new naming/approach.

novinfard (approved):
Great work 🚀
Niko already mentioned a few comments that you addressed it.

Additionally, for the integration, the focus state and UI tests function as required. ✅

PR #2090, opened 2024-10-25
Fix: BPKChip goes beyond its parent when the text is too long

iHandle:
**Changes**

* Use `.lineLimit(1)` instead of `.fixedSize(horizontal: true, vertical: false)` in `ChipButtonStyle` so that `BPKChip` will not go beyond its parent due to too long text.
* Fix bug on `BPKFlowStackView` when computing Rows

**Preview**
| Before | After |
| ------- | ----- |
|  |  |

+ [ ] `README.md`
+ [ ] Tests

### DIFF
--- a/Backpack-SwiftUI/FlowStackView/Classes/BPKFlowStackView.swift
+++ b/Backpack-SwiftUI/FlowStackView/Classes/BPKFlowStackView.swift
@@ -92,8 +92,10 @@ public struct BPKFlowStackView<Data: Collection, Content: View>: View where Data
             let doesNotFitInCurrentRow = elementSize.width + spacing.width > remainingWidth
             
             if doesNotFitInCurrentRow { // new row
-                remainingWidth = availableWidth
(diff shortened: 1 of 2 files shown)


Backpack-SwiftUI/FlowStackView/Classes/BPKFlowStackView.swift
```diff
@@ -92,8 +92,10 @@ public struct BPKFlowStackView<Data: Collection, Content: View>: View where Data
             let doesNotFitInCurrentRow = elementSize.width + spacing.width > remainingWidth
             
             if doesNotFitInCurrentRow { // new row
-                remainingWidth = availableWidth
-                currentRow += 1
+                if rows.indices.contains(currentRow) {
```
iHandle, line 95: If the current row does not have any elements yet, `currentRow` does not need to be + 1

alejandrorosas (approved):
LGTM

PR #2104, opened 2024-11-08
[DON-715] Add capabilities to `BPKcalendar` required for wholeMonth selection

novinfard:
## TL;DR;
This pull request purpose is to add the changes required for `BPKcalendar` to make wholeMonth selection using existing `CalendarMonthAccessoryAction`, while maintaining its flexibility for use-cases.

## Demo

https://github.com/user-attachments/assets/cdf93679-fe8f-4a74-a2db-6310fd5a033d

## Detail
On Backpack Android, we have whole month selection as part of the Calendar. On iOS we aimed to have more flexible approach using `CalendarMonthAccessoryAction` implemented for this reason. 

However, to keep this action generic and flexible while making whole month selection feasible we placed the following changes:

### (1) Flexible action based on given month
Not all months have the same action. For instance, for the last month (in terms of whole month use-case) we do not have any action. We used the following closure to give this flexibility:
`((Date) -> CalendarMonthAccessoryAction?)`
Where `Date` is the first day of month. So the consumer can decide which action they want to set for each month.

### (2) Adding a new state for Range Selection
The handling of the date selection occurs inside the `BPKCalendar`. We have a new state for Range selection, that can be set by the consumer. The reasons for having `wholeMonth` state are the following:
- The appearance of start and end date is different from range selection
- The business logic should be implemented for `wholeMonth` is different from `range`, and the calendar is the source of truth for the user selection.

| Range    | Whole Month |
| --------- | ------- |
|  Screenshot 2024-11-15 at 09 15 36    |    Screenshot 2024-11-15 at 09 15 25     |


Backpack-SwiftUI/Calendar/Classes/Range/Cells/RangeSelectionCalendarDayCell.swift
```diff
@@ -22,14 +22,22 @@ struct RangeSelectionCalendarDayCell: View {
     let date: Date
     let selection: ClosedRange<Date>
     let calendar: Calendar
-    
+    let highlightRangeEnds: Bool
+
+    init(date: Date, selection: ClosedRange<Date>, calendar: Calendar, highlightRangeEnds: Bool = true) {
```
frugoman, line 27: same as above! I'd be explicit everywhere!

Backpack-SwiftUI/Calendar/Classes/CalendarMonthAccessoryAction.swift
```diff
@@ -22,13 +22,14 @@ import SwiftUI
 ///     be performed when a user interacts with an accessory in the calendar.
 public struct CalendarMonthAccessoryAction {
```
frugoman, line 23: I think we may need to introduce an enum here, to encapsulate the behaviour. Using this action for for range selection is stretching this api to a point where its not clear, or not as clear as before. I suggest introducing something like so:
```swift
public struct CalendarMonthAccessoryAction {
  let title: String
  let action: Action
  
  public enum Action {
     case custom((Date) -> Void)
     case wholeMonthSelection(ClosedRange<Date> -> Void)
  }
}
```
or something like that, so that when clients use this, its more clear how it will behave and its more explicit

PR #2113, opened 2024-11-22
[DON-715] Whole month in single date selection

novinfard:
This PR adds the whole month selection as one the single date selection mode states.

### Demo
https://github.com/user-attachments/assets/04219a04-5a43-4481-99db-1886011849f0

### DIFF
--- a/Backpack-SwiftUI/Calendar/Classes/CalendarSignleSelectionState.swift
+++ b/Backpack-SwiftUI/Calendar/Classes/CalendarSignleSelectionState.swift
@@ -16,18 +16,13 @@
  * limitations under the License.
  */
 
-struct SingleDayAccessibilityProvider {
-    let accessibilityConfigurations: SingleAccessibilityConfigurations
-    let dateFormatter: DateFormatter
-    
-    func accessibilityLabel(for dayDate: Date) -> String {
-        dateFormatter.string(from: dayDate)
-    }
+import Foundation
 
-    func accessibilityHint(for dayDate: Date, selection: CalendarSingleSelectionState?) -> String {
-        if selection?.isSelected(dayDate) == true {
-            return ""
-        }
-        return accessibilityConfigurations.selectionHint
-    }
+/// The `CalendarSingleSelectionState` enum represents the different states of a single selection.
+public enum CalendarSingleSelectionState {
+    /// The state, with single date selection
+    case single(Date?)
+
+    /// The state the user has selected the whole month.
+    case wholeMonth(ClosedRange<Date>)
 }
--- a/Backpack-SwiftUI/Calendar/Classes/CalendarSignleSelectionState.swift
+++ b/Backpack-SwiftUI/Calendar/Classes/CalendarSignleSelectionState.swift
@@ -24,5 +24,5 @@ public enum CalendarSingleSelectionState {
     case single(Date?)
 
     /// The state the user has selected the whole month.
-    case wholeMonth(ClosedRange<Date>)
+    case wholeMonth(ClosedRange<Date>, accessibilityConfig: WholeMonthAccessibilityConfigurations)
 }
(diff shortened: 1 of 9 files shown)


Backpack-SwiftUI/Calendar/Classes/CalendarSignleSelectionState.swift
```diff
@@ -0,0 +1,28 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Foundation
+
+/// The `CalendarSingleSelectionState` enum represents the different states of a single selection.
+public enum CalendarSingleSelectionState {
+    /// The state, with single date selection
+    case single(Date?)
```
frugoman, line 24: I think this should probably not be optional right? the optional would be on the enum itself, but not the date.
reasoning being:
if there's a single date selected, it makes no sense that that seleciton has no date

Backpack-SwiftUI/Calendar/Classes/Single/SingleCalendarContainer.swift
```diff
@@ -30,20 +30,36 @@ struct SingleCalendarContainer<MonthHeader: View, DayAccessoryView: View>: View
...
         CalendarSelectableCell {
-            if selection == dayDate {
-                SingleSelectedCell(calendar: calendar, date: dayDate)
-            } else {
+            switch selection {
+            case .single(let date):
+                if date == dayDate {
+                    SingleSelectedCell(calendar: calendar, date: dayDate)
+                } else {
+                    DefaultCalendarDayCell(calendar: calendar, date: dayDate)
```
frugoman, line 38: if you appliy the above comment about optionality, I think this if let would be done on the selection itself, rather than the date.
```swift
switch selection {
  case single(let date): SingleSelectedCel(...)
  case wholeMonth(..): ....
  case nil: DefaultCalendarDayCell(...)
}

File: Backpack-SwiftUI/Calendar/Classes/CalendarSignleSelectionState.swift (hunk 1 above)
Comment (novinfard) [reply Backpack-SwiftUI/Calendar/Classes/CalendarSignleSelectionState.swift:24]: In both Calendar selection types (range/single) there's a use-case for not having a date being selected (when dates being cleared). So this optional date is still valid.

|      Single      |     Range    |
|---------------------|------------------|
|       Simulator Screenshot - iPhone SE (3rd generation) - 2024-11-26 at 10 38 16         |     Simulator Screenshot - iPhone SE (3rd generation) - 2024-11-26 at 10 38 10          |

File: Example/Backpack/SwiftUI/Components/Calendar/CalendarExampleSingleView.swift [hunk 3]
```diff
@@ -49,28 +49,39 @@ struct CalendarExampleSingleView: View {
...
+                        selectionHint: "Double tap to select date",
+                        wholeMonth: nil
                     )
                 ),
                 calendar: calendar,
                 validRange: validRange,
                 initialMonthScroll: monthScroll
             )
+            .monthAccessoryAction { _ in
+                return CalendarMonthAccessoryAction(
```
frugoman, line 78: I wonder if the accessibility setting for this should be here instead, to avoid being possible to send nil to that, when you do have an action. wdyt?

novinfard: Good point. 👍

I made this config as part whole month selection state. So if the consumer doesn't want to use the wholeMonth they don't have to provide this value. Removed this optional. ✅

Backpack-SwiftUI/Calendar/Classes/CalendarSignleSelectionState.swift
novinfard: Cool. Right. I applied the change. ✅

Backpack-SwiftUI/Calendar/Classes/Single/SingleCalendarContainer.swift
novinfard: I removed the optional, but we still need this statement to determine which date needs to be highlighted among other dates when a date has been selected `date == dayDate`. 

When the tap action happens, we change the state, and the change of state helps us to render the changes.

frugoman (approved):
Looks good!!

PR #2115, opened 2024-11-25
[DRAFT][DON-715] whole month with separate calendar type

novinfard:
[DON-715] 
whole month with separate calendar type


novinfard:
The PR is solely to demonstrate the approach 
- there is a scroll jumping issue with this way of implementing the whole month (because of having two different calendar container)

PR #2129, opened 2025-01-03
Make BPKCalendar accessibilty non mandatory

yurareutskiy:

+ [ ] `README.md`
+ [ ] Tests

### DIFF
--- a/Backpack-SwiftUI/Calendar/Classes/BPKCalendar.swift
+++ b/Backpack-SwiftUI/Calendar/Classes/BPKCalendar.swift
@@ -30,6 +30,7 @@ import SwiftUI
 ///   - validRange: The range of dates that the calendar should allow the user to select.
 ///     This is specified as a`ClosedRange<Date>`.
 ///   - initialMonthScroll: The initial scrolling to the month using `MonthScroll`
+///   - calendarAccessibilityConfiguration: The optional configuration to add accessibility strings to the calendar
 ///
 /// The `BPKCalendar` view also allows you to specify an accessory action. This is a closure that takes a string and
 ///     a date, and is called when the user interacts with an accessory in the calendar.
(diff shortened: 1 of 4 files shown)


Backpack-SwiftUI/Calendar/Classes/BPKCalendar.swift
```diff
@@ -50,7 +51,7 @@ public struct BPKCalendar<DayAccessoryView: View>: View {
         calendar: Calendar,
         validRange: ClosedRange<Date>,
         initialMonthScroll: MonthScroll? = nil,
-        calendarAccessibilityConfiguration: CalendarAccessibilityConfiguration,
+        calendarAccessibilityConfiguration: CalendarAccessibilityConfiguration = .init(),
```
frugoman, line 54: It's not that it shouldn't be mandatory, it should.
The thing is that the API should not allow for the component to be in an invalid state. ie, this API allows for me to not set accessibiltiy, where that should never be the case.
What the api should allow for, is so that the correct accessibility configurations are asked for the right selection mode.
example, i'm configuring a calendar with single selection only, it should not ask for a reanged selection a11y config, as its unnecesary. While also forcing the client to pass a11y configuration for single selection, as we take a11y as a requirement. it should not be possible to instantiate a calendar without setting it.

Backpack-SwiftUI/Calendar/Classes/Core/CalendarTypeContainerFactory.swift
```diff
@@ -19,9 +19,12 @@
 import SwiftUI
 
 public struct CalendarAccessibilityConfiguration {
-    public let singleSelection: SingleDayAccessibilityProvider
-    public let rangeSelection: RangeDayAccessibilityProvider
-    public init(singleSelection: SingleDayAccessibilityProvider, rangeSelection: RangeDayAccessibilityProvider) {
+    public let singleSelection: SingleDayAccessibilityProvider?
+    public let rangeSelection: RangeDayAccessibilityProvider?
```
frugoman, line 23: what'd be the reason for instantiating a configuiration with both values nil? should this be an enum instead?

novinfard:
Closing the PR as another approach being merged to address it:
https://github.com/Skyscanner/backpack-ios/pull/2130

PR #2130, opened 2025-01-03
Refactor calendar to support dynamic selection type changes

frugoman:
Refactoring implementation to revert back to previous approach for composition of views for the BPKCalendar on SwiftUI
this refactor includes changes to support proper dynamic changes over the selection mode of the Calendar, which avoids re-renders of the entire calendar when selection mode is changed dynamically.

Also, the change includes a minor refactor that extracts the code for handling range selections, preparing for a tufure change that will require different handling of this depending on a flag's value

Note: Purposefully marking this as `Minor`, as this is reverting to the previous API, which was changed as a Minor change mistakenly. This aims to not break clients that bump minor versions expecting no api changes

https://github.com/user-attachments/assets/99cc070f-c75d-4f83-b406-3784f2ede7ef


Backpack-SwiftUI/Calendar/Classes/BPKCalendar.swift
```diff
@@ -50,15 +49,13 @@ public struct BPKCalendar<DayAccessoryView: View>: View {
         calendar: Calendar,
         validRange: ClosedRange<Date>,
         initialMonthScroll: MonthScroll? = nil,
-        calendarAccessibilityConfiguration: CalendarAccessibilityConfiguration,
```
frugoman, line 53: This is now handled as it was before, as part of the binding itself, to not allow the API to lead to invalid states

Backpack-SwiftUI/Calendar/Classes/CalendarDayCells/CalendarSelectableCell.swift
```diff
@@ -18,111 +18,11 @@
 
 import SwiftUI
 
-struct CalendarSelectableCell: View {
-    let selectionType: CalendarSelectionType
-    let calendar: Calendar
-    let accessibilityProvider: CalendarAccessibilityConfiguration
-    let dayDate: Date
-    let onSelection: (Date) -> Void
+struct CalendarSelectableCell<Cell: View>: View {
```
frugoman, line 21: nothing to see here, just reverting back to the old implementation, which supports better composition of views among other things

Backpack-SwiftUI/Calendar/Classes/Core/CalendarMonthGrid.swift
```diff
@@ -30,8 +30,8 @@ struct CalendarMonthGrid<
 
     @State private var dayCellHeight: CGFloat = 0
     @ViewBuilder let dayCell: (Date) -> DayCell
-    @ViewBuilder let emptyLeadingDayCell: (EmptyCellInfo) -> EmptyLeadingDayCell
```
frugoman, line 33: Empty cell info handled internally, as it's not something of interes outside this particualr view.

Backpack-SwiftUI/Calendar/Classes/Core/CalendarMonthGrid.swift
```diff
@@ -40,56 +40,74 @@ struct CalendarMonthGrid<
         let firstWeekday = calendar.firstWeekday // Locale-aware first day of the week
         let weekdayOfMonthStart = calendar.component(.weekday, from: monthDate)
         // Calculate the offset based on the first weekday
-        let weekdaysOffset = (weekdayOfMonthStart - firstWeekday + daysInAWeek) % daysInAWeek
```
frugoman, line 43: reverting these lines to where they were before

Backpack-SwiftUI/Calendar/Classes/Core/CalendarMonthGrid.swift
```diff
@@ -40,56 +40,74 @@ struct CalendarMonthGrid<
...
-                    }
+            let totalCellsUsed = numberOfDaysInMonth + daysFromPreviousMonth
+            let remainingCells = daysInAWeek - (totalCellsUsed % daysInAWeek)
+        
+            remainingEmptyCells(remainingCells: remainingCells)
+        }
+    }
+    
+    @ViewBuilder
+    private func previousEmptyCells(daysFromPreviousMonth: Int) -> some View {
```
frugoman, line 66: Extracted this from the above method to make it easier to read, also introducing the usage of a small internal struct DayCellIdentifiable, as the preiovus range only approach weas leading to mis-reusage of some cells, as they had repeating ids (which was a bug).
Now we can set specific IDs to each cell and make sure they do not repeat across weeks/months in the calendar, fixing the bug where sometimes the cells were not rendering correctly

Backpack-SwiftUI/Calendar/Classes/Core/CalendarMonthGrid.swift
```diff
@@ -40,56 +40,74 @@ struct CalendarMonthGrid<
...
     
+    private struct DayCellIdentifiable: Identifiable {
+        let id: String
+        let index: Int
+    }
+    
     @ViewBuilder
     private func currentMonthDayCell(numberOfDaysInMonth: Int) -> some View {
-        ForEach(0..<numberOfDaysInMonth, id: \.self) { cellIndex in
+        let days = Array(0..<numberOfDaysInMonth)
```
frugoman, line 104: same as above, using the struct rather than a ForEach with a range, that led to bugs before

Backpack-SwiftUI/Calendar/Classes/Core/CalendarTypeContainerFactory.swift
```diff
@@ -18,23 +18,13 @@
 
 import SwiftUI
 
-public struct CalendarAccessibilityConfiguration {
```
frugoman, line 21: reverted this to the way it was handled before, reducing duplication and potential invalid states

Backpack-SwiftUI/Calendar/Classes/Core/CalendarTypeContainerFactory.swift
```diff
@@ -49,97 +39,64 @@ struct CalendarTypeContainerFactory<MonthHeader: View, DayAccessoryView: View>:
...
-                accessibilityProvider: calendarAccessibilityConfiguration,
-                dayDate: dayDate,
-                onSelection: handleSelection
-            )
-        }
-        
-    }
-    
-    @ViewBuilder func emptyLeadingDayCell(for emptyDayInfo: EmptyCellInfo) -> some View {
-        switch selectionType {
```
frugoman, line 79: most this changes are reverting back to the old implementation. with an exception mentioned below

Backpack-SwiftUI/Calendar/Classes/Core/CalendarTypeContainerFactory.swift
```diff
@@ -49,97 +39,64 @@ struct CalendarTypeContainerFactory<MonthHeader: View, DayAccessoryView: View>:
...
-                let lastDayOfPreviousMonth = calendar.date(byAdding: .init(day: -1), to: emptyDayInfo.month),
-                let firstDayOfCurrentMonth = calendar.date(byAdding: .init(day: 1), to: lastDayOfPreviousMonth),
-                selectionRange.contains(lastDayOfPreviousMonth),
-                selectionRange.contains(firstDayOfCurrentMonth)
-            {
-                Color(.surfaceSubtleColor)
-            } else {
-                // otherwise we occupy the space with a clear view
-                DefaultEmptyCalendarDayCell()
+            switch selectionType {
```
frugoman, line 42: the relevant change here (the point of this PR) is that the `CalendarContainer` is now here, wrapping both the single and range selection calendar month containers, instead of the `CalendarContainer` being inside the single and range respectively. This is what makes the change of selection types dynamically not re-render the entire calendar, as the only thing that's now re-rendered is the months themselves, and not the entire scrollview.

For context, the problem was that before when you changed selection types of the calendar, it'd re-render the whole thing as this switch was the entire calendar including the scrollviews, now the only thing that is re rendered is the months's grids, WITHIN the container scrollview, making the changes to this vlaue NOT trigger a scroll 'reset' 👍

frugoman: So essentially: 
now: 
```
Container {
  Range { Grid } or Single { Grid }
}
```
before:
```
Range {
  Container { Grid }
}
Single {
  Container { Grid }
}
```

Backpack-SwiftUI/Calendar/Classes/Range/RangeCalendarContainer.swift
```diff
@@ -0,0 +1,237 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import SwiftUI
+
+protocol RangeCalendarSelectionHandler {
```
frugoman, line 21: Introducing this protocol so that we can extract the selection handling logic from the Container view and eventually swap it with a new implementation, which is a change that's coming for the range selection calendars, which will be experimented upon.

When the new seleciton logic is implemented, we should only add a new class that conforms to this protocol and, if flag is enabled use one or the other implementaion.

Backpack-SwiftUI/Calendar/Classes/Core/CalendarTypeContainerFactory.swift
```diff
@@ -49,97 +39,64 @@ struct CalendarTypeContainerFactory<MonthHeader: View, DayAccessoryView: View>:
...
+        RangeCalendarMonthContainer(
+            selectionState: selection,
+            calendar: calendar,
+            validRange: validRange,
+            accessibilityProvider: RangeDayAccessibilityProvider(
+                accessibilityConfigurations: accessibilityConfigurations,
+                dateFormatter: accessibilityDateFormatter
+            ),
+            month: month,
+            selectionHandler: DefaultRangeCalendarSelectionHandler(
```
frugoman, line 94: When new one is implemented later on, here we'd have something like:
```swift
selectionHandler: flagEnabled ? NewHandler() : DefaultHandler()

File: Backpack-SwiftUI/Calendar/Classes/Single/SingleCalendarContainer.swift [hunk 12]
```diff
@@ -0,0 +1,108 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import SwiftUI
+
+struct SingleCalendarMonthContainer<MonthHeader: View, DayAccessoryView: View>: View {
```
frugoman, line 21: nothng to see here, just reverting

Example/Backpack/SwiftUI/Components/Calendar/CalendarExampleDynamicView.swift
```diff
@@ -0,0 +1,104 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import SwiftUI
+import Backpack_SwiftUI
+
+struct CalendarExampleDynamicView: View {
```
frugoman, line 23: created this example to play around the idea of dynamically changing the selection mode of the calendar, which is now better supported wityh this pr

brunomdac (approved):

PR #2138, opened 2025-01-16
DON-1029 Segmented control

faurevid:
Created new segmented control component
Added documentation, tests and screenshots

Jira ticket: [DON-1029](https://skyscanner.atlassian.net/browse/DON-1029)
Figma design: https://www.figma.com/design/irZ3YBx8vOm16ICkAr7mB3/Backpack-Components?m=auto&node-id=49086-3871&t=pQApQKbd31dMaLrK-1

| Day | Night |
| --- | --- |
| iPhone-swiftui_segmented-control___default_lm |iPhone-swiftui_segmented-control___default_dm|

+ [ ] `README.md`
+ [ ] Tests


Backpack-SwiftUI/SegmentedControl/Classes/BPKSegmentedControl.swift
```diff
@@ -0,0 +1,71 @@
...
+    ///   - selectedIndex: A binding to an Int that indicates which segment is selected
+    ///   - accessibilityLabel: A string used to identify the component and each segment
+    public init(items: [String], selectedIndex: Binding<Int>, accessibilityLabel: String? = nil) {
+        // Limit the number of items to 4
+        self.items = Array(items.prefix(4))
+        self.accessibilityLabel = accessibilityLabel
+        if let accessibilityLabel = accessibilityLabel {
+            self.itemsAccessibilityLabel = Array(repeating: "", count: items.count)
+            for index in items.indices {
+                self.itemsAccessibilityLabel[index] = "\(accessibilityLabel)_\(index)"
```
novinfard, line 45: We do not need `itemsAccessibilityLabel` as stored property. Instead we can use a helper when we set the elements. (line 60). 

Also in another note `for index in items.indices` has potential for out-of-range crash, and if we wanted to keep it we should have changed that to `for index in self.items.indices`

Backpack-SwiftUI/SegmentedControl/Classes/BPKSegmentedControl.swift
```diff
@@ -0,0 +1,71 @@
...
+    public var body: some View {
+        Picker("", selection: $selectedIndex) {
+            ForEach(0..<items.count, id: \.self) { index in
+                Text(self.items[index])
+                    .accessibilityIdentifier((accessibilityLabel != nil) ? itemsAccessibilityLabel[index] : "")
+            }
+        }
+        .pickerStyle(.segmented)
+        .padding(.horizontal, BPKSpacing.base)
+        .accessibilityIdentifier(accessibilityLabel ?? "")
```
novinfard, line 65: We should use `If` modifier (defined in ConditionalModifier.swift) to avoid setting empty accessibility identifiers.

novinfard:
Hey @faurevid 
Two additional point:
- It seems we have major restrictions on how we apply the style individually to iOS Segment control. (It can only be applied globally) We already have more constraints on this iOS native component and with the current situation I can see a custom component as a way around the problem. I made the draft version of that here:
https://github.com/Skyscanner/backpack-ios/compare/main...donburi/DON-1029-segmented-control-custom
We should align that with the product and current Backpack owner before implementation.

- Also this change can be labeled as `major` instead of `minor`, as we didn't have the component before.

PR #2141, opened 2025-01-21
DON-1029 segmented control custom

faurevid:
Created new segmented control component
Added documentation, tests and screenshots

Jira ticket: [DON-1029](https://skyscanner.atlassian.net/browse/DON-1029)
Figma design: https://www.figma.com/design/irZ3YBx8vOm16ICkAr7mB3/Backpack-Components?m=auto&node-id=49086-3871&t=pQApQKbd31dMaLrK-1

| Day | Night |
| --- | --- |
| iPhone-swiftui_segmented-control___default_lm |iPhone-swiftui_segmented-control___default_dm|

+ [ ] `README.md`
+ [ ] Tests


Backpack-SwiftUI/SegmentedControl/Classes/BPKSegmentedControl.swift
```diff
@@ -0,0 +1,131 @@
...
+        }
+    }
+
+    @ViewBuilder
+    private func segmentButton(for index: Int) -> some View {
+        HStack(spacing: 0) {
+            Button(
+                action: { withAnimation { selectedIndex = index } },
+                label: {
+                    Text(items[index])
```
novinfard, line 82: Based on the accessibility a11y snapshot (test_accessibility.a11y.png), this component supports dynamic type up to `extraExtraExtraLarge` which is a suitable support range. 

We just need to specify that in the UI component to make it work as expected. I think we need to set the support on this element (you can assert it)
`.dynamicTypeSize(...DynamicTypeSize.xxxLarge)`
https://developer.apple.com/documentation/swiftui/dynamictypesize

Could we please update it?

yurareutskiy: @novinfard are you sure we need to make it grow? Of course it depends on product requirements, but wouldn't be fixed by a layout of a parent view? And I would suggest if it doesn't fit, it should truncate it

Example/Backpack/SwiftUI/Components/SegmentedControl/SegmentedControlExampleView.swift
```diff
@@ -0,0 +1,55 @@
...
+ * limitations under the License.
+ */
+
+import SwiftUI
+import Backpack_SwiftUI
+
+struct SegmentedControlExampleView: View {
+    @State var selectedIndex = 0
+    private let accessibilityLabel = "bpk_segmented_control_example"
+    private let items: [String] = ["1", "2", "3", "4"]
```
yurareutskiy, line 25: I think it's worth to add different text options, not only with short strings like those, but also with longer content inside.
That is not really required, but it would be good to add

Backpack-SwiftUI/SegmentedControl/Classes/BPKSegmentedControl.swift
novinfard: Hey @yurareutskiy
The text needs to be readable by the user.
- Using minimumFontScale works too, but it is not helpful for dynamic font and accessibility.

novinfard (approved):
Looks good 🚀 Nice work
We just need to have one-final commit to verify snapshots

maxgirt (approved):
LGT RALP

PR #2142, opened 2025-01-23
Make bundle parameter optional in BPKIcon methods

frugoman:
Update BPKIcon to make the bundle parameter optional, set a default bundle if none is provided, and refactor the initializer for improved readability.

### DIFF
--- a/Backpack-SwiftUI/Icons/Classes/BPKIcon.swift
+++ b/Backpack-SwiftUI/Icons/Classes/BPKIcon.swift
@@ -16,13 +16,15 @@
  * limitations under the License.
  */
 
+import Backpack_Common
+
 public struct BPKIcon {
     public let name: String
     let bundle: Bundle?
 
     init (name: String, bundle: Bundle? = nil) {
         self.name = name
-        self.bundle = bundle
+        self.bundle = bundle ?? BPKCommonBundle.iconsBundle
     }
     
     public enum Size {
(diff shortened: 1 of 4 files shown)


templates/swiftui/BPKIcons.njk
```diff
@@ -24,12 +24,12 @@ public extension BPKIcon {
 }
 
 public extension BPKIcon {
-    static func named(_ iconName: String, bundle: Bundle) -> BPKIcon {
```
frugoman, line 27: this method was itnroduced as we before thought this was the problem. the actual api we want is the one from this PR here, one method with a default value.

Backpack-SwiftUI/Icons/Classes/BPKIcon.swift
```diff
@@ -15,15 +15,16 @@
...
+
 import Backpack_Common
 
 public struct BPKIcon {
     public let name: String
-    let bundle: Bundle
+    let bundle: Bundle?
 
-    init (name: String, bundle: Bundle = BPKCommonBundle.iconsBundle) {
+    init(name: String, bundle: Bundle? = nil) {
```
frugoman, line 25: not using BPKCommon in the initialiser definition is the fix of the problem, as now the compiler and linker is not aware that this type exists, only at runtime this is resolved, so this makes that users of this struct dont need to import Common explicitly

brunomdac (approved):

PR #2158, opened 2025-02-03
DON-1089: Fixing a scroll issue in BPKCalendar 

yurareutskiy:
[DON-1089](https://skyscanner.atlassian.net/browse/DON-1089)

When scrolling back upwards sometimes from a future month to a previous month the header jumps down and causes the overlapping issue on the right 
image

The solution was to calculate heights and keep them cached. Also make all cells in a LazyVGrid contained in one array, so just one ForEach would be used (apparently, there is an issue with using multiple ForEach)

More info:
https://skyscanner.atlassian.net/wiki/x/Ty9rTQ

+ [ ] `README.md`
+ [ ] Tests


Backpack-SwiftUI/Calendar/Classes/BPKCalendar.swift
```diff
@@ -96,6 +89,22 @@ public struct BPKCalendar<DayAccessoryView: View>: View {
...
+            dateFormatter: monthHeaderDateFormatter,
+            calendar: calendar,
+            validRange: validRange,
+            accessoryAction: accessoryAction,
+            currentlyShownMonth: $currentlyShownMonth,
+            parentProxy: calendarProxy
+        )
+        .frame(height: headerHeight)
+        .modifier(ReadSizeModifier {
+            headerHeight = max($0.height, headerHeight ?? 0)
```
novinfard, line 104: How about having the initial headerHeight as 0 instead of optional nil ?

Backpack-SwiftUI/Calendar/Classes/Core/CalendarContainer.swift
```diff
@@ -32,7 +33,7 @@ struct CalendarContainer<MonthContent: View>: View {
                 LazyVStack(spacing: BPKSpacing.none) {
                     ForEach(0...monthsToShow, id: \.self) { monthIndex in
                         let firstDayOfMonth = firstDayOf(monthIndex: monthIndex)
-                        monthContent(firstDayOfMonth)
+                        monthContent(firstDayOfMonth, $accessoryViewHeight)
```
novinfard, line 36: Do we still need `accessoryViewHeight` for` monthContent` ?

Backpack-SwiftUI/Calendar/Classes/Core/CalendarMonthGrid.swift
```diff
@@ -28,7 +40,8 @@ struct CalendarMonthGrid<
     let calendar: Calendar
     let validRange: ClosedRange<Date>
 
-    @State private var dayCellHeight: CGFloat = 0
+    @State private var dayCellHeight: CGFloat?
+    @Binding var accessoryViewHeight: CGFloat?
```
novinfard, line 44: How about having the initial `accessoryViewHeight` and `dayCellHeight` as 0 instead of optional nil ?

Backpack-SwiftUI/Calendar/Classes/Core/CalendarContainer.swift
yurareutskiy: so, the idea is to share between each month view the accessory view height, so it wouldn't need to calculate it each time
this should help with the consistency
the problem could be because the magnifier icon and the text have different heights (15.6 vs 16.0), so in my understanding, if one month has only one magnifiers, then when it happened to display a text, then it would need to recalculate the heights
and to avoid that issue I thought about this sharable height value, so every component already would aware of the target height

however (!), I've checked the latest changes with removing that sharable height and keeping it as a state (as it was before), and I haven't noticed anything wrong with that after some tests

what do you think?

Backpack-SwiftUI/Calendar/Classes/Core/CalendarMonthGrid.swift
yurareutskiy: we shouldn't set 0s here
it's important to set nil as an initial value, so swiftui could calculate it correctly and return the correct view when we call `.frame()` modifier
I tried to set to 0, it didn't work
that's an example

https://github.com/user-attachments/assets/3d6a270f-4060-4833-bee8-e3f848746668

Backpack-SwiftUI/Calendar/Classes/BPKCalendar.swift
yurareutskiy: Answered to that below

Backpack-SwiftUI/Calendar/Classes/Core/CalendarMonthGrid.swift
novinfard: That makes sense as we want to get the size behaviour from the view for the first time. 
https://developer.apple.com/documentation/swiftui/view/frame(width:height:alignment:)
Screenshot 2025-02-06 at 15 02 49

Thank you for checking.

Backpack-SwiftUI/Calendar/Classes/Core/CalendarContainer.swift
novinfard: I think we can keep it as it was before. It seems other fixes helped us to reduce/escape the flakiness on this height. I couldn't reproduce the issue again.

novinfard (approved):
Great work and investigation 🚀

mcf21 (approved):

Backpack-SwiftUI/Calendar/Classes/Core/CalendarContainer.swift
frugoman: I'm not sure caching a height is a good solution here, it should be left to SwiftUI to calculate on real time based on constraints, I think there is something else happening here, perhaps something we're doing wrong when reusing month cells or similar? setting a frame for things should almost never be the solution unless specifically due to a design reason

novinfard: @frugoman 
This change has been reverted ✅ It was just an open comment.

frugoman:
Fixed in https://github.com/Skyscanner/backpack-ios/pull/2167

PR #2167, opened 2025-02-10
Refactor calendar components to use CalendarGridCalculator for improved performance

frugoman:
Addresses: [DON-1089](https://skyscanner.atlassian.net/browse/DON-1089)

Enhance calendar layout handling and performance by integrating CalendarGridCalculator across various components, while adding unit tests to ensure correct grid calculation and caching behavior.

This is due to a problem that the calendar had, where some times it would render inconsistently, hiding or distorting the month header views or some day cells. This was due to the LazyVGrid not being consistent with the calculations for cell reutilisation. 

This new approaches changes the calculation of the calendar grid less ui-based (where it dependend on the cell position before) to a more consistent input-output based way of calculating it that is detached from UI, leaving inconsistencies behind and also making this part of the logic unit testable


Backpack-SwiftUI/Calendar/Classes/Core/CalendarGridCalculator.swift
```diff
@@ -0,0 +1,81 @@
...
+    /// Calculated will look like this:
+    ///     [
+    ///         [nil, nil, nil, nil, 1, 2, 3],
+    ///         [4, 5, 6, 7, 8, 9, 10],
+    ///         ...,
+    ///         [27, 28, 29, 30, 31, nil, nil]
+    ///     ]
+    /// Where initial nils are days from previous month, then 1...`numberOfDaysInMonth``,
+    /// then nils to fill the last week
+    func calculateCalendarGrid(monthDate: Date) -> [[Date?]] {
```
frugoman, line 53: this logic is a bit complex (as it was before) but now its contained in a structure that can be unit tested! no more guessing work when changing it.

Backpack-SwiftUI/Calendar/Classes/Core/CalendarGridCalculator.swift
```diff
@@ -0,0 +1,81 @@
...
+        let daysInAWeek: Int = 7
+        let firstWeekday = calendar.firstWeekday // Locale-aware first day of the week
+        let weekdayOfMonthStart = calendar.component(.weekday, from: monthDate)
+        // Calculate the offset based on the first weekday
+        let daysFromPreviousMonth = (weekdayOfMonthStart - firstWeekday + daysInAWeek) % daysInAWeek
+        let numberOfDaysInMonth = calendar.range(of: .day, in: .month, for: monthDate)!.count
+        let totalCellsUsed = numberOfDaysInMonth + daysFromPreviousMonth
+        
+        var calculated: [[Date?]] = Array(
```
frugoman, line 62: starts of with an all-nil array, to set its size correclty

Backpack-SwiftUI/Calendar/Classes/Core/CalendarGridCalculator.swift
```diff
@@ -0,0 +1,81 @@
...
+        )
+
+        // Iterate through each day of the month
+        for day in 0..<numberOfDaysInMonth {
+            // Calculate the position in the grid
+            let row = (day + daysFromPreviousMonth) / daysInAWeek
+            let column = (day + daysFromPreviousMonth) % daysInAWeek
+            
+            // Calculate the date for the current day
+            let date = calendar.date(byAdding: .day, value: day, to: monthDate)
```
frugoman, line 74: now we go through all days in a month and set the value of the day in the correct cell in the grid.

Backpack-SwiftUI/Calendar/Classes/Core/CalendarMonthGrid.swift
```diff
@@ -20,110 +20,78 @@ import SwiftUI
...
-            }
-        ForEach(preEmptyCells) { _ in
-            VStack(spacing: BPKSpacing.none) {
-                emptyLeadingDayCell()
-                    .frame(height: dayCellHeight)
-                Spacer(minLength: BPKSpacing.none)
+    private func weekRow(grid: [[Date?]], row: Int) -> some View {
+        HStack(alignment: .top, spacing: BPKSpacing.none) {
+            ForEach(0..<grid[0].count, id: \.self) { col in
+                dayCellView(grid: grid, row: row, col: col)
```
frugoman, line 54: We simply go through the grid and render the corresponding cell, given the value

Backpack-SwiftUI/Calendar/Classes/Core/CalendarMonthGrid.swift
```diff
@@ -20,110 +20,78 @@ import SwiftUI
...
-        let id: String
-        let index: Int
+    @ViewBuilder
+    private func dayCellView(dayDate: Date) -> some View {
+        if !validRange.contains(dayDate) {
+            disabledDayCell(dayDate)
+                .frame(height: dayCellHeight)
+                .frame(maxWidth: .infinity)
+        } else {
+            VStack(spacing: BPKSpacing.sm) {
```
frugoman, line 76: the logic for rendering cells is the same as before, not much to look at in this aspect

frugoman: Except is a bit simplified due to not needing a LazyVGrid

frugoman (commented):
What's reused now (as in Lazy drawing of UI) is on a per-month basis. once a month is rendered, the 'day cells' are not reusable, until we scroll enough where the month entirely is not visible and then is reused on its entirety.

Having lazy under lazy was one of the problems before, where when SwiftuI was calculating sizes while scrolling, it did not have all the values when needed, leading to inconsistencies at render time

brunomdac (approved):

PR #2183, opened 2025-03-07
[ACC-1027] Adds `BPKDynamicStack` based on native `AnyLayout`

novinfard:
[ACC-1027] Adds `BPKDynamicStack` based on native `AnyLayout`
https://skyscanner.atlassian.net/browse/ACC-1027

📑 Check the documentation for more detail:
https://github.com/Skyscanner/backpack-ios/blob/9adde213c2f0850dd4aa68d397d34ccac48e3999/Backpack-SwiftUI/DynamicStack/README.md

👁️ `BPKDynamicTypeStack` Example
test_dynamicTypeStack_accessibility a11y

[ACC-1027]: https://skyscanner.atlassian.net/browse/ACC-1027?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Backpack-SwiftUI/DynamicStack/Classes/BPKDynamicStack.swift
```diff
@@ -0,0 +1,82 @@
...
+       - horizontalAlignment: The alignment for the vertical stack.
+       - verticalAlignment: The alignment for the horizontal stack.
+       - spacing: The spacing between elements in both layouts.
+       - threshold: The Dynamic Type size threshold at which the layout switches.
+       - content: The view content inside the stack.
+     */
+    public init(
+        horizontalAlignment: HorizontalAlignment = .center,
+        verticalAlignment: VerticalAlignment = .center,
+        spacing: BPKSpacing = .none,
```
novinfard, line 63: Two spacing for Vstack, HStack

novinfard:
@github-copilot please review this PR

Backpack-SwiftUI/DynamicStack/Classes/BPKDynamicStack.swift
```diff
@@ -0,0 +1,60 @@
...
+ */
+
+import SwiftUI
+import Backpack_Common
+
+/// A stack view that switches between two layouts depending on a given condition.
+///
+/// `BPKDynamicStack` allows you to provide two different layout configurations (e.g., HStack or VStack)
+/// and dynamically switches between them based on an external Boolean condition, such as a Dynamic Type size threshold.
+public struct BPKDynamicStack<Content: View>: View {
```
yurareutskiy, line 26: It seems this container is not focusing on a stack specifically
What if we would call it something like BPKDynamicLayout? Then we would be able to use that in other components if we need

yurareutskiy (approved):
Looks really good! Very important component for us, good work

Backpack-SwiftUI/DynamicStack/Classes/BPKDynamicStack.swift
novinfard: Good point Yura!
We can have Grid layout for instance, which is not a stack. I updated all naming to reflect `Layout` instead of `Stack`

PR #2185, opened 2025-03-11
DON-1275: Fix bpkcalendar accessibility

yurareutskiy:
There are two potential accessibility issues:

1. Apparently, if there is a higher level .accessibilityHidden() modifier, it would override everything for children
In the DateSelector case we added this for the Cheapest month, this harmless modifier affects everything inside. But this header always should be hidden for accessibility. The easiest workaround it to override its accessibility behaviour.

2. Because of the complex structure of the day and accessory view cells, there is a major accessibility issue when a day view is separately from an accessory view. It makes a lot of confusion for accessibility users. So, we had to combine its content. (small note about a button trait, it had to be placed into a different part as it affected accessibility view's structure).

https://github.com/user-attachments/assets/f5a761c6-8396-42a5-a23d-bca114d927de

+ [ ] `README.md`
+ [ ] Tests
+ [ ] Adding a component? Remember to expose it in the [main `Backpack.h` header file

](https://github.com/Skyscanner/backpack-ios/tree/main/Backpack/Backpack.h)

### DIFF
--- a/Backpack-SwiftUI/Calendar/Classes/CalendarHeader.swift
+++ b/Backpack-SwiftUI/Calendar/Classes/CalendarHeader.swift
@@ -34,7 +34,7 @@ struct CalendarHeader: View {
             }
             Divider()
         }
-        .accessibilityHidden(true)
+        .accessibilityChildren { EmptyView().accessibilityHidden(true) }
     }
     
     private var weekdays: [String] {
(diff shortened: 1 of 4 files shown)


Backpack-SwiftUI/Calendar/Classes/CalendarHeader.swift
```diff
@@ -34,7 +34,7 @@ struct CalendarHeader: View {
             }
             Divider()
         }
-        .accessibilityHidden(true)
+        .accessibilityChildren { EmptyView().accessibilityHidden(true) }
```
novinfard, line 37:
```suggestion
         .accessibilityElement(children: .ignore)
```
Wouldn't this have the same effect? To avoid using EmptyView

yurareutskiy: I tried this and even that this header is not readable anymore, this solution is not ideal as it left the whole empty block and it could be confusing

https://github.com/user-attachments/assets/3b32d140-6bf3-48db-9b9d-41025b523720

maxgirt (approved):
LGTM!

novinfard (approved):

PR #2212, opened 2025-04-22
[MUON-868] Gallery Preview "Default" variant

darioroa:
This PR:

- Creates a `default` variant for the Gallery Preview (renaming the existing one to `Hero` as per the [Backpack design updates](https://www.figma.com/design/irZ3YBx8vOm16ICkAr7mB3/Backpack-Components?node-id=54200-107133&t=h10ozAxTmjO7fx9B-4).

- Updates docs with the corresponding rename
- Has no breaking changes for the existing variant

+ [ ] `README.md`
+ [ ] Tests


Backpack-SwiftUI/ImageGalleryPreview/Classes/BPKImageGalleryPreview.swift
```diff
@@ -19,21 +19,53 @@
 import SwiftUI
 
 public struct BPKImageGalleryPreview<Content: View>: View {
+    private enum Variant {
+        case hero
+        case `default`
+    }
```
darioroa, line 25: Opting to not expose this, instead relying on different `init`s for the two variants, since they require different inputs.

Backpack-SwiftUI/ImageGalleryPreview/Classes/BPKImageGalleryPreview.swift
```diff
@@ -61,7 +93,29 @@ public struct BPKImageGalleryPreview<Content: View>: View {
...
         }
-        
+    }
+    
+    private var defaultView: some View {
+        ZStack(alignment: .bottomTrailing) {
+            GeometryReader { geometry in
+                images.first
+                    .frame(width: geometry.size.width, height: geometry.size.height)
+            }
```
darioroa, line 103: Using a `GeometryReader` to ensure the image does not dictate the size of the stack - rather, it adapts to it.

Backpack-SwiftUI/ImageGalleryPreview/Classes/BPKImageGalleryPreview.swift
```diff
@@ -19,21 +19,53 @@
...
         self.onImageClicked = onImageClicked
+        self.buttonText = nil
+    }
+    
+    public init(
+        image: Content,
+        onButtonClicked: (() -> Void)? = nil,
+        buttonText: String
+    ) {
+        self.variant = .default
```
novinfard, line 50: Can't we make variant part of initiliser by default value and avoid making secondary initialiser?

novinfard: Have we updated the Backpack API table about as a source of reference ?

darioroa: > Can't we make variant part of initiliser by default value and avoid making secondary initialiser?

This is a good point! The thing is we also require different inputs for the variants. Instead of an array, we just need one image, and we do need the button text. Also ,the action does not make sense to have an index.

In that sense the variant parameter would be redundant - i.e it's impossible to have `hero` variant with the `default` init

darioroa: > Have we updated the Backpack API table about as a source of reference ?

I can ask the designers and double check

novinfard: Got it. In this case we can have associate values for each variant please? Then the old initialiser can just utilise the new initialiser that accept the variant enum with associate value as parameter. Also having some concise comment in each initialiser would be helpful

novinfard: Thank you @darioroa. Looks good! 🚀

novinfard (approved):
Nice work 🚀

PR #2243, opened 2025-06-02
[DON-1257] Fix clear button accessibility issue in `BPKSearchInputSummary`

novinfard:
[DON-1257]
Fix clear button accessibility issue in `BPKSearchInputSummary`


Backpack-SwiftUI/SearchInputSummary/Classes/BPKSearchInputSummary.swift
```diff
@@ -102,6 +102,7 @@ public struct BPKSearchInputSummary: View {
         .if(!BPKFont.enableDynamicType, transform: {
             $0.sizeCategory(.large)
         })
+        .id("\(accessibilityIdentifier)_accessory_\(text.isEmpty ? "hidden" : "visible")")
```
novinfard, line 105: Without this change, the accessibility tree inside this view will not update correctly when switching between hidden and visible accessory.

faurevid (approved):

PR #2350, opened 2025-09-10
[MUON-1148] `BPKCardList` Make single rail element fill full width

darioroa:
[MUON-1148](https://skyscanner.atlassian.net/browse/MUON-1148?atlOrigin=eyJpIjoiYmIzM2IxMDM4NTk0NDRmNzg4NmIzM2U3NjIwZTY4ODgiLCJwIjoiaiJ9)

Rail layout now expands a single element to occupy available horizontal space instead of using a horizontal scroll container.

+ [ ] `README.md`
+ [ ] Tests

### DIFF
--- a/Backpack-SwiftUI/CardList/Classes/BPKCardList.swift
+++ b/Backpack-SwiftUI/CardList/Classes/BPKCardList.swift
@@ -50,23 +50,36 @@ public struct BPKCardList<Element: Identifiable, Content: View>: View {
 
     public var body: some View {
         switch layout {
-        case .rail(let sectionHeaderAction):
-            railLayout(with: sectionHeaderAction)
+        case .rail(let sectionHeaderAction, let cardWidth):
+            railLayout(with: sectionHeaderAction, cardWidth: cardWidth)
         case .stack(let accessory):
             stackLayout(with: accessory)
         }
     }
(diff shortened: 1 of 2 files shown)


Backpack-SwiftUI/CardList/Classes/BPKCardList.swift
```diff
@@ -50,23 +50,36 @@ public struct BPKCardList<Element: Identifiable, Content: View>: View {
...
+                        .frame(maxWidth: .infinity)
                 }
                 .padding(.horizontal, .base)
                 .padding(.vertical, .sm)
+            } else {
+                ScrollView(.horizontal, showsIndicators: false) {
+                    HStack(alignment: .top, spacing: .base) {
+                        ForEach(elements) { element in
+                            cardForElement(element)
+                                .applyCardWidth(cardWidth)
```
darioroa, line 77: Applying the width here. If we delegate to the implementation, we are then unable to modify for the 1-item case.

darioroa: This is not breaking for existing implementations, as those already enforce the width, and by default they'll have `nil` in this case, which leaves them unmodified.

PR #2357, opened 2025-09-18
DON-2091 Fix accessibility 

faurevid:
Fix accessibility where last day of the week can't be selected in calendar

Following this conversation with designers it has been decided to add a padding around the calendar when voice over is activated to prevent the scroll bar to cover the activation point of the right side dates: https://skyscanner.slack.com/archives/C07LH0LRCAK/p1758121883831479

+ [ ] `README.md`
+ [ ] Tests

### DIFF
--- a/Backpack-SwiftUI/Calendar/Classes/Core/CalendarContainer.swift
+++ b/Backpack-SwiftUI/Calendar/Classes/Core/CalendarContainer.swift
@@ -29,7 +29,7 @@ struct CalendarContainer<MonthContent: View>: View {
 
     @State private var hasScrolledToItem = false
     @StateObject private var visibilityObserver: ItemVisibilityObserver
-    @Environment(\.accessibilityVoiceOverEnabled) var vo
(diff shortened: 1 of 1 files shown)


Backpack-SwiftUI/Calendar/Classes/Core/CalendarContainer.swift
```diff
@@ -29,6 +29,7 @@ struct CalendarContainer<MonthContent: View>: View {
 
     @State private var hasScrolledToItem = false
     @StateObject private var visibilityObserver: ItemVisibilityObserver
+    @Environment(\.accessibilityVoiceOverEnabled) var vo
```
kynansongwork: Should this be something like voiceOverEnabled? Or is that just a bit of repitition?

faurevid: You're right it's more readable that way

PR #2662, opened 2026-08-17
[MUON-1998] Observe video player progress

darioroa:
iOS Counterpart to https://github.com/Skyscanner/backpack-android/pull/2793
## What changed

- Add `BPKVideoPlayerProgress` with cumulative `playTime`, finite `duration`, and maximum `fractionPlayed`.
- Expose current progress and a replaying Combine stream through `BPKVideoPlayerController.progress` and `progressPublisher`.
- Add the SwiftUI `.onProgress` convenience API.
- Track active playback across pauses, buffering, loops, completion, resets, and seeks without counting seek distance.
- Add a Backpack example HUD that displays live progress and demonstrates the 2-second and quartile thresholds, plus fullscreen play/pause controls.

Screenshot 2026-08-21 at 13 07 25

## Why

Homepage Hero Video analytics needs playback facts for a view event after two watched seconds and quartile events at 25%, 50%, 75%, and 100%. Visibility and one-shot analytics policy remain in the consuming application rather than Backpack.

The public vocabulary is aligned with the Android proposal: progress is state-based and carries cumulative play time, duration, and fraction played.

## API impact

This adds a concise SwiftUI- and ViewModel-friendly observable surface. Existing video-player initializers and playback controls are unchanged.

[MUON-1998]: https://skyscanner.atlassian.net/browse/MUON-1998?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Backpack-SwiftUI/VideoPlayer/README.md
darioroa, line 1: Exposing the metrics through a combine publisher. This gives a lot of flexibility to inspect both in the view or by using combine in a viewModel to observe the changes.

There is also an `onProgress` that returns the `BPKVideoPlayerProgress` directly on the view.

Backpack-SwiftUI/VideoPlayer/Classes/BPKVideoPlayerProgress.swift
darioroa, line 1: **Decision:** We use `TimeInterval` because it is the standard iOS representation for time in seconds and integrates naturally with `Foundation` and `AVFoundation`. Android uses Long milliseconds because that is ExoPlayer’s native format. Matching each platform’s conventions avoids unnecessary conversions while keeping the API semantically aligned. Float would provide less precision without making the unit clearer.

The data model we'll use to communicate changes to the consumer

Backpack-SwiftUI/VideoPlayer/Classes/BPKVideoPlayer.swift
darioroa, line 1: **Decision:** Using an additional callback to access the same `progress` without Combine. We should not assume BPK users will use anything but vanilla SwiftUI.

A custom `onProgress` modifier we use on `BPKVideoPlayer` for consumers that don't need Combine.

Backpack-SwiftUI/VideoPlayer/Classes/BPKVideoPlayerProgressAccumulator.swift
darioroa, line 1: **Decision:** Using a very stateful component, handling logic locally. Supporting seeking and looping behaviours - some of it which isn't necessarily used in this target implementation but serves as a foundation and completeless for BPKPlayer.

Serves as a translator between our target metrics and what `AVPlayer` exposes

Example/Backpack/SwiftUI/Components/VideoPlayer/VideoPlayerExampleView.swift
```diff
@@ -183,6 +184,138 @@ struct VideoContinuousPlaybackExampleView: View {
     }
 }
 
+// MARK: - Use case 4: Live playback progress
+
+struct VideoProgressExampleView: View {
```
darioroa, line 189: Preview mostly for debug and internal development. Can simplify or remove if adds too much overhead or loc:

Image

Backpack-SwiftUI/VideoPlayer/Classes/BPKVideoPlayerController.swift
```diff
@@ -119,11 +171,20 @@ public final class BPKVideoPlayerController: ObservableObject {
...
-        player.seek(to: time, toleranceBefore: .zero, toleranceAfter: .zero)
+        prepareProgressForSeek()
+        player.seek(to: time, toleranceBefore: .zero, toleranceAfter: .zero) { [weak self] _ in
+            DispatchQueue.main.async {
+                self?.finishProgressSeek()
+            }
+        }
     }
 
     /// Pause and seek to the beginning — use for carousel items that reset on scroll-off.
```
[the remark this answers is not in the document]
darioroa: Fixed in 5c6735de8. Seek finalization now requires AVPlayer to report `finished == true` and is gated by a monotonically increasing seek ID, so a stale completion cannot clear the accumulator state for a newer seek. Added regression coverage for two overlapping seeks.

Backpack-SwiftUI/VideoPlayer/Classes/BPKVideoPlayerController+ProgressObserving.swift
```diff
@@ -0,0 +1,126 @@
...
+
+    func stopProgressObserving() {
+        if let periodicTimeObserverToken {
+            periodicTimeObserver.removePeriodicTimeObserver(periodicTimeObserverToken, from: player)
+        }
+        if let itemCompletionToken {
+            notificationCenter.removeObserver(itemCompletionToken)
+        }
+    }
+
```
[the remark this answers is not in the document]
darioroa: Fixed in 5c6735de8. Observer tokens are now cleared immediately after removal, and periodic observer registration is guarded against duplicates. Added tests covering repeated start and repeated stop.

Backpack-SwiftUI/VideoPlayer/README.md
```diff
@@ -140,6 +140,46 @@ controller.state.isPlaying  // true only when .playing
...
+    }
+```
+
+`progress` is `nil`, and the publisher remains silent, until the video has a finite, positive duration. Each emitted `BPKVideoPlayerProgress` contains:
+
+| Property | Meaning |
+| --- | --- |
+| `playTime` | Cumulative media time that actually advanced. Pauses and buffering do not add time; loops continue accumulating. |
+| `duration` | The finite, positive media duration in seconds. |
+| `fractionPlayed` | The highest playhead fraction reached, clamped to `0...1`. It does not reset when a video loops. |
```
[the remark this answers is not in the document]
darioroa: Checked the current branch: the table already uses a single leading pipe for the header, separator, and every row, and renders as a normal two-column Markdown table. No change needed here.

Backpack-SwiftUI/VideoPlayer/Classes/BPKVideoPlayerController.swift
```diff
@@ -61,29 +62,79 @@ public final class BPKVideoPlayerController: ObservableObject {
...
+    var progressAccumulator = BPKVideoPlayerProgressAccumulator()
+    let progressSubject = CurrentValueSubject<BPKVideoPlayerProgress?, Never>(nil)
 
     private var playerLooper: AVPlayerLooper?
     private var itemStatusObservation: NSKeyValueObservation?
     private var timeControlObservation: NSKeyValueObservation?
     private var currentItemObservation: NSKeyValueObservation?
+    var periodicTimeObserverToken: Any?
+    var itemCompletionToken: NSObjectProtocol?
     private var loadTimeoutTask: DispatchWorkItem?
```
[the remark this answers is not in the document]
darioroa: This visibility is intentional and limited to the progress-observing implementation. Swift `private` is file-scoped for these members, so keeping the observation hooks in the requested `BPKVideoPlayerController+ProgressObserving.swift` extension requires internal access. Moving them back into the controller would restore `private` but would also retangle the progress mechanics with the existing controller observations. The exposed configuration dependencies are immutable `let`s; only the observation/accumulator state is mutable. No public API is widened.

faurevid (approved):

PR #2675, opened 2026-09-03
[MUON-2058][MUON-2026] Fix video playback state and mute controls

darioroa:
## Summary

This PR fixes looping playback-state synchronization for `BPKVideoPlayerController` and adds the non-breaking mute-control API requested by [MUON-2026](https://skyscanner.atlassian.net/browse/MUON-2026).

## MUON-2058: looping playback state

When `AVPlayerLooper` reaches an item boundary, AVFoundation replaces `currentItem` and can briefly report `.paused` or `.waitingToPlayAtSpecifiedRate` while the replacement item is prepared. Treating that transient signal as a real pause can leave published SwiftUI state and playback controls out of sync.

The controller now:

- Reconciles current-item and transport KVO against the observed item, so callback ordering cannot publish a false pause.
- Keeps loop handoff state in `.buffering` while a replacement item becomes ready, avoiding `.readyToPlay` flicker.
- Clears the transition guard when the current item becomes `nil`.
- Rebinds item-status and completion observation for replacement items and ignores stale item callbacks.
- Keeps explicit pauses distinct from loop-boundary transport pauses.
- Isolates controller state and tests on `@MainActor`.
- Fails async tests immediately on timeout instead of cascading later assertions.
- Centralizes the repeated active-state check in `BPKVideoPlayerState.isActive`.

The continuous-playback example also had a separate observation bug: its fullscreen custom button received the shared controller through a plain property, so the video surface updated while the button icon stayed stale. The fullscreen wrapper now observes the resolved shared/owned controller and the continuous-playback example includes a working mute control.

## MUON-2026: observable mute controls

Custom integrations previously needed a local `@State` value and direct access to `player.isMuted`:

```swift
isMuted.toggle()
playerController.player.isMuted = isMuted
```

The controller now exposes:

```swift
@Published public private(set) var isMuted: Bool
public func mute()
public func unmute()
public func toggleMute()
```

Mute actions publish synchronously for immediate UI reads, while `AVPlayer.isMuted` observation keeps existing direct `player.isMuted` writes synchronized with the published UI state.

The Jira ticket also mentions making `AVPlayer` private. That would be a breaking change, so this PR intentionally keeps the existing public `player` property and only adds the new API. This keeps current consumers source-compatible while giving new custom UIs a controller-level seam.

## Test coverage

New controller regressions use a small bundled local MP4 and run offline:

- `test_muteActions_publishObservableState()` covers synchronous `mute()`, `unmute()`, `toggleMute()`, and player state.
- `test_isMuted_tracksLegacyPlayerMutation()` covers backward compatibility with direct `player.isMuted` writes.
- Loop tests cover item replacement without false `.paused` or `.readyToPlay`, continued playback, explicit pause at the item boundary, and a missing current item.
- Tests use the required `12 Mini` simulator and the class is main-actor isolated.
- The README documents the corrected state semantics, valid shared-controller examples, and mute integration.

## Verification

- Seven focused controller tests pass in 3.0 seconds on the required `12 Mini` simulator.
- The Backpack Native example app build passes.
- Running the complete `BPKVideoPlayerTests` class leaves only the three existing snapshot test cases failing because the current simulator renders `375x500` while the checked-in references are `250x333`; no controller behavior test fails.
- No UI automation test was added for tapping through the continuous-playback demo; the example build verifies the wiring compiles, and the controller tests cover the state transitions.

[MUON-2026]: https://skyscanner.atlassian.net/browse/MUON-2026?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Backpack-SwiftUI/VideoPlayer/Classes/BPKVideoPlayerController.swift
```diff
@@ -221,36 +227,66 @@ public final class BPKVideoPlayerController: ObservableObject {
...
         case .playing:
             transition(to: .playing)
         case .paused:
-            // Suppress .paused during initial load — only meaningful after we were playing
-            if state == .playing { transition(to: .paused) }
+            // AVPlayerLooper briefly reports `.paused` while it replaces a completed item.
+            // Wait for the replacement item's status before publishing a state change.
+            if !isLoopItemTransitioning {
+                transition(to: .paused)
+            }
```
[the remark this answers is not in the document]
darioroa: Addressed in commit 0d9a47178. Added an explicit pause-request guard so an initial paused transport callback cannot override the expected ready-to-play state or trigger autoplay after an explicit pause.

Backpack-SwiftUI/Tests/VideoPlayer/BPKVideoPlayerTests.swift
```diff
@@ -46,6 +47,47 @@ final class BPKVideoPlayerTests: XCTestCase {
...
+        )
+
+        try await waitUntil { controller.state == .readyToPlay }
+        let initialItem = try XCTUnwrap(controller.player.currentItem)
+
+        controller.play()
+
+        try await waitUntil { controller.player.currentItem !== initialItem }
+        try await waitUntil { controller.state.isPlaying }
+
```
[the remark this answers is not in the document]
darioroa: Addressed in commit 0d9a47178. The regression tests now seek to 2.5 seconds of the 3-second fixture before starting playback, making the loop-boundary transition deterministic.

Backpack-SwiftUI/VideoPlayer/Classes/BPKVideoPlayerController.swift
```diff
@@ -221,36 +227,66 @@ public final class BPKVideoPlayerController: ObservableObject {
...
-            DispatchQueue.main.async { self?.handle(itemStatus: item.status) }
+            DispatchQueue.main.async { self?.handle(itemStatus: item.status, for: item) }
         }
     }
 
-    private func handle(itemStatus: AVPlayerItem.Status) {
+    private func handleCurrentItemChange(_ item: AVPlayerItem?) {
+        isLoopItemTransitioning = loop && state == .playing
+        observeItemStatus(item)
+    }
```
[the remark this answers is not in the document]
darioroa: Addressed in commit 0d9a47178. Loop handoff detection now includes both controller buffering state and active or waiting AVPlayer transport status, so the transient paused callback remains suppressed during buffering handoffs too.

Package.swift
```diff
@@ -351,7 +351,8 @@ let targets: [Target] = [
       ],
       path: "Backpack-SwiftUI/Tests",
       resources: [
-        .process("Images.xcassets")
+        .process("Images.xcassets"),
+        .process("skyscanner_test.mp4")
```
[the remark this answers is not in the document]
darioroa: Addressed in commit 0d9a47178. The MP4 test fixture is now declared with SwiftPM copy so it is shipped unchanged.

Backpack-SwiftUI/VideoPlayer/Classes/BPKVideoPlayerController.swift
```diff
@@ -221,36 +230,71 @@ public final class BPKVideoPlayerController: ObservableObject {
...
-            DispatchQueue.main.async { self?.handle(itemStatus: item.status) }
+            DispatchQueue.main.async { self?.handle(itemStatus: item.status, for: item) }
         }
     }
 
-    private func handle(itemStatus: AVPlayerItem.Status) {
+    private func handleCurrentItemChange(_ item: AVPlayerItem?) {
+        let transportIsActive = player.timeControlStatus == .playing ||
+            player.timeControlStatus == .waitingToPlayAtSpecifiedRate
+        isLoopItemTransitioning = loop && (state == .playing || state == .buffering || transportIsActive)
```
darioroa: Reviewer context: this flag is intentionally limited to loop transitions. When AVPlayerLooper swaps currentItem, the transport can briefly report paused before the replacement item is ready; suppressing that transient signal prevents a false UI pause. pause() clears the flag before calling player.pause(), so an explicit user pause is still published immediately.

Example/Backpack/SwiftUI/Components/VideoPlayer/VideoPlayerExampleView.swift
```diff
@@ -111,21 +111,16 @@ struct VideoGraphicPromoExampleView: View {
...
 struct VideoFullscreenExampleView: View {
-    // When injected (use case 3), playback continues uninterrupted across the transition.
-    // When nil (use case 2 standalone), owns its own controller.
-    @StateObject private var ownedController = BPKVideoPlayerController(
-        url: SampleVideo.url,
-        autoPlay: true,
-        loop: true
-    )
-    private let injectedController: BPKVideoPlayerController?
+    @StateObject private var activeController: BPKVideoPlayerController
```
darioroa, line 116: Reviewer context: this is intentionally an observed StateObject even when the controller is injected. The continuous-playback sheet passes the existing shared controller, and the custom button must observe that same instance; the previous plain let left the button frozen while BPKVideoPlayer updated. When no controller is supplied, this also gives the standalone fullscreen example its own retained controller.

Package.swift
```diff
@@ -351,7 +351,8 @@ let targets: [Target] = [
       ],
       path: "Backpack-SwiftUI/Tests",
       resources: [
-        .process("Images.xcassets")
+        .process("Images.xcassets"),
+        .copy("skyscanner_test.mp4")
```
darioroa, line 355: Reviewer context: the fixture is copied as a raw test resource rather than processed as an asset because the AVFoundation integration test needs a stable local file URL and the original MP4 bytes. This keeps the regression tests offline and independent of the demo HLS stream. Happy to discuss this tradeoff.

Backpack-SwiftUI/Tests/VideoPlayer/BPKVideoPlayerTests.swift
```diff
@@ -46,6 +48,49 @@ final class BPKVideoPlayerTests: XCTestCase {
         })
     }
 
+    // MARK: - Controller playback state
+
+    func test_loopingPlayback_remainsPlayingWhenCurrentItemChanges() async throws {
```
darioroa, line 85: Covering the loop behaviour by using a local video.

darioroa: Over 100 runs, these two tests are 0.64s and 0.58s - happy to talk about the tradeoff if the cost is too high.

Backpack-SwiftUI/VideoPlayer/Classes/BPKVideoPlayerController.swift
```diff
@@ -204,14 +233,18 @@ public final class BPKVideoPlayerController: ObservableObject {
     private func observePlayer() {
         startProgressObserving()
 
+        mutedObservation = player.observe(\.isMuted, options: [.initial, .new]) { [weak self] player, _ in
```
darioroa, line 275: MUON-2026 context: this is intentionally published from AVPlayer KVO rather than maintained as a second independent source of truth. We are keeping the existing public player property for source compatibility, so legacy direct player.isMuted writes must also update the observable controller state. The public mute actions delegate to AVPlayer and the same observation path keeps all consumers synchronized.

Backpack-SwiftUI/VideoPlayer/Classes/BPKVideoPlayerController.swift
```diff
@@ -221,36 +254,76 @@ public final class BPKVideoPlayerController: ObservableObject {
...
+        let shouldAutoPlay = !hasLoadedInitialItem && autoPlay &&
+            !hasExplicitPauseRequest && !UIAccessibility.isReduceMotionEnabled
+        hasLoadedInitialItem = true
+
+        switch player.timeControlStatus {
+        case .playing:
+            transition(to: .playing)
+        case .waitingToPlayAtSpecifiedRate:
+            transition(to: .buffering)
+        case .paused:
```
novinfard, line 360: ⚠️  Small gap here: 

If the transport happens to read `.paused` while a loop replica becomes ready, this branch publishes `.readyToPlay` even though we were playing. 

- That's the same flicker the `isLoopItemTransitioning` guard in `handle(timeControlStatus:)`  is trying to stop, just coming in through the other door. 

Could we keep the current state (or go to .buffering) when we land here during a loop swap? For what it's worth, with the bundled MP4 the transport never left `.playing` across the swap in a traced run, so this only matters if the transient really shows up on HLS.

Backpack-SwiftUI/VideoPlayer/Classes/BPKVideoPlayerController.swift
novinfard: This flag only gets set once the currentItem change reaches the main queue. 

Both that callback and the `timeControlStatus` one are `DispatchQueue.main.async` hops, so if the `.paused` lands first the guard is still `false` and we publish `.paused` anyway. 

The fix depends on which hop wins. One way out is a single `reconcile()` that every KVO closure calls, reading `timeControlStatus`, `reasonForWaitingToPlay` and currentItem?.status together and deriving the state from that snapshot. 

Another is to mark the boundary from the `didPlayToEndTime` notification we already receive.

Backpack-SwiftUI/VideoPlayer/Classes/BPKVideoPlayerController.swift
```diff
@@ -221,36 +254,76 @@ public final class BPKVideoPlayerController: ObservableObject {
...
         }
     }
 
-    private func handle(itemStatus: AVPlayerItem.Status) {
+    private func handle(muted: Bool) {
+        guard isMuted != muted else { return }
+        isMuted = muted
+    }
+
+    private func handleCurrentItemChange(_ item: AVPlayerItem?) {
```
novinfard, line 306: If item is `nil` here (let's say the looper fails and the queue plays out), we set the flag to true and never clear it, since nothing else runs without an item. 

The next real `.paused` gets swallowed and state stays `.playing` for good, with no `.failed` and no timeout. The old code at least published `.paused` in that case. Maybe treat a nil item as terminal rather than a transition, and consider watching `playerLooper.status` so we can publish .failed.

Backpack-SwiftUI/Tests/VideoPlayer/BPKVideoPlayerTests.swift
```diff
@@ -46,6 +48,79 @@ final class BPKVideoPlayerTests: XCTestCase {
...
+            loop: true
+        )
+
+        try await waitUntil { controller.state.isPlaying }
+        let initialItem = try XCTUnwrap(controller.player.currentItem)
+
+        try await seekNearLoopBoundary(controller)
+        try await waitUntil { controller.player.currentItem !== initialItem && controller.state.isPlaying }
+
+        controller.pause()
```
novinfard: I ran this test against the old state machine (with only the mute API added) and it still passes. 

By the time we get past the wait on line 114 the swap has finished and `isLoopItemTransitioning` is already false, so this pause never touches the new logic. Either rename it to what it checks (pause works on the replacement item), or call `pause()` right after currentItem changes and assert the state never leaves `.paused` for a short window. 

Recording `controller.$state` into an array makes that easy. The first loop test does fail on the old code, so that one is doing its job.

I hope it makes sense?

Backpack-SwiftUI/Tests/VideoPlayer/BPKVideoPlayerTests.swift
```diff
@@ -77,6 +152,36 @@ final class BPKVideoPlayerTests: XCTestCase {
...
+    }
+
+    private func waitUntil(
+        _ condition: @escaping () -> Bool,
+        timeout: TimeInterval = 5
+    ) async throws {
+        let deadline = Date().addingTimeInterval(timeout)
+        while !condition() {
+            if Date() >= deadline {
+                XCTFail("Timed out waiting for video-player state")
```
novinfard: When this times out we log the failure and carry on, so every later wait and assert in the test fails too and the real cause gets buried under extra red lines. 

I was thinking ... the tests are already async throws, so could this throw a small error instead? `XCTNSPredicateExpectation`, or `controller.$state with fulfillment(of:)` like the progress tests use, would also get rid of the 50ms polling.

Backpack-SwiftUI/Tests/VideoPlayer/BPKVideoPlayerTests.swift
```diff
@@ -46,6 +48,79 @@ final class BPKVideoPlayerTests: XCTestCase {
         })
     }
 
+    // MARK: - Controller playback state
+
+    func test_muteActions_publishObservableState() async throws {
```
novinfard, line 55: These async tests run off the main actor, so `play()`, `pause()` and `seek()` mutate the controller's `@Published` state on a background thread, and the wait closures read state, `isMuted` and `currentTime()` while the controller writes them on main. 

I saw the `.paused` transition land on a **non-main thread** in a traced run. Marking the class `@MainActor`, like BPKVideoPlayerControllerProgressTests, would sort it. WDYT?

Backpack-SwiftUI/VideoPlayer/Classes/BPKVideoPlayerController.swift
```diff
@@ -183,6 +197,21 @@ public final class BPKVideoPlayerController: ObservableObject {
         }
     }
 
+    /// Mutes the player.
+    public func mute() {
```
novinfard, line 235: 🧼 Nit picking only
Right after `mute()` or `toggleMute()`, `isMuted` still reads the old value until the KVO hop lands on main, while `pause()` updates state straight away. 

So `toggleMute(); log(controller.isMuted)` logs the stale value. Right?
I think setting `isMuted = player.isMuted` inside these three can close the gap, and `handle(muted:)`

Backpack-SwiftUI/VideoPlayer/README.md
```diff
@@ -82,6 +82,22 @@ BPKVideoPlayer(url: videoURL, autoPlay: true) { controller in
...
+BPKVideoPlayer(url: videoURL, autoPlay: true) { controller in
+    Button(action: controller.toggleMute) {
+        Image(systemName: controller.isMuted ? "speaker.slash.fill" : "speaker.fill")
+    }
+    .accessibilityLabel(controller.isMuted ? "Unmute video" : "Mute video")
+    .accessibilityValue(controller.isMuted ? "Muted" : "Unmuted")
+}
+```
+
+The existing `player` property remains available for source compatibility, and direct `player.isMuted` changes are reflected in `controller.isMuted`.
```
novinfard, line 99: The table still says `.readyToPlay` means autoplay will call `play()`. 

Image


After this PR that **only happens for the first item**, and **not after an explicit pause()**. Right?

`.buffering` can also show up during a loop swap now, not just when rebuffering. 

Worth a line or two so nobody relying on `resetToStart()` before the asset is ready gets surprised. 

♻️ Side Note:  While you're in here: the `BPKVideoPlayer(controller:) snippets near lines 57 and 170 don't compile, there's only `init(controller:overlay:)`. Not from this PR I acknowledge, but easy to fix in passing, in case you'd like to improve it.

Backpack-SwiftUI/VideoPlayer/Classes/BPKVideoPlayerController.swift
```diff
@@ -221,36 +254,76 @@ public final class BPKVideoPlayerController: ObservableObject {
...
     private func handle(timeControlStatus: AVPlayer.TimeControlStatus) {
         switch timeControlStatus {
         case .playing:
             transition(to: .playing)
         case .paused:
-            // Suppress .paused during initial load — only meaningful after we were playing
-            if state == .playing { transition(to: .paused) }
+            // AVPlayerLooper briefly reports `.paused` while it replaces a completed item.
+            // Wait for the replacement item's status before publishing a state change.
+            let shouldPublishPause = state == .playing || state == .buffering ||
```
novinfard: The `state == .playing || state == .buffering` condition now appears here + in `pause()` + in `handleCurrentItemChange`. 

A small computed property next to `isPlaying` and `isLoading` (let's say `isActive`) would keep them in step. 

Also, as far as I can trace it (please double check Dario), the `hasExplicitPauseRequest && state == .readyToPlay` clause only fires if `play()` then `pause()` happen inside one runloop turn with the waiting KVO coalesced away. 

If that's the case it's meant for, a short comment would help. If not, it can probably go. So up to you.

Example/Backpack/SwiftUI/Components/VideoPlayer/VideoPlayerExampleView.swift
```diff
@@ -111,21 +111,16 @@ struct VideoGraphicPromoExampleView: View {
...
 
 struct VideoFullscreenExampleView: View {
-    // When injected (use case 3), playback continues uninterrupted across the transition.
-    // When nil (use case 2 standalone), owns its own controller.
-    @StateObject private var ownedController = BPKVideoPlayerController(
-        url: SampleVideo.url,
-        autoPlay: true,
-        loop: true
-    )
-    private let injectedController: BPKVideoPlayerController?
```
novinfard, line 121: 🧼 Nit picking 

Nothing in the example app uses the new mute API yet. A mute button on this screen or the GraphicPromo one would give people something to tap and show the `player.isMuted` sync working. 

Also, this `@StateObject` works fine for the sheet, and it drops the throwaway controller the old code built even when one was injected, which is nice. I'd just bring back a one-line comment saying **an injected controller is kept for the view's lifetime and the fallback is only built once**, since that's not obvious anymore.

Backpack-SwiftUI/Tests/VideoPlayer/BPKVideoPlayerTests.swift
darioroa: Agree! Changing 👍

Backpack-SwiftUI/VideoPlayer/Classes/BPKVideoPlayerController.swift
darioroa: Fixed in `af988e107edb457eea0d4bc2026f9cd6f5f903e6`: loop handoff detection now reconciles against the observed/current item from both KVO paths, so callback ordering cannot publish a transient pause.

Backpack-SwiftUI/VideoPlayer/Classes/BPKVideoPlayerController.swift
darioroa: Fixed in `426b185ead1852bd740819d89980f070cb147fad`: a ready replacement item preserves `.buffering` while the prior playback state was active, preventing `.readyToPlay` flicker during loop handoff.

Backpack-SwiftUI/VideoPlayer/Classes/BPKVideoPlayerController.swift
darioroa: Fixed in `dca69d9b5807a69615f4702bf4426451454b474d`, with deterministic coverage in `eef83f1209e1540567a0d49946fd72c4edabb340`: nil current items clear the transition guard instead of latching it.

Backpack-SwiftUI/Tests/VideoPlayer/BPKVideoPlayerTests.swift
darioroa: Fixed in `6177ee52984ba37177916889eee345093d470c61`: the test pauses from the current-item KVO boundary and records subsequent state history, proving explicit pause remains paused.

Backpack-SwiftUI/Tests/VideoPlayer/BPKVideoPlayerTests.swift
darioroa: Fixed in `8842b3aa58f0033be7a4115ae2168a667cc0b71d`: timeout now throws immediately, so later waits and assertions cannot cascade over the original failure.

Backpack-SwiftUI/Tests/VideoPlayer/BPKVideoPlayerTests.swift
darioroa: Fixed in `30be9bf3b56b192fb4338285637ecaea3c551fea`, with actor-compatible cleanup in `0ba55a96857b06e7dfc47d64cdc3cb438be43dca`: the controller and async test class are main-actor isolated.

Backpack-SwiftUI/VideoPlayer/Classes/BPKVideoPlayerController.swift
darioroa: Fixed in `c446b25178076f97890239a0f4a48718c497f7d3`: mute, unmute, and toggle update `isMuted` synchronously while KVO continues to cover legacy direct player mutations.

Backpack-SwiftUI/VideoPlayer/README.md
darioroa: Fixed in `5ad7a5101d295a4051211d6375f098ef5519f16c`: README state semantics and the invalid shared-controller examples are corrected.

Backpack-SwiftUI/VideoPlayer/Classes/BPKVideoPlayerController.swift
darioroa: Fixed in `79fe25b6cce33dcd60420c49f4562c7c95220e00`: added `BPKVideoPlayerState.isActive` and centralized the active-state checks.

Example/Backpack/SwiftUI/Components/VideoPlayer/VideoPlayerExampleView.swift
darioroa: Fixed in `85b5dc5733af83548f00634d1feb14c5277b47df`: the continuous-playback example now exposes mute/unmute UI, and the injected-controller lifetime rationale is documented.

Backpack-SwiftUI/VideoPlayer/Classes/BPKVideoPlayerController.swift
```diff
@@ -379,7 +379,13 @@ public final class BPKVideoPlayerController: ObservableObject {
...
             let shouldPublishPause = state.isActive ||
                 (hasExplicitPauseRequest && state == .readyToPlay)
-            if !isLoopItemTransitioning && shouldPublishPause {
+            let replacementIsStillLoading = isLoopItemTransitioning &&
+                player.currentItem?.status == .unknown
+            if !replacementIsStillLoading && shouldPublishPause {
+                // A paused transport is an external stop once the replacement
+                // item is ready. Clear the handoff guard so ready-item handling
+                // cannot turn the state back into buffering.
+                isLoopItemTransitioning = false
```
darioroa, line 388: Covers case where status hangs indefinitely in `buffering` when AVSession interrupted right on item transition (thanks for finding out,  @novinfard !)

novinfard (approved):
Looks good. Loads of great work happened in this PR by @darioroa ! 💯

